PR #60, opened 2017-05-17
Cublas

liruipeng:
CUDA CUBLAS GEMM SUPPORT IN ELEMENTAL


bvanessen:
I have updated this so that the changes to Elemental are applied via a patch file rather than overwriting the .[ch]pp directly.

PR #93, opened 2017-07-07
Add hwloc to link libraries for lbann

trb413:
this could be made more robust (in the traditional "FindHwloc.cmake"
sense) if desired. but this was sufficient for my test.


ndryden:
Confirmed this builds fine on LC (catalyst tested) and it's simple enough, so merging now.

PR #103, opened 2017-08-15
Updated doxygen documentation and readme.

DylanMcKinney:
Creates doxygen documentation that looks similar to keras. Fleshes out github readme. Documentation and readme are now consistent in how to get started with lbann.


bvanessen:
I am unclear as to why some of the spack options are being removed from the build_lbann.sh script.  For example, why are these being removed: build_type=${BUILD_TYPE} dtype=${DTYPE}

PR #109, opened 2017-09-08
LBANN Tests

forsyth2:
Initial tests for LBANN performance and accuracy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ data.prototext*
 
 # Can also ignore all directories and files in a directory.
 # tmp/**/* 
+build
--- a/model_zoo/tests/ridge_regression_test.sh
+++ b/model_zoo/tests/ridge_regression_test.sh
@@ -8,7 +8,7 @@ LBANN_DIR=$(git rev-parse --show-toplevel)
 LBANN_EXE="${LBANN_DIR}/build/${CLUSTER}.llnl.gov/model_zoo/lbann"
 
 # Run experiment
-srun --nodes=1 --ntasks=1 ${LBANN_EXE} --model=${LBANN_DIR}/model_zoo/tests/model_mnist_ridge_regression.prototext --optimizer=${LBANN_DIR}/model_zoo/optimizers/opt_adam.prototext --reader=${LBANN_DIR}/model_zoo/data_readers/data_reader_mnist.prototext
+salloc -N1 srun --ntasks=1 ${LBANN_EXE} --model=${LBANN_DIR}/model_zoo/tests/model_mnist_ridge_regression.prototext --optimizer=${LBANN_DIR}/model_zoo/optimizers/opt_adam.prototext --reader=${LBANN_DIR}/model_zoo/data_readers/data_reader_mnist.prototext
 
 # Return status
 if [ $? -ne 0 ]; then
--- a/python_tests/clang4_test.py
+++ b/python_tests/clang4_test.py
@@ -1,0 +1,20 @@
+import os
+import unittest
+
+from test_tools import *
+
+EXECUTABLE = 'NO_EXECUTABLE'
+
+class Clang4Test(unittest.TestCase):
+
+  def test_mnist_distributed_io(self):
+    mnist_distributed_io_skeleton(self, EXECUTABLE)
+
+  def test_alexnet(self):
+    alexnet_skeleton(self, EXECUTABLE)
+
+  def test_resnet50(self):
+    resnet50_skeleton(self, EXECUTABLE)
+
+if __name__ == '__main__':
+  unittest.main()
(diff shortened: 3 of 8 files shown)


forsyth2:
@DylanMcKinney Here's the tests.

.gitignore
```diff
@@ -7,3 +7,4 @@
 
 # Can also ignore all directories and files in a directory.
 # tmp/**/* 
+build
```
forsyth2: Add a new line.

scripts/test.py
```diff
@@ -0,0 +1,97 @@
...
+  return '--optimizer=../model_zoo/optimizers/%s.prototext' % optimizer_name
+
+def executable(cluster):
+  return '../build/%s/model_zoo/lbann' % cluster
+
+def run_lbann(model_name, reader_name, optimizer_name, output_file_name):
+  if (model_name == None or reader_name == None or optimizer_name == None or output_file_name == None):
+    raise Exception("Invalid number of arguments")
+  else:
+    os.system('hostname > %s' % output_file_name)
```
forsyth2: Inefficient to write to file just to get this information? Would it be better to use a subprocess?

scripts/test.py
```diff
@@ -0,0 +1,97 @@
...
+    e = executable(cluster)
+    m = model(model_name)
+    r = reader(reader_name)
+    o = optimizer(optimizer_name)
+    # N => number of nodes                                                         
+    # p => partition
+    # t => timeout period, in minutes
+    # n => number of processes to run. MPI Rank. How many models being made.
+    # num-epochs => The number of epochs
+    os.system('salloc -N 1 -p pbatch -t 30 srun -n 2 %s %s %s %s --num_epochs=5 > %s' % (e, m, r, o, output_file_name))
```
forsyth2: Remove `-t 30` for longer running times? Are there any options besides model, reader, optimizer we want to specify?

scripts/test.py
```diff
@@ -0,0 +1,97 @@
...
+      performance_dict[model]['epoch_times'].append(float(m.group(2)))
+      performance_dict[model]['mean_minibatch_times'].append(float(m.group(4)))
+      performance_dict[model]['mins'].append(float(m.group(6)))
+      performance_dict[model]['maxs'].append(float(m.group(8)))
+      performance_dict[model]['stdevs'].append(float(m.group(10)))
+    a = re.match('(Model [0-9]+) @[0-9]+ testing steps external validation categorical accuracy: ([0-9]+(\.[0-9]+)*)', line)
+    if a:
+      model = a.group(1)
+      if model not in performance_dict.keys():
+        performance_dict[model] = {
```
forsyth2: Make an `accuracy_dict` and put accuracy in that one?

scripts/test.py
```diff
@@ -0,0 +1,97 @@
...
+  run_lbann(model_name, reader_name, optimizer_name, output_file_name)
+  return get_performance(output_file_name)
+
+class LBANNTest(unittest.TestCase):
+  def test_performance(self):
+    performance = main(model_name='model_mnist_multi', reader_name='data_reader_mnist', optimizer_name='opt_adagrad', output_file_name='output.txt')
+    # performance = get_performance(output_file_name='output.txt')
+    for model in performance.keys():
+      for epoch_time in performance[model]['epoch_times']:
+        self.assertTrue(epoch_time < 67)
```
forsyth2: What values should we be testing against?

scripts/test.py
```diff
@@ -0,0 +1,97 @@
...
+      performance_dict[model]['accuracies'].append(float(a.group(2)))
+  output_file.close()
+  return performance_dict
+
+def main(model_name, reader_name, optimizer_name, output_file_name):
+  run_lbann(model_name, reader_name, optimizer_name, output_file_name)
+  return get_performance(output_file_name)
+
+class LBANNTest(unittest.TestCase):
+  def test_performance(self):
```
forsyth2: Maybe we should keep performance and accuracy together per test. However, each test will test a different model.
Perhaps rename this `test_mnist_performance` and add `test_resnet_performance` and `test_alexnet_performance`. Abstract out common code.

scripts/test.py
bvanessen: Eventually we are going to want to vary the number of nodes and processes, which also dictates the processes per node.

davidHysom:
You may have noticed, I've instituted a convention that prototext file names are of the form:

  model_XX.prototext

  data_reader_XX.prototext

  model_XX.prototext

If we want to adopt this convention, I would suggest that,

when specifying files for the tests, users should not have

to specify the 'model_', etc, portion of the names.

OTOH, perhaps my convention is overly restrictive.


forsyth2:
@DylanMcKinney I added some minor improvements to the test script but I still need to run the tests to make sure nothing broke.

.gitignore
forsyth2: Added.

PR #120, opened 2017-10-19
DARC1 regularization

timmoon10:
This is an implementation of the DARC1 regularization described in "Generalization in Deep Learning" by Kawaguchi, Kaelbling, and Bengio (2017). Preliminary experiments with Resnet-50 on ILSVRC2012 do not show any improvement over the baseline, although results are tentative.

### First 10 image classes of ILSVRC2012
Each line shows the average over 5 runs. 10% of training set was held out as a validation set. Solid line uses DARC1 and dotted line is baseline.

### Entirety of ILSVRC2012
Solid line uses DARC1 and dotted line is baseline. 10% of training set was held out as a validation set.

Note that this model has not been run to convergence, so it is possible DARC1 achieves something in that regime. In addition, I just used the hyperparameter from the paper, so hyperparameter tuning could yield some result.

### DIFF
--- a/include/lbann/layers/regularizers/darc1.hpp
+++ b/include/lbann/layers/regularizers/darc1.hpp
@@ -1,0 +1,191 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2016, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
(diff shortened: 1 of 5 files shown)


timmoon10:
In any case, I don't think we should merge this pull request until we have a more compelling reason to maintain this functionality.

PR #126, opened 2017-11-03
Bamboo fixes

forsyth2:
More updates to the testing framework.


bamboo/integration_tests/accuracy_tests/accuracy_test.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -o pipefall
```
forsyth2: Remove.

PR #132, opened 2017-11-17
Refactoring weights and optimizers

timmoon10:
This pull request implements a weights object that handles layer parameters. Each weights object maintains an optimizer (possibly a null pointer) and an initializer (e.g. Glorot normal). Layer have a list of pointers to their weights. If two layers share weights, they can just point to the same weights object. Importantly, weights no longer have to be collapsed into a single weights matrix. For instance, the parameters for a model-parallel fully-connected matrix are the matrix weights in MC,MR format and the bias weights in MC,STAR format.

This refactoring is not quite complete, but I think it is best to merge this in quickly before the code base changes too much. The following tasks need to be done after merging:
- [x] Restoring L2 regularization
- [ ] Restoring group lasso regularization.
- [ ] Restoring the planar model.
- [ ] Restoring GPU support for Adam.
- [ ] Restoring historical non-prototext model zoo examples.
- [ ] Prototext interface to explicitly manage weights.
- [ ] Fixing checkpointing.
- [ ] Clearing out deprecated code.


ndryden:
Not to say that it should block merging this, but how hard is it to restore L2 regularization? I worry a bit about breaking our AlexNet model.

timmoon10:
I think we should restore L2 regularization by implementing a better objective function framework. If we allow models to own multiple objective functions, L2 regularization can just be implemented as one. This is much less kludgy than our old approach of making it a part of learning layers and it avoids the problem of double-counting when a weight is shared amongst multiple layers.

bvanessen:
I don’t disagree that we should rework this mechanism, but I don’t want to break the entire code base. I understand the desire to merge sooner rather than later to minimize merge conflicts, but we need a working implementation.


Brian C. Van Essen

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Nov 17, 2017, at 4:59 PM, Tim Moon <notifications@github.com<mailto:notifications@github.com>> wrote:


I think we should restore L2 regularization by implementing a better objective function framework. If we allow models to own multiple objective functions, L2 regularization can just be implemented as one. This is much less kludgy than our old approach of making it a part of learning layers and it avoids the problem of double-counting when a weight is shared amongst multiple layers.

timmoon10:
Would it be reasonable to hard-code an L2 regularization factor of 0.0005 as a temporary kludge and then spend a day or two writing a proper implementation?

davidHysom:
"Would it be reasonable to hard-code an L2 regularization factor of 0.0005 as a temporary kludge and then spend a day or two writing a proper implementation?"


Um, what problem are you trying to solve? What's important enough that we need a temporary kludge?

timmoon10:
I agree it's really ugly, but I'd like to get these changes consolidated into our code base. Adding the 0.0005 L2 regularization factor makes sure these changes don't affect our AlexNet test bed.

PR #133, opened 2017-11-18
Evil hack to hard-code procs on each node

timmoon10:
Please do not merge this into develop.

### DIFF
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -891,6 +891,7 @@ void lbann_comm::rabenseifner_allreduce(
 
 void lbann_comm::setup_node_comm() {
 
+    #if 0
   // Get string specifying compute node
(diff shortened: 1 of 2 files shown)


bvanessen:
Thanks for the hack.  I appreciate it.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

bvanessen:
Well, that didn’t do what I needed.  I believe that that was forcing a distribution of two ranks per node, which I am able to get through the mpi runtime on bridges.  The problem that I am having is that both ranks on the node are attaching to both gpus simultaneously.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #136, opened 2017-11-28
Refactoring Siamese model.

timmoon10:
This pull request is intended to fix errors that were introduced in PR #132 and to take advantage of duplicated functionality in the DAG model class and weights class. In particular, Siamese networks are a type of DAG network with duplicated "heads" that share weights. Instead of maintaining multiple layer stacks, it suffices to construct and setup layers carefully during the setup phase. To make the terminology of "heads" easier to understand, I renamed the class from "planar" to "Siamese," although this can be reverted.


JaeseungYeom:
Looks fine by me. One suggestion is to think about a better name for set_layer_pointers() such as substitute_layer_pointers()

PR #142, opened 2017-12-06
Update Bamboo README

forsyth2:
Update the README in the Bamboo directory to be clearer.

The README can be previewed at https://github.com/forsyth2/lbann/tree/update-readme/bamboo.

### DIFF
--- a/bamboo/README.md
+++ b/bamboo/README.md
@@ -1,10 +1,10 @@
-# Plan - Nightly Develop
+# LBANN CI
 
-The Nightly Develop plan runs using the latest commits to LLNL/lbann/develop.
+Bamboo is the continuous integration (CI) framework we use. The Nightly Develop plan runs using the latest commits to `LLNL/lbann/develop` each night at midnight. The plan consists of two stages, detailed below.
 
 ## Stage - Build
 
-Checkout the latest develop and execute `./build_default_executable.sh`
+Checkout the latest develop and execute `./build_default_executable.sh`.
 
 ## Stage - Tasks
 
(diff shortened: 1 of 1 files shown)


forsyth2:
@DylanMcKinney Do these changes look good?

forsyth2:
@davidHysom Is there anything else you would like included in the README?

bamboo/README.md
```diff
@@ -1,10 +1,10 @@
-# Plan - Nightly Develop
+# Nightly Testing of Nightly Develop
```
DylanMcKinney: I think the title should maybe be something more along the lines of "LBANN CI" or something more general. Then maybe below we state that every test in this directory runs in our nightly plan

bamboo/README.md
```diff
@@ -18,17 +18,28 @@ Each job has the same basic structure:
 
 # Directory Structure
 
-'bamboo/compiler_tests', 'bamboo/integration_tests', 'bamboo/unit_tests' each have a 'conftest.py' that pytest requires. They also contain one or more python files. Each file has a number of tests to run. pytest recognizes methods that begin with `test_` as tests. Test methods should use the `assert` keyword.
+'bamboo/compiler_tests', 'bamboo/integration_tests', 'bamboo/unit_tests' each have a 'conftest.py' that pytest requires. They also contain one or more python files. Each file has a number of tests to run. 
+
+# Writing Your Own Tests
+
+The way we have Bamboo setup requires having tests use pytest. pytest recognizes methods that begin with `test_` as tests. Test methods should use the `assert` keyword.
```
DylanMcKinney: Worded a bit awkwardly. "Tests must be written using the python pytest module."

davidHysom: Hm, ""Tests must be written using the python pytest module" is meaningful only if you're familiar with pytest. IMO, the current wording is more informative. Though I think the last statement needs amplyifying. I think it means that if you write a test (a method that begins with test_) it will fail if an assert() within the test fails. Otherwise it passes -- or doesn't it also fail if it returns non-zero?

PR #147, opened 2017-12-14
Shared checkpoint support

DylanMcKinney:
This PR brings shared checkpoint functionality. Confirmed correct on LeNet with adam optimizer, but potentially still buggy elsewhere


DylanMcKinney:
Don't merge this yet. Incorrect number of epochs run after checkpoint restart. Figuring out why that's happening

PR #154, opened 2017-12-22
Objective function refactor

timmoon10:
This pull request adds functionality to compute separate statistics for each execution mode. This will be especially useful for LTFB.

It also adds the EvalType typedef, currently defined as a double. This is intended to provide a consistent datatype for objective function values, metrics, and timings that is independent of DataType.

### DIFF
--- a/include/lbann/base.hpp
+++ b/include/lbann/base.hpp
@@ -32,16 +32,22 @@
 #include "El.hpp"
 #include "lbann/Elemental_extensions.hpp"
 
+// Datatype for numerical computation
+// Default: float
 #if LBANN_DATATYPE == 8
 typedef double DataType;
 #elif LBANN_DATATYPE == 4
 typedef float DataType;
 #else
-// Default to floats
 #define LBANN_DATATYPE 4
 typedef float DataType; 
 #endif
 
+// Datatype for model evaluation
+// Examples: timing, metrics, objective functions
+typedef double EvalType;
+
+// Elemental matrices
 typedef El::Grid EGrid;
 typedef El::Grid Grid;
 typedef El::Matrix<DataType> Mat;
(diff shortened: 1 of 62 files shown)


davidHysom:
Are we going to continue to use typedefs, instead of "using?" I believe Jae-Seung changed some of his typedefs to using. Is there a consensus that we should be consistent? Or, does it really matter?

ndryden:
I would prefer `using`. It's the same as `typedef`, but (to me at least) more clear, and, it also supports templates (i.e. `template <typename T> using foo = bar<T>`).

timmoon10:
Would it be worthwhile changing DataType to a `using` as well?

ndryden:
If we want to clean up all the uses of `typedef`, then sure, but I wouldn't do it piecemeal unless you're touching the code already.

davidHysom:
If there's a consensus, I can attempt to use clang-tidy to modify our entire codebase (develop) to transition from typedefs to usings

bvanessen:
Yes let’s switch to using and the EvalType for metrics and objective functions.


Brian C. Van Essen

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Dec 22, 2017, at 3:18 PM, davidHysom <notifications@github.com<mailto:notifications@github.com>> wrote:


If there's a consensus, I can attempt to use clang-tidy to modify our entire codebase (develop) to transition from typedefs to usings

bvanessen:
Dave, do you want to run clang-tidy on the pull request before it gets merged, or have us merge it and then clean it up?

bvanessen:
Oh, nevermind, I reread the comments about typedef and using and I think that we should get this merged and then switch everything over from typedef to using.

davidHysom:
Thinking it's best if you merge, then I'll clean.

bvanessen:
Okay, I am downloading it now to build and if it compiles clean, I will merge it tonight.

PR #158, opened 2018-01-03
Checkpoint unit test 

DylanMcKinney:
Just a simple checkpoint unit test. Right now it only tests lenet restarted after 1 epoch, with 1 mpi rank. This PR shouldnt be merged just yet, as I am still trying to figure out why one of the saved parameter files is different after restarting. 

Completely unrelated, but this PR also has checkpoint code for rmsprop optimizer.


bamboo/unit_tests/test_checkpoint.py
```diff
@@ -0,0 +1,38 @@
+import os, pytest
+
+def test_checkpoint_lenet(exe, dirname):
+    return_code_nockpt = os.system('salloc -N1 srun -n1 %s --model=%s/model_zoo/tests/model_lenet_mnist_ckpt.prototext --optimizer=%s/model_zoo/optimizers/opt_sgd.prototext --reader=%s/model_zoo/data_readers/data_reader_mnist.prototext --num_epochs=2' % (exe, dirname, dirname, dirname))
```
forsyth2: I don't think `nockpt` is a great abbreviation. I see that it means "no checkpoint" but that is not immediately clear. "ckpt" isn't a great abbreviation of "checkpoint" either. It might be better to just not abbreviate.

bamboo/unit_tests/test_checkpoint.py
```diff
@@ -0,0 +1,38 @@
...
+    if return_code_nockpt != 0:
+        sys.stderr.write('LeNet execution failed, exiting with error')
+        sys.exit(1)
+    os.system('patch -f  %s/model_zoo/tests/model_lenet_mnist_ckpt.prototext %s/model_zoo/tests/checkpoint_dir.patch' % (dirname, dirname))
+    
+    return_code_ckpt_1 = os.system('salloc -N1 srun -n1 %s --model=%s/model_zoo/tests/model_lenet_mnist_ckpt.prototext --optimizer=%s/model_zoo/optimizers/opt_sgd.prototext --reader=%s/model_zoo/data_readers/data_reader_mnist.prototext --num_epochs=1' % (exe, dirname, dirname, dirname))
+
+    if return_code_ckpt_1 != 0:
+        sys.stderr.write('LeNet execution failed, exiting with error')
```
forsyth2: It might be a good idea to specify which LeNet failed for each of these writes - ie. was it `return_code_nockpt` or `return_code_nockpt` or `return_code_ckpt_2`?

DylanMcKinney: Good call. That would be especially important if it fails on attempted restart.

bvanessen:
I will work on debugging the differences with the train file once I merge this pull request into my refactoring of the I/O layers.

PR #161, opened 2018-01-12
KL Divergence

samadejacobs:
@timmoon10, I have committed a draft implementation of KL divergence, please review.
You will notice that the differentiate function is not implemented yet. I am thinking that  it would be neater  if the differentiate function takes the m_kl_loss computed at the evaluate phase and then return it to the base class in order to add it to the objective loss function gradient. Will it be possible for you to fix the base class to support this feature?
This will be similar to how evaluate is done:

EvalType evaluate() override;

float differentiate() override;

Also, we may want to reconsider renaming the weight_regularization folder but that can wait.

### DIFF
--- a/src/objective_functions/kl_divergence.cpp
+++ b/src/objective_functions/kl_divergence.cpp
@@ -34,53 +34,24 @@ namespace lbann {
 kl_divergence::kl_divergence(std::string layer1, std::string layer2)
                : objective_function_term(),
                  m_z_mean_layer_name(layer1), 
-                 m_z_log_sigma_layer_name(layer2), 
-                 m_kl_loss(EvalType(0)),
-                 m_target_layer(nullptr),
-                 m_gradient(nullptr) { } 
-
-kl_divergence::kl_divergence(const kl_divergence& other)
-  : objective_function_term(other),
-    m_gradient(other.m_gradient) {
-  if (m_gradient != nullptr) { m_gradient = m_gradient->Copy(); }
-}
-
-kl_divergence& kl_divergence::operator=(const kl_divergence& other) {
-  objective_function_term::operator=(other);
-  if (m_gradient != nullptr && other.m_gradient != nullptr
-      && m_gradient->DistData() == other.m_gradient->DistData()) {
-    El::Copy(*other.m_gradient, *m_gradient);
-  }
-  else {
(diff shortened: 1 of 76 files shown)


samadejacobs:
@timmoon10, diffferentiate() is implemented, code is now ready for review.

src/objective_functions/kl_divergence.cpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+}
+
+EvalType kl_divergence::evaluate() {
+
+  //Get matrices of input (latent space) layers
+  AbsDistMat& z_mean_acts = m_z_mean_layer->get_activations();
+  AbsDistMat& z_log_sigma_acts = m_z_log_sigma_layer->get_activations();
+  
+  AbsDistMat* z_mean = z_mean_acts.Construct(z_mean_acts.Grid(),
```
timmoon10: These matrices should be deallocated at the end of the function.

src/objective_functions/kl_divergence.cpp
```diff
@@ -0,0 +1,133 @@
...
+  El::Copy(z_log_sigma_acts,*z_log_sigma_gradient);
+
+  const DataType scale = DataType(1) / DataType(2)*height;
+  
+  //Compute z_log_sigma_gradient = -(1/2n)(1-exp(Z_logsigma))
+  auto expn  = [&](const DataType& x) {
+    return std::exp(x);
+  };
+  El::EntrywiseMap(*z_log_sigma_gradient,El::MakeFunction(expn));
+  El::Axpy(DataType(-1),*z_log_sigma_gradient, *z_log_sigma_gradient);
```
timmoon10: This will result in a matrix full of zeros. You should change the `expn` lambda to something like `return 1 - std::exp(x)`.

src/objective_functions/kl_divergence.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  //add gradients to respective error signal
+  //z_mean_gradient = (1/2n)z_mean 
+  m_z_mean_layer->add_to_error_signal(*z_mean_gradient, DataType(scale));
```
timmoon10: I think this should be scaled by `2*scale`.

PR #163, opened 2018-01-16
Refactored I/O layers

bvanessen:
Simplified the structure of the I/O layers and the data distribution classes.  Removed the multiple inheritance of the input and target layers.  The behavior of the data distribution classes is now contained in the io_buffer class, which is used by the new input and target layer to handle distributing data across the buffer.


davidHysom:
I didn't see any changes to anything in lbann/data_readers; were there any? I ask since this would affect the sharded readers (lbann/data_store/*) I'm developing.

bvanessen:
Nope, no changes there.  This was all confined to the I/O layers themselves and how they call the data readers.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #165, opened 2018-01-17
Fix Autoencoder Tests

forsyth2:
#160 was merged before it was finished. #164 puts the `develop` branch into the proper working state. This pull request is to finish the work from #160. 

Compiler Tests
- Add better logging.
- Update Bamboo to not be hard-coded to use Catalyst in `test_compilers.py`.

Integration Tests
- Refactor to allow multiple tests to use the same common code. This code is now found in `common_code.py`.
- Write tests for autoencoders (`model_conv_autoencoder_imagenet.prototext` and `model_conv_autoencoder_mnist.prototext`) using `common_code.py`.
- Added test of `mnist_partitioned_io` and `mnist_distributed_io` with `test_io_buffers.py`.
- Existing performance tests have been updated to use `common_code.py`.
- Added performance test for Alexnet with caching that will run weekly (`test_performance_cache_alexnet`).

Unit Tests
- Update commands in `unit_tests` to only use `salloc` if a node is not already allocated. Note that tests in `integration_tests` require specific `salloc` arguments and thus don't have this functionality.


bamboo/integration_tests/common_code.py
```diff
@@ -0,0 +1,154 @@
...
+        command = 'salloc -N 16 -p pbatch -t 600 srun -n 32 %s %s %s %s --num_epochs=20 --data_reader_percent=0.10 > %s' % (executable, model_option, reader_option, optimizer_option, output_file_name)
+    elif model_name == 'conv_autoencoder_mnist':
+        reader_option = get_reader_option(dir_name, 'mnist')
+        command = 'salloc -N 1 -p pbatch -t 60 srun -n 2 %s %s %s %s --num_epochs=5 > %s' % (executable, model_option, reader_option, optimizer_option, output_file_name)
+    elif model_name == 'conv_autoencoder_imagenet':
+        reader_option = get_reader_option(dir_name, 'imagenet')
+        command = 'salloc -N 16 -p pbatch -t 600 srun -n 32 %s %s %s %s --num_epochs=20 --data_reader_percent=0.10 > %s' % (executable, model_option, reader_option, optimizer_option, output_file_name)
+    elif model_name == 'lenet_mnist':
+        reader_option = get_reader_option(dir_name, 'mnist')
+        command = 'salloc -N 1 -p pbatch -t 60 srun -n 2 %s %s %s %s --num_epochs=5 > %s' % (executable, model_option, reader_option, optimizer_option, output_file_name)
```
DylanMcKinney: Let's change this partition to debug per Nikoli's suggestion. I reckon we can lower the time limit as well since this run takes like a minute total.

PR #166, opened 2018-01-19
Fixed a bug in the initialization of the io_buffer

bvanessen:


samadejacobs:
FYI, this doesn't fix crash with LTFB I reported.

PR #170, opened 2018-01-23
data readers for siamese models

JaeseungYeom:
1. data_reader_multi_images: a generic reader that inherits from 'imagenet_reader'
    - While 'm_image_list' is a vector of pairs of a file name string and integer label for imagenet,
      it is a vector of pairs of a vector of file names for multiple input patch files and a label.
      In short, 'vector<pair<string,int>>' becomes 'vector<pair<vector<string>, int>>'
    - The overridden 'fetch_datum()' copies each image referenced in 'vector<string>' into the
      corresponding view of the specified matrix column.
2. data_reader_mnist_siamese:
    - For ongoing development of Lenet-based siamese model
    - Pre-load all the data into memory as 'mnist_reader' does.
      Made the core part of 'mnist_reader::load()' into a separate reusable function.
    - There is no individual image file name per sample. All the images/labels are packed
       into a single binary file.
    - Randomly choose the paired input on-line. For this, the derived class maintains another
      data index list, 'm_shuffled_indices2'.
    - To make sure the secondary indices of the derived class gets shuffled whenever
      the primary ones of the base class get shuffled, the shuffling operation (the direct call to
      std::shuffle() on 'm_shuffled_indices') is wrapped into a virtual function 'shuffle_indices()'.  
      Then, data_reader's 'update()' will call the overridden 'shuffle_indices()' which shuffle the
      secondary indices as well as the primary.
    - 'fetch_datum()' and 'fetch_label()' take a pair of indices instead of a single index to 
      'm_shuffled_indices' of the base class.
    - The derived class overrides 'fetch_data()' and 'fetch_label()' to call the new overloaded 
      'fetch_datum()' and 'fetch_label()' interfaces. Another approach could embed the index 
      translation (from shuffle index to data index') into fetch_datum() and fetch_label(),
      making them truly overridable. However, this would break 'data_reader_merge_samples'.
3. data_reader_triplet
   - load three patches per sample. Use the patch image files generated off-line.
   - As the data list file is quite large itself in the ASCII text format, it is packed and loaded as
     a compressed NumPy file (*.npz)
   - A file name is compressed further by representing it as a sequence of indices to dictionaries.
   - There are two arrays for index sequences, two dictionaries (one for each array), and a label
     array.  For example, a file path train/n000111/abc.tag1.tag2.jpg would be represented as 
     'r[i][j][k]', 'v[i][j][x]', 'v[i][j][y]', 'v[i][j][z]' for the j-th patch of the i-th sample where 'r[i][j][k]' is 
     "train/n000111", and 'v[i][j][x]', 'v[i][j][y]' and 'v[i][j][z]' is "abc", "tag1", and "tag2" respectively.
     'r' is the root dictionary and 'v' is the variant dictionary.
   - Instead of uncompressing the list, list is kept as a compressed form, and uncompressed
     on-demand during execution.
   - The index sequence arrays are kept unchanged as CNPY data structures, but
     the rest are converted into std::vector while loading.
   - The dictionaries are arrays of character arrays. The space is allocate as a single buffer that
     accommodates N number of L characters, where L is the longest string in the original data,
     and N is the number of samples.
   - The label array is similar to the dictionary array. The longest entry is a two-digit number as
      the labels are from 0 to 19.
   - Overrides fetch_datum() and fetch_label() to expand file names on-demand and retrieve label
     from a label-only vector.


davidHysom:
All looks good to me. What you added for using the data_store should work. One comment on related code: I recently added this function:

 load_image(std::vector<unsigned char>& image_buf, ...)
  which is nearly identical to
load_image(const std::string& filename, ...)
These should be refactored to put repeated code in a single place (I haven't gotten around to that yet

JaeseungYeom:
I think import_image() probably does the same. I might be wrong, I will try and let you know when I get a chance. It basically takes stl vector or opencv mat, or any memory block wrapped (as a view) in opencv matrix structure

Issue #153, opened 2017-12-21
Layer refactoring for Hadamard layers


timmoon10:
Hadamard layers are needed for LSTMs but they break some of the assumptions baked into our code. In particular, each parent layer of a Hadamard layer sees a different error signal. The simplest generalization that fixes this problem is allowing each layer to have multiple activation and error signal matrices, one for each parent and child.

If we're going to dive into the guts of the layer class, I have a wishlist of other improvements:
- Modify the fp_compute and bp_compute functions so they pass in all needed parameters (input and output matrices, mini-batch size, execution mode) as arguments. This way it is easier to implement derived layer classes without detailed knowledge of the base class.
- Start using smart pointers. In particular, I'd like to keep distributed matrices in unique_ptr's.
- Remove the m_activations_v and m_error_signal_v matrices. It would be simpler to just reshape m_activations and m_error_signal when needed (typically only once per epoch).
- Better encapsulation of protected/private class members.

PR #171, opened 2018-01-23
Support for layers with multiple inputs and outputs

timmoon10:
This pull request addresses #153. Major changes:
- The number of error signal and previous activation matrices maintained by a layer is equal to the number of parent layers. Similarly, the number of activation and previous error signal matrices is equal to the number of child layers. They are accessed with the `get_prev_activations(int parent_index=0)` , `get_activations(int child_index=0)`, `get_prev_error_signals(int child_index=0)`, `get_error_signals(int parent_index=0)` functions.
- Layer input and output dimensions are now accessed with the `get_prev_neuron_dims(int parent_index=0)`, `get_neuron_dims(int child_index=0)`, `get_num_prev_neurons(int parent_index=0)`, and `get_num_neurons(int child_index=0)` functions. These are virtual functions.
- We no longer maintain matrix views for the active activation/error signal matrices. The data matrices are just resized to the proper dimensions. Elemental's memory manager should not reallocate memory when we do this.
- Layer back prop now adds to a layer's error signals instead of overwriting them. This is needed for activation regularization objective function terms like the KL divergence in variational autoencoders.
- Layer matrices are instantiated during the setup phase rather than in the constructors. This avoids the need to call `initialize_distributed_matrices()` in every derived class constructor.
- Added a `cudnn::matrix` class to handle memory allocation and views on GPUs. This is intended to be deprecated by GPU support in Hydrogen.
- Added a dummy layer that has a parent layer and no child layers. This is needed for our recurrent models.
- GPU implementation for the constant layer.
- The weights in the fully-connected layer are now called the "linearity" and the "bias."
- The CPU fully-connected layer now calls a purely local GEMM whenever the entire linearity matrix is entirely on a single process, i.e. the model-parallel case with one process or the data-parallel case.
- The GPU fully-connected layer no longer depends on cuDNN. It just uses cuBLAS and custom CUDA kernels.
- Removed preprocessor flags that make the GPU implementation for softmax layers optional.
- Renamed id_layer to identity_layer.
- Bug fixes to GPU workspace allocation. It is no longer possible to set a workspace size since the cuDNN manager attempts to maximize workspace size whenever possible.
- The `allreduce_on_gpus` routine can now accommodate arbitrary workspace sizes larger than 1KB.
- I've added comments to base.hpp that deprecate the typedefs DistMat, RowSumMat, ColSumStarVCMat, and ColSumMat in favor of MCMRMat, MCStarMat, VCStarMat, and MRStarMat, respectively. I prefer the latter since they use Elemental's matrix distribution notation consistently and since they don't imply any specific usage of the matrices.
- Many of the layer implementations use an internal variable name convention of input/output/gradient_wrt_output/gradient_wrt_input instead of prev_activations/activations/prev_error_signal/error_signal. I think this is more precise mathematically, but it can be confusing since the back prop input is the objective function gradient w.r.t. the forward prop output. If there is a consensus, the old variable name convention can be restored.

CPU and GPU gradient checks on Resnet-50 come back clean.


timmoon10:
Review requests:
- @bvanessen: the big OK, proper gradient scaling in the multi-model case, merge conflicts with #163.
- @ndryden: math sanity.
- @naoyam: GPU implementation of fully-connected and softmax layers and `cudnn_manager::allreduce_on_gpus`.
- @JaeseungYeom: slice and concatenation layers.

include/lbann/layers/regularizers/local_response_normalization.hpp
```diff
@@ -107,10 +101,10 @@ class local_response_normalization_layer : public regularizer_layer {
   #endif // LBANN_HAS_CUDNN
   }
 
-  ~local_response_normalization_layer() override {
+  virtual ~local_response_normalization_layer() override {
```
ndryden: Don't need virtual here.

include/lbann/layers/regularizers/dropout.hpp
```diff
@@ -65,15 +62,13 @@ class dropout : public regularizer_layer {
   dropout& operator=(const dropout& other) {
     regularizer_layer::operator=(other);
     m_keep_prob = other.m_keep_prob;
-    if(m_mask) {
-      delete m_mask;
-    }
+    if(m_mask) delete m_mask;
```
ndryden: Space, to be consistent.

bvanessen: I believe that in the code style section we were avoiding if statements with no braces.

JaeseungYeom:
A question on the third point, I understand that even if a matrix is initially allocated to the full minibatch size, and resized to a partial batch size in later iteration., the original allocation capacity remains the same. I am not sure if this behavior is guaranteed or simply a current implementation choice. Will it be still general if we decide to switch to some other framework than elemental. Finally, what happens if you copy construct from a resized matrix. Is the capacity of a new matrix will be same as the full capacity of the original one or limited to the resized one?

ndryden:
I believe it's a property of Elemental/Hydrogen; but our code bakes in a lot of assumptions about El/H anyway.

In any case, unless something is particularly poorly implemented, we should be resizing at most twice per epoch: once possibly for the last mini-batch; once possibly if we have something like variable mini-batch sizes. Even if we have to allocate new memory, the cost of that should be cheap relative to the cost of the epoch.

JaeseungYeom:
slice/concatenation layer behaviors as well as the code have not changed much other than ifdef/ifndef macro flips which seems reflected to the upstream early.

timmoon10:
Thinking more about this, I don't think it's a good idea to try emulating virtual templated functions in the layer classes. C++ currently doesn't support virtual templated functions, so we have been mimicking that functionality by having derived classes override the templated function and call it directly. This is an inelegant method because each derived class must call the function and the base class can't call the function itself, breaking encapsulation. We have been getting away with it so far since this idiom only shows up when instantiating matrices, which is rather simple. However, I think it would be a mistake to start implementing fp_compute and bp_compute as templated functions. Each derived class must overload these functions, so they would have to call them as well. They are called in the middle of the forward_prop and backward_prop methods, which are rather intricate and will require drastic implementation changes as we integrate with Hydrogen. Forcing each derived class to overload the forward_prop and backward_prop functions will introduce a ton of boilerplate and will make implementation changes a nightmare.

That being said, I agree that templated functions make the code a lot more manageable, especially in the face of a combinatorial explosion of data layouts and devices. Perhaps a good compromise is to rely heavily on templated helper functions within the derived classes. The virtual function would call the appropriate helper function with the help of a switch-case statement. This is essentially a more principled version of what we are currently doing (for example, the fp_compute in the fully-connected layer just calls fp_compute_cuda or fp_compute_cpu).

timmoon10:
Running the test_io_buffers unit test with a debug build and a `double` datatype, single models get identical results with different MPI process topologies and with both partitioned and distributed data readers. In the multi-model case, there are small discrepancies that are probably due to the way objective functions and metrics compute averages across models. This behavior seems identical to that in the develop branch.

include/lbann/base.hpp
```diff
@@ -33,26 +33,30 @@
...
+using MCMRMat    = El::DistMatrix<lbann::DataType, El::MC  , El::MR  >;
+using CircMat    = El::DistMatrix<lbann::DataType, El::CIRC, El::CIRC>;
+using StarMat    = El::DistMatrix<lbann::DataType, El::STAR, El::STAR>;
+using StarVCMat  = El::DistMatrix<lbann::DataType, El::STAR, El::VC  >;
+using VCStarMat  = El::DistMatrix<lbann::DataType, El::VC  , El::STAR>;
+using MCStarMat  = El::DistMatrix<lbann::DataType, El::MC  , El::STAR>;
+using MRStarMat  = El::DistMatrix<lbann::DataType, El::MR  , El::STAR>;
+using StarMRMat  = El::DistMatrix<lbann::DataType, El::STAR, El::MR  >;
+
+// Deprecated typedefs for Elemental matrices
```
timmoon10: These old typedefs are still functional, but I think they should be retired in favor of clearer names.

include/lbann/layers/activations/activation.hpp
```diff
@@ -32,45 +32,32 @@
...
-  }
+  /** Activation function.
+   *  This function is applied independently to each input entry.
+   */
+  virtual DataType activation(DataType x) const = 0;
 
- protected:
-  
-  virtual DataType activation_function(DataType x) = 0;
-  virtual DataType activation_function_gradient(DataType x) = 0;
```
timmoon10: This function should be the derivative of a single-variable function, so calling it a "gradient" is less clear.

include/lbann/layers/activations/softmax.hpp
```diff
@@ -34,411 +32,335 @@
...
-#include <typeindex>
 
 #include <cassert>
 
+// Output has minimum value to avoid denormalized floats
 #define LBANN_ENABLE_SOFTMAX_CUTOFF
 
 namespace lbann {
 
+#ifdef LBANN_HAS_CUDNN
```
timmoon10: Moving function prototypes from softmax_cuda.hpp allowed me to delete that file.

include/lbann/layers/activations/softmax_cuda.hpp
```diff
@@ -1,58 +0,0 @@
-////////////////////////////////////////////////////////////////////////////////
```
timmoon10: These function prototypes were moved to softmax.hpp.

include/lbann/layers/io/target/generic_target_layer.hpp
```diff
@@ -44,30 +44,51 @@ class generic_target_layer : public io_layer {
   generic_input_layer *m_paired_input_layer;
   generic_io_buffer *io_buffer;
 
+  /** Ground truth matrix. */
+  AbsDistMat* m_ground_truth;
```
bvanessen: Why has the generic _target_layer switched over to using a ground truth matrix rather than staying with the activations matrix.  I realize that the name is a bit clearer, so is the change just for that sake?

include/lbann/layers/activations/softmax.hpp
```diff
@@ -34,411 +32,335 @@
...
+      break;
+    case data_layout::DATA_PARALLEL:
+      m_workspace = new StarVCMat(grid);
+      break;
+    default:
+      std::stringstream err;
+      err << __FILE__ << " " << __LINE__ << " :: "
+          << "invalid distributed matrix layout";
+      throw lbann_exception(err.str());
+    }
```
bvanessen: Is this data allocation going to be updated to the new version that we discussed, using templated functions for each data layout.

include/lbann/layers/layer.hpp
```diff
@@ -209,182 +216,267 @@ class Layer {
...
+   *  Called by the setup function. The base method sets the previous
+   *  activation, activation, previous error signal, and error signal
+   *  matrices to zero matrices with the proper dimensions. Matrix
+   *  buffers are pinned if needed for GPU transfers.
    */
   virtual void setup_data();
+  /** Setup matrix views.
+   *  Called by the setup function. The base method does nothing.
+   */
+  virtual void setup_views() {}
```
bvanessen: Does this function still need to exist now that we have removed matrix views?

include/lbann/layers/io/target/generic_target_layer.hpp
timmoon10: Target layers don't output to any child layers, so they have no activation matrices. We need to maintain a matrix specifically for the ground truth data.

include/lbann/layers/layer.hpp
timmoon10: That's a good point. The setup_data function has subsumed its functionality.

include/lbann/layers/io/target/generic_target_layer.hpp
samadejacobs: Reconstruction layer is a target layer and could have child layer, I can think of greedy-layerwise autoencoder where reconstruction layers are placed and removed between layers.

timmoon10: I think that is more complicated that it needs to be. In my opinion, it would be clearer if we put split layers after the original layer and after the decoder layer and made them the parents of the reconstruction layer. This is also a nicer solution to #195.

include/lbann/layers/learning/base_convolution.hpp
```diff
@@ -593,107 +484,97 @@ class base_convolution_layer : public learning_layer {
     const DataType zero = DataType(0);
     const DataType one = DataType(1);
     const int num_gpus = this->m_cudnn->get_num_gpus();
-    const int mini_batch_size = this->m_model->get_current_mini_batch_size();
-    const int effective_mini_batch_size = this->m_model->get_effective_mini_batch_size();
```
bvanessen: Why has the effective mini-batch size been removed and substituted with just the mini-batch size?  We just fixed this so that it was using the effective mini-batch and not dividing by the number of models and the mini-batch size in two separate steps.

timmoon10: `mini_batch_size`  has been set to the effective mini-batch size. This seemed cleaner to me, but I can see how it could be obfuscating.

include/lbann/layers/regularizers/batch_normalization.hpp
```diff
@@ -537,90 +437,107 @@ class batch_normalization : public regularizer_layer {
...
+    const std::vector<DataType*> var_d = (is_training ?
+                                          m_var_d.get_locked_data() :
+                                          m_weights[3]->get_values_gpu());
 
     // Matrix parameters
-    const int effective_mini_batch_size = this->m_model->get_effective_mini_batch_size();
-    const int height = this->m_prev_activations_v->Height();
-    const int width = this->m_prev_activations_v->Width();
-    const int local_width = this->m_prev_activations_v->LocalWidth();
+    const int mini_batch_size = this->m_model->get_effective_mini_batch_size();
```
bvanessen: Please rename this to effective_mini_batch_size as well.

include/lbann/layers/regularizers/batch_normalization.hpp
```diff
@@ -744,65 +662,64 @@ class batch_normalization : public regularizer_layer {
...
+    auto& local_mean_gradient = m_mean_gradient->Matrix();
+    auto& local_var_gradient = m_var_gradient->Matrix();
+    auto& local_scale_gradient = m_scale_gradient->Matrix();
+    auto& local_bias_gradient = m_bias_gradient->Matrix();
+    
     // Matrix parameters
-    const int effective_mini_batch_size = this->m_model->get_effective_mini_batch_size();
-    const int width = this->m_prev_activations_v->Width();
-    const El::Int local_width = this->m_prev_activations_v->LocalWidth();
+    const int mini_batch_size = this->m_model->get_effective_mini_batch_size();
```
bvanessen: Please rename.

include/lbann/layers/transform/noise.hpp
```diff
@@ -87,68 +57,23 @@ class noise_layer : public transform_layer {
...
-    El::Copy(*this->m_prev_activations_v, *this->m_activations_v);
-    AbsDistMat* noise_mat = this->m_activations_v->Construct(this->m_activations_v->Grid(),
-                                                             this->m_activations_v->Root());
-    El::Gaussian(*noise_mat, this->m_activations_v->Height(), 
-                 this->m_activations_v->Width(),
-                 DataType(0), DataType(1));
-    El::Axpy(m_noise_factor,*noise_mat,*this->m_activations_v);
-    
-    //@todo - clip to min and max of input entry
-    auto clip = [&](const DataType& z) {
```
bvanessen: Did we intend to lose the clip function on the entries?

include/lbann/layers/transform/reshape.hpp
```diff
@@ -99,15 +89,15 @@ class reshape_layer : public transform_layer {
   }
 
   void fp_compute() override {
-    El::LockedView(*this->m_activations_v, *this->m_prev_activations_v);
+    El::LockedView(get_activations(), get_prev_activations());
   }
 
   void bp_compute() override {
-    El::LockedView(*this->m_error_signal_v, *this->m_prev_error_signal_v);
+    El::Axpy(DataType(1), get_prev_error_signals(), get_error_signals());
```
bvanessen: I have seen in several places where you use an Axpy instead of a copy.  Is the Axpy parallelized better than the copy?

include/lbann/layers/transform/slice.hpp
```diff
@@ -236,256 +198,281 @@ class slice_layer : public transform_layer {
...
+      const int output_slice_dim = m_slice_points[i+1] - m_slice_points[i];
+      const int output_region_stride = output_slice_dim * unit_region_size;
+
+      // Get position of first contiguous region in input tensor
+      const int input_region_start = m_slice_points[i] * unit_region_size;
+      const int input_region_end = m_slice_points[i+1] * unit_region_size;
+
+      // Populate current output tensor
+      output.Resize(num_regions * output_region_stride, width);
+      if (num_regions == 1) {
```
bvanessen: I believe that the num_regions trick being used here is the same thing as how we were using templated helper functions.  I still think that the num_regions is hard to remember when and why it works.

include/lbann/layers/transform/slice.hpp
```diff
@@ -236,256 +198,281 @@ class slice_layer : public transform_layer {
...
+
+      // Get stride between contiguous regions in output tensor slices
+      const int output_slice_dim = m_slice_points[i+1] - m_slice_points[i];
+      const int output_region_stride = output_slice_dim * unit_region_size;
+
+      // Get position of first contiguous region in input tensor
+      const int input_region_start = m_slice_points[i] * unit_region_size;
+
+      // Populate current output tensor
+      if (num_regions == 1) {
```
bvanessen: same as above.

src/io/data_buffers/distributed_io_buffer.cpp
```diff
@@ -72,7 +72,7 @@ void lbann::distributed_io_buffer::distribute_from_local_matrix(AbsDistMat& Ms,
           << " :: lbann_distributed_io_buffer: No valid data for this step -- local data was invalid";
       lbann_exception(err.str());
     }
-    CopyFromRoot(buf->M_local, buf->Ms);
+    CopyFromRoot(buf->M_local(El::ALL, El::IR(0, Ms.Width())), buf->Ms);
```
bvanessen: How does adding the explicit arguments help? Does it avoid a bounds check?

src/layers/activations/softmax.cu
```diff
@@ -91,44 +82,6 @@ void bp_cutoff(cudnn::cudnn_manager &cudnn,
     bp_cutoff_kernel<<<grid_dim, block_dim>>>(activations[i], error_signals[i],
                                               num_elms, min_output);
   }
-#endif
-}
-
-
-__global__ void bp_compute_cross_entropy_shortcut_kernel(const DataType *activations,
```
bvanessen: Do we no longer use this code?

src/layers/activations/softmax.cu
```diff
@@ -91,44 +82,6 @@ void bp_cutoff(cudnn::cudnn_manager &cudnn,
...
-#ifdef LBANN_ENABLE_SOFTMAX_CUTOFF_CUDA
-  DataType e = a > min_output ? a - p : DataType(0);
-#else
-  DataType e = a - p;
-#endif
-  error_signals[tid] = e;
-  return;
-}
-
-void bp_compute_cross_entropy_shortcut(cudnn::cudnn_manager &cudnn,
```
bvanessen: Same question here.

src/layers/learning/fully_connected.cu
```diff
@@ -1,145 +0,0 @@
...
-    sum += m[tid + i * h];
-  }
-  sum *= factor;
-  if (accum) sum += v[tid];
-  v[tid] = sum;
-}
-
-
-
-void row_sum(cudnn::cudnn_manager &cudnn,
```
bvanessen: Do we no longer need this code?

include/lbann/layers/io/target/generic_target_layer.hpp
bvanessen: I think that we should use the split layer to add in the layer-wise reconstruction layer, but I feel like we should talk through it.

include/lbann/layers/transform/reshape.hpp
timmoon10: Copying overwrites the error signal, disrupting possible contributions from activation regularization. This introduces a little extra overhead, but it would take some very ugly code to optimize it.

Alternatively, implementing #193 will make sure that the objective function only interacts with the error signal in an objective_function_layer. In this case, overwriting the error signal will be correct and we can perform copies and views instead of axpys.

include/lbann/layers/transform/slice.hpp
timmoon10: This is needed in case the slice dimension is not the first tensor dimension, e.g. we slice an image along its rows instead of along its channels.

src/io/data_buffers/distributed_io_buffer.cpp
timmoon10: `CopyFromRoot` resizes `buf->Ms` to have the same dimensions as `buf->M_local`. This caused the activations matrix to have the wrong dimensions at the last iteration in an epoch. A better approach would be to resize `buf->M_local` to match the current mini-batch step, but this patch was less intrusive.

src/layers/activations/softmax.cu
timmoon10: This functionality has been dead for a while. It greatly expanded the complexity of the code without delivering a measurable performance boost. If we want to implement this shortcut, I think it would be more sound to support a fused softmax-cross-entropy layer.

src/layers/learning/fully_connected.cu
timmoon10: I've replace our custom CUDA kernels with cuBLAS calls.

include/lbann/layers/transform/noise.hpp
timmoon10: This functionality is non-standard and makes the name "noise_layer" misleading. If we want this, we should implement a clip layer.

include/lbann/layers/learning/fully_connected.hpp
```diff
@@ -294,32 +215,33 @@ class fully_connected_layer : public learning_layer {
     }
 
     // Setup weights
-    if (this->get_data_layout() == data_layout::MODEL_PARALLEL) {
+    switch (get_data_layout()) {
```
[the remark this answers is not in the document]
timmoon10: It should be, but I think it would make the code messier. How about we duplicate the weights matrix by default and only use a special distribution for the model-parallel case?

src/layers/activations/softmax.cu
timmoon10: If we want to make it easier on the end-user, perhaps the model class can detect when a softmax layer outputs to cross entropy loss. It will then swap them out with a fused layer.

Issue #151, opened 2017-12-21
Exporting portable models


bvanessen:
Lets add support for exporting a trained model into a portable network format.  This should enable us to pre-train a network, export it, and allow someone to fine tune it in a different toolkit.

adammoody:
Haven't had a chance to look at it deeply, but ONNX seems to have good support so far:
https://research.fb.com/onnx-v1-released/

davidHysom:
This seems interesting ... but it seems so recent (early release: Sept of this year) that I would hesitate to commit much energy. If it's a going thing six months from now, maybe.


I think most or all of the interoperability API efforts are or will be python based.


Original post, Brian: "Lets add support for exporting a trained model into a portable network format"


Am thinking we shouldn't roll-our-own at this point in time, or jump on any bandwagon. Unless there's a motivating (political) need.

________________________________

bvanessen:
At the moment all I want is a numpy array per weight matrix that can be loaded into TensorFlow or something like that.


Brian C. Van Essen

vanessen1@llnl.gov<mailto:vanessen1@llnl.gov>

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Dec 21, 2017, at 5:12 PM, davidHysom <notifications@github.com<mailto:notifications@github.com>> wrote:

This seems interesting ... but it seems so recent (early release: Sept of this year) that I would hesitate to commit much energy. If it's a going thing six months from now, maybe.


I think most or all of the interoperability API efforts are or will be python based.


Original post, Brian: "Lets add support for exporting a trained model into a portable network format"


Am thinking we shouldn't roll-our-own at this point in time, or jump on any bandwagon. Unless there's a motivating (political) need.

________________________________

PR #173, opened 2018-01-26
Tensor representation of weights

timmoon10:
Weights can now be represented as tensors. Since they are stored internally as distributed matrices, some tensor dimensions correspond to the matrix height and the rest to the matrix width. This should make it easier to export trained weights into a format readable by Tensorflow (#151). I am not sure if it will affect weight checkpointing.

Currently, none of the layers initialize weights using this new feature because I want to minimize merge conflicts with #171. It will just require simple modifications to the `setup_data` function in the base convolution, convolution, deconvolution, and batch normalization layers.

### DIFF
--- a/include/lbann/weights/weights.hpp
+++ b/include/lbann/weights/weights.hpp
@@ -44,12 +44,13 @@ class optimizer;
 /** Neural network weights.
  *  Weights are tensors that act as trainable parameters for a neural
  *  network. The values can be initialized with a weights initializer
- *  and are optimized with first-order methods.
+ *  and are optimized with first-order methods (e.g. stochastic
+ *  gradient descent).
  *
- *  Internally, the weight values are stored in a distributed
- *  matrix. The matrix height dimensions are tensor dimensions that
- *  correspond to the matrix height. Similarly, the matrix width
- *  dimensions correspond to the matrix width.
+ *  Internally, the weight values are stored in a 2D distributed
+ *  matrix. The "matrix height dimensions" are tensor dimensions that
+ *  correspond to the matrix height. The remaining dimensions, the
+ *  "matrix width dimensions," correspond to the matrix width.
  *
  *  Note that LBANN weights are similar to Tensorflow variables and
  *  Caffe parameters.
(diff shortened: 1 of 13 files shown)


ndryden:
This should be helpful for things like stacking convolutions channel-wise in a concatenation layer. 👍

include/lbann/weights/weights.hpp
```diff
@@ -43,57 +41,79 @@ namespace lbann {
...
+                     std::vector<int> matrix_width_dims,
+                     El::Distribution col_dist,
+                     El::Distribution row_dist);
 
-  /** Get height of weights matrix. */
-  int get_height() const { return m_height; }
-  /** Get width of weights matrix. */
-  int get_width() const { return m_width; }
+  /** Get weight tensor dimensions. */
+  std::vector<int> get_dims() const;
```
samadejacobs: Could you please provide a note/example as part of documentation as to what each entry of the vector return  get_dims() (weight tensor dimensions) may represent e.g., dims[0] is N, dims[1] is channel etc?

timmoon10: I've expanded the documentation, although you should check if it is clear. Ultimately, this feature just implements a mathematical tensor and it is the responsibility of the layer class to interpret it.

samadejacobs:
The setup function (line 128 in weights.cpp) that takes tensor dimensions as input parameter doesn't work, that is if I replace: 
this->m_weights[0]->setup(m_kernel_size / this->m_neuron_dims[0],
                              this->m_neuron_dims[0],
                              El::STAR, El::STAR);
with 
this->m_weights[0]->setup(this->m_kernel_dims)

The code should behave the same but it crashes with: 
Process 0 caught error message:
Nonconformal Gemm(T/C)N
Which I believe may have to do with not setting up  the matrix dimensions properly.

timmoon10:
The convolution/deconvolution layers still expect the kernel matrix to have specific dimensions. When the internal implementation is changed, they should be more flexible.

As far as I am aware of, the only thing that's affected by the dimensions of the kernel matrix is quantization. Each matrix column is quantized separately. I think @ndryden had mixed results when quantizing convolution kernels, so this should be left as an expert mode feature.

PR #177, opened 2018-01-27
Model zoo cleanup

timmoon10:
This pull request removes all historical model zoo files that no longer compile and all experiment scripts that call them. The obsolete files can still be accessed by going into the repo history, but I think they were cluttering up the model zoo directory. I've also added an executable model zoo file that trains LeNet on MNIST.

This is probably overzealous, so discussion is encouraged!


davidHysom:
What is the purpose for having a separate non-prototext executable? Once we get rid of all 'historical' execs, we can delete the params class.

timmoon10:
This is mostly to provide a template in case we want to implement features in the future without implementing prototext support. The LeNet executable doesn't use the params class, but the cnn_mnist does. I would support deleting the cnn_mnist executable and the params class.

davidHysom:
Agreed: it makes sense to be able to try out/develop new features without immediately adding prototext support. This could shorten development time.

PR #179, opened 2018-01-30
Add -Werror

benson31:
adds an easy option for adding -Werror to the list of compiler flags.


benson31:
Note that this is only for C++ targets; Werror does NOT propagate to the CUDA targets.

PR #180, opened 2018-01-30
Minor clean up of a type mismatch in stack tracing code and new io_buffer unit test.

bvanessen:


timmoon10:
I think the io_buffer unit test might be a little overkill. It requires training a network 2x6x6 times to get all permutations of buffer layout, mini-batch size, and process topology.

PR #187, opened 2018-02-02
Update Documentation and add lbann.proto 

DylanMcKinney:
This PR adds documentation for lbann.proto using a doxygen filter tool I found online. With this addition I updated our documentation to populate our pages (callbacks, layers, metrics etc) with these lbann.proto fields. The pro to this change is that these pages will stay up to date automatically, whereas our current pages only get new changes when manually added. The con here is that since these pages (callbacks, layers etc..) are automatically generated, they may be more unorganized and potentially more confusing. 

So take a look at the current documentation: https://software.llnl.gov/lbann/ 
and my proposed new style: https://dylanmckinney.github.io/lbann/index.html 
and we can decide which works better. Or if perhaps we want some mix of the two. 

Its worth noting that part of what led me to make these changes is that the lbann.proto documentation is very buried in the class list if not linked by any pages. I think the lbann.proto documentation could be a good reference for new users trying to create their own prototext files, so I wanted to make them easier to find. This is a first pass so if anyone has suggestions to improve, or any changes they want to make themselves, feel free.


DylanMcKinney:
I think I have found a better way to improve our documentation. I am closing this PR and will introduce the potentially improved way during the 4/25 meeting.

PR #188, opened 2018-02-02
Consistent objective function average

timmoon10:
When computing the average objective function across multiple models, this PR weights the average such that each mini-batch sample contributes equally. With a `double` datatype and sequential initialization, we now get identical objective function values in the single model and multi-model cases (as long as the effective mini-batch is the same).

I'm not sure how this affects our integration tests. I've heard that the current reporting format can be a bit annoying since it changes depending on the single model or multi-model case. This could be a good time to develop a nicer format.

### DIFF
--- a/include/lbann/metrics/metric.hpp
+++ b/include/lbann/metrics/metric.hpp
@@ -114,8 +114,14 @@ class metric {
   /** Setup metric. */
   virtual void setup(model& m);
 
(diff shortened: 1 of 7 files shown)


bvanessen:
Dylan and Ryan, please note that this may change the absolute value of the objective function with multiple models.  CI scripts with specific values may need to be updated.

PR #189, opened 2018-02-03
Binary cross entropy can accommodate non-binary data

### LINKED ISSUE #186: Binary cross entropy on MNIST data
I have seen binary cross entropy loss function used in a couple of public blogs on MNIST dataset as a substitute for MSE (l[ink1](https://blog.keras.io/building-autoencoders-in-keras.html))([link2)](https://github.com/stitchfix/fauxtograph/issues/15). LBANN currently doesn't support this: it segfaults with error message: /usr/workspace/wsa/jacobs32/timmoon10-lbann-layerrf/src/objective_functions/loss_functions/binary_cross_entropy.cpp 56 :: binary cross entropy loss function requires binary ground truth.
We may want to revisit the formula for computing binary cross entropy with non-binary ground truth

timmoon10:
This closes #186. The ground truth and prediction values must still lie in the range [0,1]. In the debug build, we check whether input entries are valid.

### DIFF
--- a/src/objective_functions/loss_functions/binary_cross_entropy.cpp
+++ b/src/objective_functions/loss_functions/binary_cross_entropy.cpp
@@ -28,6 +28,46 @@
 
 namespace lbann {
 
+#ifdef LBANN_DEBUG
+namespace binary_cross_entropy_debug {
+  /** Check inputs for binary cross entropy.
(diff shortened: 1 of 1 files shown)


samadejacobs:
Reopening this because though exception is no longer thrown but learning quickly diverge and objective function returns "nan"

Issue #182, opened 2018-01-31
MPI error with multi-model imcomm


ndryden:
Running in a similar configuration to #178 except with 32 ranks and the (normal) imcomm callback, with 16 processes per model:

```
--------------------------------------------------------------------------------
[0] Epoch : stats formated [tr/v/te] iter/epoch = [716/0/160]
            global MB = [2048/   0/1024] global last MB = [2033/   0/ 113]
             local MB = [1024/   0/1024]  local last MB = [1017/   0/ 113]
--------------------------------------------------------------------------------
Fatal error in PMPI_Reduce_scatter_block: Message truncated, error stack:
PMPI_Reduce_scatter_block(1186).....: MPI_Reduce_scatter_block(sbuf=MPI_IN_PLACE, rbuf=0x38554c0, recvcount=16256, MPI_FLOAT, MPI_SUM, comm=0xc4000020) failed
MPIR_Reduce_scatter_block_impl(1042): 
MPIR_Reduce_scatter_block_intra(464): 
(unknown)(): Message truncated
```

Edit: Also note this switched the validation_percent to 0.

ndryden:
This appears to be caused by an issue with setting the right mini-batch size in the last mini-batch. The `input_layer` on rank 0 sets its mini-batch size (hence the input layer activations size) to be 1017 in `generic_input_layer::fp_set_std_matrix_view`. The other ranks have a mini-batch size of 1016. This is broadcast out, and so in `generic_input_layer::fp_compute`, the model's mini-batch size is set to be 1016. The activation size for the first layer are then set according to a mini-batch of size 1016.

Thus the first layer sees `m_matrix_weights_v` as 500x1023, `m_prev_activations_v` as 1023x1017, and `m_activations_v` as 500x1016. The reduce-scatter error comes from the GEMM, which is trying to receive into a matrix that's too small to hold the output.

@bvanessen Can you take a look at what the right way to handle this is? The I/O layers are a little too intricate for me to want to poke at them too much.

bvanessen:
Okay, the problem is that the reader associated with model 0 has one extra sample and thus sends more data than the other models expect.  I will work on a fix so that all ranks know the size of model 0's mini-batch size and setup their matrices appropriately.

bvanessen:
Fixed in f1a19d269fbb9d8eb8c72285e06286f11c1449bb

PR #196, opened 2018-02-07
Multiple bug fixes for the distributed I/O buffer related to Issue #182

bvanessen:
*Fixed a bug there the mini-batch size would be different for multiple
models on the last round when model 0 had an extra sample.

*Fixed a bug where the distributed I/O buffer always assumed that rank
0 of model 0 had the extra sample on a partial round.

*Fixed a bug where the combination of the base and model offset were
incorrect for multiple model when there was so little data that a
ranks only mini-batch was a partial mini-batch (and thus the lower
rank models did not complete a full mini-batch for that round either).

*Fixed a bug where a null pointer was dereferenced in the callback for
printing if there was no valid testing data.

*Changed the reporting for the number of mini-batches and size of the
last mini-batch to include the world remainder data if is owned by a
rank other than the world master.

### DIFF
--- a/include/lbann/base.hpp
+++ b/include/lbann/base.hpp
@@ -143,7 +143,7 @@ static void __attribute__((used)) _display_matrix(ElMat *m, const char *name) {
 #define log_msg(...) {\
   char str[256];\
   sprintf(str, __VA_ARGS__);\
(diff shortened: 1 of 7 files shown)


bvanessen:
This fixes the issue #182 but I noticed that the second model gets a NaN for the pearson correlation.  I think that the issue is that there is an extra column in the matrix that is zero and that is somehow messing up the calculation. @samadejacobs @ndryden Could either of you please look at it to see if there is an easy fix.

PR #197, opened 2018-02-07
Data store

davidHysom:


davidHysom:
Minor changes to data_reader.cpp and lbann.cpp; aside from that everything is new and/or refactored code in the data_store directories. I've tested data_store_imagenet with the alexnet on the smallish dataset, and appears to be working correctly.

include/lbann/data_store/generic_data_store.hpp
```diff
@@ -82,30 +79,40 @@ class generic_data_store {
     //m_cur_idx = 0;
   }
 
-  virtual void get_data_buf(std::string dir, std::string filename, std::vector<unsigned char> *&buf, int tid) = 0;
-
-  virtual void get_data_buf(int data_id, std::vector<unsigned char> *&buf, int tid) = 0; 
+  /// image data readers call this method
+  /// @todo: do we need tid?
```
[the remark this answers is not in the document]
ndryden: You can always just run `omp_get_thread_num` instead of passing the `tid` explicitly in the cases it's needed.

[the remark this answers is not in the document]
davidHysom: In the existing data_store_imagenet and data_store_multi_images the tid is not needed. I sort of doubt it will be needed in other data_stores, so, per Nikoli's suggestion, I'll remove it.

include/lbann/data_readers/data_reader_multi_images.hpp
```diff
@@ -90,6 +90,12 @@ class data_reader_multi_images : public imagenet_reader {
     return m_image_list.at(idx);
   }
 
+  /// The number of image sources or the number of siamese heads. e.g., 2;
+  /// this method is added to support data_store functionality
+  unsigned int get_num_img_srcs() {
```
JaeseungYeom: could you make this const method?

include/lbann/data_store/generic_data_store.hpp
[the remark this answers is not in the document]
bvanessen: sounds good.

JaeseungYeom:
In my opinion, the first argument of the setup() method should be always true. Thus, it does not need to be an argument. The setup function is not frequently called, and the check does not cost much. I see that, in other member functions, the check is omitted since it is supposed to be done in setup().
Also, data_reader_multi_images inherits imagenet_datareader.

compute_num_images() is not overriden for data_store_multi_images although it seems not used anywhere.
In addition, the implementation of the function and get_my_data_store_indices() can be improved. For example, you could have subtracted the rank from the total and divided it by the rank instead of looping with modulo check, and could have used m_rank as the stride.

davidHysom:
>the first argument of the setup()
bool test_dynamic_cast -- there was some reason I added that, but I think it went away when I refactored the code (and I can't now remember why it was needed)

> compute_num_images() is not overriden for data_store_multi_images although it seems not used anywhere.

compute_num_images() fills in the std::vector<int> m_num_images data structure; this is used in data_store_image::exchange_file_sizes. It does not need to be overridded in data_store_multi_images.

src/data_store/data_store_multi_images.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include "lbann/data_readers/data_reader_triplet.hpp"
+#include "lbann/data_readers/data_reader_multi_images.hpp"
+#include "lbann/utils/exception.hpp"
+#include "lbann/utils/options.hpp"
+#include "lbann/utils/timer.hpp"
+#include <sys/stat.h>
+
+namespace lbann {
+
+std::unordered_map<std::string, size_t> name_2_idx;
```
JaeseungYeom: Are these used anywhere?

JaeseungYeom:
Then, I think compute_num_images() is more like compute_num_samples()

src/data_store/data_store_multi_images.cpp
davidHysom: left over debug code; no longer used

davidHysom:
changed compute_num_images to compute_num_samples; no longer virtual; moved to data_store_image.cpp (and that was definitely a bug!)

davidHysom:
>In my opinion, the first argument of the setup() method should be always true. ...
>Also, data_reader_multi_images inherits imagenet_datareader.
Now I remember: data_store_images::setup() calls data_store_imagenet::setup() -- and we don't want the latter to try to cast the data_reader to imagenet_reader.

JaeseungYeom:
> Now I remember: data_store_images::setup() calls data_store_imagenet::setup() -- and we don't want the latter to try to cast the data_reader to imagenet_reader.

Where does this occur?

JaeseungYeom:
That line calls data_store_imagenet::setup() which does dynamic_cast<imagenet_reader*>(m_reader), where m_reader is a pointer to an object of data_reader_multi_images type. Since data_reader_multi_images class inherits imagenet_reader class, I believe that the cast should work. Did you get a nullptr there?

davidHysom:
that works ... forgot about the inheritance ...

davidHysom:
I've pushed changes that I believe address all comments from Jae-seung and Brian; also some additional refactoring and bug fixes. data_store_multi_image and data_store_imagenet are both working, and no longer segfault on surface.

src/data_store/data_store_multi_images.cpp
```diff
@@ -26,20 +26,17 @@
...
 #include "lbann/data_readers/data_reader_multi_images.hpp"
 #include "lbann/utils/exception.hpp"
 #include "lbann/utils/options.hpp"
 #include "lbann/utils/timer.hpp"
-#include <sys/stat.h>
 
 namespace lbann {
 
-std::unordered_map<std::string, size_t> name_2_idx;
 std::unordered_map<size_t, std::string> idx_2_name;
```
JaeseungYeom: name_2_idx and idx_2_name do not seem to be in use anymore. If this is the case, please remove them.

PR #203, opened 2018-02-08
Add 1/2 scale factor to L2 weight regularization term

timmoon10:
This changes the L2 regularization term to L2(w) = 0.5 * sum(w_i^2), which was our (implicit) convention before #132. Barring bugs, this will fix the performance regression we have been seeing with AlexNet on ILSCVR2012.

@forsyth2, @DylanMcKinney: this may cause a change in our regression tests.

### DIFF
--- a/src/objective_functions/weight_regularization/l2.cpp
+++ b/src/objective_functions/weight_regularization/l2.cpp
@@ -30,6 +30,37 @@
 #include "lbann/utils/cublas_wrapper.hpp"
 #endif // LBANN_HAS_CUDNN
 
+namespace {
+
+  /** Compute the entry-wise sum of squares of a local matrix. */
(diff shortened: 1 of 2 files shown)


src/objective_functions/weight_regularization/l2.cpp
```diff
@@ -30,6 +30,37 @@
...
+        for (El::Int i = 0; i < height; ++i) {
+          const EvalType val = buf[i + j*ldim];
+          sqsum += val * val;
+        }
+      }
+    }
+    return sqsum;
+  }
+
+} // namespace
```
timmoon10: This functionality used to be the private class function `local_squared_l2_norm`. That name is not as clear as `sum_of_squares` and I prefer to hide implementation details from the user.

PR #209, opened 2018-02-09
Aluminum build system integration

DylanMcKinney:
Links Aluminum in our build system and sets LBANN_HAS_ALUMINUM macro. Additionally adds very naive build script option --with-aluminum. When specified aluminum is added to the build if the environment variable ALUMINUM_DIR is set. I also added aluminum to the superbuild (i.e pull down and build aluminum) but this function will currently not work, because the repo is private.


CMakeLists.txt
```diff
@@ -183,10 +186,10 @@ if (LBANN_HAS_CUDA)
...
     if (NOT LBANN_HAS_NCCL2)
       message(FATAL_ERROR
-        "Requested LBANN_WITH_CNPY but CNPY not found. "
-        "CNPY is now disabled. "
-        "Try specifying CNPY_DIR as the root of a CNPY install. "
-        "Alternatively, build with LBANN_WITH_CNPY=OFF.")
+        "Requested LBANN_WITH_NCCL but NCCL not found. "
+        "NCCL is now disabled. "
+        "Try specifying NCCL_DIR as the root of a NCCL install. "
+        "Alternatively, build with LBANN_WITH_NCCL=OFF.")
```
benson31: Good catch!

cmake/modules/FindALUMINUM.cmake
```diff
@@ -0,0 +1,52 @@
+# Defines the following variables:
+#   - ALUMINUM_FOUND
+#   - ALUMINUM_LIBRARIES
+#   - ALUMINUM_INCLUDE_DIRS
+#
+# Also creates an imported target ALUMINUM
+
+# Find the header
+find_path(ALUMINUM_INCLUDE_DIRS common.h
```
benson31: I think we should look for "allreduce.hpp" instead. First, "common.h" is a common name; second, it shouldn't even be in the Aluminum project (nor should nccl.h, btw) since it's just the CUDA/NCCL_CHECK macros.

cmake/modules/FindALUMINUM.cmake
```diff
@@ -0,0 +1,52 @@
...
+  HINTS ${ALUMINUM_DIR} $ENV{ALUMINUM_DIR}
+  PATH_SUFFIXES include
+  NO_DEFAULT_PATH
+  DOC "Directory with ALUMINUM header.")
+find_path(ALUMINUM_INCLUDE_DIRS common.h)
+
+# Find the library
+find_library(ALUMINUM_LIBRARY liballreduce.so
```
benson31: Don't need the ".so" at the end or the "lib" at the front. CMake knows to add those, so you can just do ```find_library(ALUMINUM_LIBRARY allreduce)```

cmake/modules/FindALUMINUM.cmake
ndryden: Some of Al got refactored; nccl.h is gone.

superbuild/aluminum/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+  PREFIX ${CMAKE_CURRENT_BINARY_DIR}
+  TMP_DIR ${CMAKE_CURRENT_BINARY_DIR}/tmp
+  STAMP_DIR ${CMAKE_CURRENT_BINARY_DIR}/stamp
+  GIT_REPOSITORY ${ALUMINUM_URL}
+  GIT_TAG ${ALUMINUM_TAG}
+  SOURCE_DIR ${CMAKE_CURRENT_BINARY_DIR}/src
+  BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/build
+  INSTALL_DIR ${ALUMINUM_CMAKE_INSTALL_PREFIX}
+  BUILD_COMMAND  ${GNU_MAKE_PROGRAM} -j${ALUMINUM_MAKE_JOBS} 
+
```
benson31: This will require some updates because there's no configure step yet, but as per our discussion yesterday, I'm ok with not including aluminum in the superbuild until Aluminum is public.

Issue #207, opened 2018-02-09
proposal to remove generic_data_reader::swap_role()


JaeseungYeom:
I don't think we need this function.
The only place we use it is in the src/proto/proto_common.cpp
It is followed by use_unused_index_set(). We can change it to set_role().

The function does not make sense to me because `m_use_percent` and `m_validation_percent` are not compatible. In addition, `use_unused_index_set()` does not preserve the swapped indices anyway. `m_use_percent`, `m_validation_percent`, and `m_absolute_sample_count` should only be used in `select_subset_of_data()` once, and should not be used beyond that point. We can query the number of samples maintained by a reader by `get_num_data()`

davidHysom:
agree

JaeseungYeom:
I will create PR tomorrow if there is no objection until then.

PR #212, opened 2018-02-09
To remove swap_role()

JaeseungYeom:
To address the issue #207 
- change the only use of `swap_role()` to `set_role()`
- remove the `swap_role()` method
- change the `public` member methods for determining the number of samples to use to `protected` ones:
  `get_use_percent()`, `get_validation_percent()`, and `get_absolute_sample_count()`
However this depends on PR #208, which removes the external use of `get_use_percent()` under src/io/data_buffers. Those changes also show up here as I started new topic branch off from it.

### DIFF
--- a/include/lbann/data_readers/compound_data_reader.hpp
+++ b/include/lbann/data_readers/compound_data_reader.hpp
@@ -92,12 +92,14 @@ class generic_compound_data_reader : public generic_data_reader {
     }
   }
 
+/*
   void swap_role(std::string role) override {
     generic_data_reader::swap_role(role);
     for (auto&& reader : m_data_readers) {
       reader->swap_role(role);
     }
   }
+*/
 
   void set_master(bool m) override {
     generic_data_reader::set_master(m);
--- a/include/lbann/data_readers/compound_data_reader.hpp
+++ b/include/lbann/data_readers/compound_data_reader.hpp
@@ -92,15 +92,6 @@ class generic_compound_data_reader : public generic_data_reader {
     }
   }
 
-/*
-  void swap_role(std::string role) override {
-    generic_data_reader::swap_role(role);
-    for (auto&& reader : m_data_readers) {
-      reader->swap_role(role);
-    }
-  }
-*/
-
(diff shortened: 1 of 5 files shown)


include/lbann/data_readers/compound_data_reader.hpp
```diff
@@ -92,12 +92,14 @@ class generic_compound_data_reader : public generic_data_reader {
...
   }
 
+/*
   void swap_role(std::string role) override {
     generic_data_reader::swap_role(role);
     for (auto&& reader : m_data_readers) {
       reader->swap_role(role);
     }
   }
+*/
```
ndryden: Can you just delete this instead of commenting it out?

JaeseungYeom: Will do

src/proto/proto_common.cpp
```diff
@@ -2038,7 +2038,7 @@ void init_data_readers(bool master, const lbann_data::LbannPB& p, std::map<execu
         (*(ascii_reader *)reader_validation) = (*(ascii_reader *)reader);
       }
 
-      reader_validation->swap_role("validate");
+      reader_validation->set_role("validate");
```
bvanessen: I am not sure that what you are proposing here is correct.  The validation reader is a copy of the primary reader, but we then need to adjust the sample indices after the copy.

bvanessen:
Why was this merged?

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

JaeseungYeom:
Sorry, it was unintentional. I was trying to merge the other PR on which this was dependent on. I can undo it.

src/proto/proto_common.cpp
JaeseungYeom: By the way,  the adjustment of the sample indices is done by the following line reader_validation->use_unused_index_set() in a non-reversible way. What swap_role does is simply to swap m_use_percent and m_validation_percent, which is not a valid thing to do as these percents are not compatible.

PR #215, opened 2018-02-12
Vae icf

samadejacobs:
Minor refactor to noise and constant layer and then use them in vae prototext file.


model_zoo/models/autoencoder_mnist/vae_mnist.prototext
```diff
@@ -113,10 +122,30 @@ model {
...
+  #Hadamad1 zlogsigma_const05
+  layer {
+    parents: "const_05 z_log_sigma"
+    name: "const05_zlogsigma"
+    children: "exp"
+    data_layout: "model_parallel"
+    hadamard {
+    }
+  }
+
```
timmoon10: This can be replaced with a sum layer using a scale factor of 0.5.

PR #218, opened 2018-02-13
Parallelizing CPU-GPU memory transfers.

timmoon10:
This yields a ~20% performance boost with AlexNet.

@ndryden This PR also removes the summarizer callback from the AlexNet and Resnet-50 prototexts. The summarizer requires accessing matrix entries serially, so it shouldn't be enabled by default.


ndryden:
The code looks fine, but I'm confused by what you mean when you say the summarizer requires accessing matrix entries serially. Do you just mean that it's slow? (Which it is.)

PR #219, opened 2018-02-13
Checkpoint bug fix

DylanMcKinney:
This PR addresses the bug which leads a restarted LBANN run to differ from a non-restarted run. 

Notable changes:
- Checkpoint callback now saves certain parameters to new 'validate' file during on_validation_end callbacks. This fixed the issue where parameters we were saving on_epoch_end were changing during validation phase but not being updated. 
- Added save/reload for RNG state. This fixed the issue where shuffled indices differed after a reload, which caused all other parameters to differ in each subsequent epoch. 
- Refactored data reader save/load to use the pack/unpack headers style seen in the rest of the checkpoint code. 
-  Checkpoint unit test now passes (for the first time ever)

To be fixed:
- I have not added the 'on_validation_end' functionality to metric save/load yet, so that call is currently commented out. 
- I am going to take this opportunity to clean up some of the checkpoint code as well (remove commented out portions etc).

### DIFF
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -531,10 +531,10 @@ class generic_data_reader : public lbann_image_preprocessor {
   void use_unused_index_set();
 
   /** \brief Given directory to store checkpoint files, write state to file and add to number of bytes written */
-  bool saveToCheckpointShared(persist& p, const char *name);
+  bool save_to_checkpoint_shared(persist& p, const char *name);
 
   /** \brief Given directory to store checkpoint files, read state from file and add to number of bytes read */
-  bool loadFromCheckpointShared(persist& p, const char *name);
+  bool load_from_checkpoint_shared(persist& p, const char *name);
 
   struct packing_header {
     uint64_t mini_batch_size;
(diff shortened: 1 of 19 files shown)


include/lbann/data_readers/data_reader.hpp
```diff
@@ -531,11 +531,263 @@ class generic_data_reader : public lbann_image_preprocessor {
...
+    uint64_t sample_stride;
+    uint64_t iteration_stride;
+    uint64_t loaded_mini_batch_idx;
+    uint64_t reset_mini_batch_index;
+    uint64_t last_mini_batch_size;
+    uint64_t num_iterations_per_epoch;
+    uint64_t global_mini_batch_size;
+    uint64_t global_last_mini_batch_size;
+    uint64_t num_parallel_readers;
+    uint64_t model_rank;
```
bvanessen: I added a new field to the data reader in my last string of fixes. The field relates to the world model remained and needs to be stored.

bvanessen:
Once you update the data reader feel free to merge.

include/lbann/data_readers/data_reader.hpp
DylanMcKinney: added parameter 'world_master_mini_batch_adjustment' to save/load, and am merging now.

PR #220, opened 2018-02-14
Optimizer refactor

timmoon10:
This PR makes the following changes to the base optimizer class:
- GPU memory management is handled with GPU matrices.
- The GPU implementation is simpler since it keeps all gradient data on GPU.
- Added a function to launch an allreduce (possibly non-blocking) on the gradient staging matrix.
- Added flags to determine progress of an allreduce on the gradient staging matrix.
- Removed flags to determine if matrices are zero since they had no effect.

These changes provide a small performance boost to the GPU implementation since they remove some unnecessary calls to `El::Zero`. They will also make it easier to add a GPU allreduce that is non-blocking w.r.t. the GPU.

### DIFF
--- a/include/lbann/optimizers/optimizer.hpp
+++ b/include/lbann/optimizers/optimizer.hpp
@@ -41,16 +41,10 @@ namespace lbann {
 class optimizer {
  public:
 
-  /** Constructor. */
   optimizer(lbann_comm* comm, DataType learning_rate = DataType(0));
-
-  /** Copy constructor. */
   optimizer(const optimizer& other);
-  /** Copy assignment operator. */
   optimizer& operator=(const optimizer& other);
-  /** Destructor. */
   virtual ~optimizer();
-  /** Create a copy of the optimizer. */
   virtual optimizer* copy() const = 0;
 
   /** Get the optimizer name. */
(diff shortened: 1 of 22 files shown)


include/lbann/optimizers/optimizer.hpp
```diff
@@ -152,43 +153,40 @@ class optimizer {
...
+  /** GPU gradient matrix. */
+  cudnn::matrix m_gradient_d;
 #endif // LBANN_HAS_CUDNN
 
  private:
 
-  /** Whether the CPU gradient matrix is non-zero. */
-  bool m_cpu_gradient_is_nonzero;
-  /** Whether the CPU staging matrix is non-zero. */
-  bool m_cpu_staging_is_nonzero;
```
timmoon10: In the old implementation, we set these flags to avoid unnecessary CPU-GPU memory transfers. However, it's easier if the CPU implementation always keeps data on CPU and the GPU implementation on GPU.

src/callbacks/callback_imcomm.cpp
```diff
@@ -132,16 +132,20 @@ void lbann_callback_imcomm::on_epoch_end(model *m) {
...
-        reshape_mat(opt->get_gradient().Matrix(),
+        reshape_mat(gradient->Matrix(),
                     reshaped, params.reshape_height, params.reshape_width);
         local_gradients = &reshaped;
       } else {
-        local_gradients = &(opt->get_gradient().Matrix());
+        local_gradients = &(gradient->Matrix());
       }
       *local_gradients = params.error;
+      opt->add_to_gradient(*gradient);
```
timmoon10: The old implementation did not properly set the gradient if the data was on GPU.

src/utils/cudnn_wrapper.cpp
```diff
@@ -64,8 +64,8 @@ matrix::matrix(const matrix& other)
     m_is_locked(false) {
   if (m_cudnn != nullptr) {
     m_data.assign(m_cudnn->get_num_gpus(), nullptr);
+    copy(other);
   }
-  copy(other);
```
timmoon10: `copy()` fails if `m_cudnn` is not set.

src/optimizers/optimizer.cpp
```diff
@@ -318,94 +274,154 @@ void optimizer::add_to_gradient(const AbsDistMat& gradient,
...
+      m_cudnn->broadcast_to_gpus(gradient_d.get_data(),
+                                 gradient.LockedMatrix());
+      add_to_gradient(gradient_d, scale);
+      #endif // LBANN_HAS_CUDNN
+    }
   }
 }
-
-void optimizer::stage_gradient_for_accumulation(const AbsDistMat& gradient,
-                                                DataType scale) {
```
timmoon10: I switched the function order so that the CPU and GPU implementations of `add_to_gradient` and `stage_gradient_for_accumulation`/`add_to_gradient_staging` would be next to each other. Sorry if it makes it harder to diff the new and old versions.

src/callbacks/callback_imcomm.cpp
```diff
@@ -132,16 +132,20 @@ void lbann_callback_imcomm::on_epoch_end(model *m) {
     optimizer *opt = w->get_optimizer();
     if (ct_does_quantization(params.ct)) {
       comm->intermodel_sum_matrix(params.error);
+      opt->clear_gradient();
+      auto gradient = opt->get_gradient().Copy();
```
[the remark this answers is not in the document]
timmoon10: I've changed `optimizer::get_gradient()` to return a const reference to a matrix since things can get squirrelly if the data lives on GPU. In particular, someone could think they are changing the gradient by changing the CPU matrix, but it would actually have no effect. Once we have GPU support in Hydrogen, we'll no longer need to maintain separate matrices for CPU and GPU matrices and this concern will be moot.

PR #221, opened 2018-02-15
Early termination in back prop

timmoon10:
~~This PR builds on #220, so that should be accepted first.~~

Major changes:
- Back prop stops early if all required gradients have already been computed.
- We now handle the case where weights are being shared and optimizers use non-blocking allreduces.
- CPU layers keep all activations and error signals on CPU and GPU layers on GPU.

I get a ~10% performance boost since we no longer have to do a CPU-GPU data transfer during backprop in the input layer.

### DIFF
--- a/src/layers/layer.cpp
+++ b/src/layers/layer.cpp
@@ -247,55 +247,10 @@ void Layer::forward_prop() {
   // Setup matrix data, e.g. input matrices
   fp_setup_data(m_model->get_current_mini_batch_size());
 
-#ifdef LBANN_HAS_CUDNN
-  if(m_using_gpus) {
-
-    // Determine mini-batch size per GPU
-    const int num_gpus = m_cudnn->get_num_gpus();
-    const int local_mini_batch_size = (get_num_parents() > 0 ?
-                                       get_prev_activations().LocalWidth() :
-                                       get_activations().LocalWidth());
-    m_mini_batch_size_per_gpu = (local_mini_batch_size + num_gpus - 1) / num_gpus;
-
-    // Transfer inputs from CPU to GPUs if needed
-    for (int i = 0; i < get_num_parents(); ++i) {
-      const auto& parent = *m_parent_layers[i];
-      auto& prev_activations_d = m_prev_activations_d[i];
-      if (parent.using_gpus()) {
-        parent.get_gpu_fp_output(prev_activations_d, this);
-      } else {
-        prev_activations_d.resize(m_num_prev_neurons,
-                                  m_max_mini_batch_size_per_gpu);
-        m_cudnn->scatter_to_gpus(prev_activations_d.get_data(),
-                                 get_local_prev_activations(i),
-                                 m_mini_batch_size_per_gpu,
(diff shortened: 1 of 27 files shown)


src/layers/layer.cpp
```diff
@@ -247,99 +247,28 @@ void Layer::forward_prop() {
   // Setup matrix data, e.g. input matrices
   fp_setup_data(m_model->get_current_mini_batch_size());
 
-#ifdef LBANN_HAS_CUDNN
-  if(m_using_gpus) {
```
timmoon10: GPU matrices for previous activations are now setup in `fp_setup_data()`.

src/layers/layer.cpp
```diff
@@ -247,99 +247,28 @@ void Layer::forward_prop() {
...
-                               m_num_neurons :
-                               m_activations_d[0].get_leading_dim());
-    cudnn::set_tensor_cudnn_desc(m_prev_activations_cudnn_desc,
-                                 m_mini_batch_size_per_gpu,
-                                 m_prev_neuron_dims,
-                                 input_stride);
-    cudnn::set_tensor_cudnn_desc(m_activations_cudnn_desc,
-                                 m_mini_batch_size_per_gpu,
-                                 m_neuron_dims,
-                                 output_stride);
```
timmoon10: These descriptors are now initialized in `fp_setup_data()`.

src/layers/layer.cpp
```diff
@@ -1118,10 +1093,42 @@ void Layer::get_gpu_fp_output(cudnn::matrix& output_d,
...
+      El::Copy(activation, workspace);
+    }
+
+    // Initialize GPU matrix
+    auto&& cudnn = output_d.get_cudnn_manager();
+    const int num_gpus = cudnn->get_num_gpus();
+    const int num_processes = m_comm->get_procs_per_model();
+    const int max_local_mini_batch_size = (m_model->get_max_mini_batch_size() + num_processes - 1) / num_processes;
+    const int max_mini_batch_size_per_gpu = (max_local_mini_batch_size + num_gpus - 1) / num_gpus;
+    output_d.resize(get_num_neurons(child_index), max_mini_batch_size_per_gpu);
```
timmoon10: This should only allocate memory at the first iteration.

src/optimizers/optimizer.cpp
```diff
@@ -163,62 +139,56 @@ AbsDistMat& optimizer::get_gradient() {
...
-      m_comm->allreduce(*m_staging, m_staging->RedundantComm());
-    }
-#else
-    m_comm->allreduce(*m_staging, m_staging->RedundantComm());
-#endif  // LBANN_NBALLREDUCE_GRADIENT
-    add_to_gradient(*m_staging);
-    El::Zero(*m_staging);
-    m_cpu_staging_is_nonzero = false;
+  // Check if all gradient sources have made contributions
+  m_gradient_sources.erase(nullptr);
```
ndryden: What's with erasing `nullptr`?

timmoon10: Paranoia mostly.

PR #227, opened 2018-02-20
Mesh reader

ndryden:
This adds an initial implementation of a data reader for loading mesh data.

Loading the data seems to work, but LBANN isn't able to learn it, so I don't want to merge yet in case the issue turns out to be in this code.

@timmoon10 I do some "interesting" things in the `mesh_reader::fetch_datum`/`fetch_response`/`load_file` methods involving `El::Transpose` and views. It appears to work but I'd appreciate a second glance since you implemented `El::Transpose`.

### DIFF
--- a/src/data_readers/data_reader_mesh.cpp
+++ b/src/data_readers/data_reader_mesh.cpp
@@ -90,7 +90,8 @@ void mesh_reader::load_file(const std::string filename, Mat& mat) {
   if (std::is_same<float, DataType>::value) {
     // Need to transpose from row-major to column-major order.
     Mat tmp_mat(m_data_width, m_data_height, buf, m_data_width);
-    El::Transpose(tmp_mat, mat);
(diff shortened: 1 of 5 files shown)


src/data_readers/data_reader_mesh.cpp
```diff
@@ -0,0 +1,108 @@
...
+      El::IR(mb_idx, mb_idx + 1));
+    std::string filename = construct_filename(m_channels[i], data_id);
+    load_file(filename, X_view);
+  }
+  return true;
+}
+
+bool mesh_reader::fetch_response(Mat& Y, int data_id, int mb_idx, int tid) {
+  std::string filename = construct_filename(m_target_name, data_id);
+  Mat Y_view = El::View(Y, El::IR(0, Y.Height()), El::IR(mb_idx, mb_idx + 1));
```
timmoon10: `El::View(Y, El::ALL, El::IR(mb_idx))` is a bit more concise.

PR #228, opened 2018-02-20
Matrix testing

forsyth2:
Implement matrix testing - i.e. running variations of the following dimensions:
- Release or Debug
   - Updated compiler tests to allow for compiling with the debug flag (`-d`).
- Compilers: clang4, gcc4, gcc7, intel18
   - Updated output files to include compiler name, if necessary.
   - Updated integration tests `conftest.py` to handle multiple compilers.
   - Added new expected values files for different compilers, if necessary.
   - Updated performance tests to run with different compilers.
- Data Percentage / Frequency: 10% nightly, 100% weekly
- Clusters: Catalyst, Ray, Surface
   - Updated `build_default_executable.sh` to work with Surface.
   - Added function to determine proper command given a cluster's scheduler. See `tools.py`.
   - Updated compiler tests, integration tests, and unit tests to take into account the current cluster.
   - Updated output files to include cluster name, if necessary.
   - Added new expected values for different clusters, if necessary.

Other Improvements:
- Added script (`clean.sh`) to clean up generated files from previous builds.
- Created `output` directories for output files from compiler tests and integration tests.
- Updated existing expected values, if necessary.
- Renamed test functions to include `compiler`, `integration`, or `unit` since the new bamboo setup (running compiler tests, integration tests, and unit tests as tasks rather than jobs) does not separate these tests visually.
- Also renamed test files accordingly. Since Bamboo lists tests by file alphabetically, by putting `compiler`, `integration` or `unit` in the file name, similar tests will be grouped together.
- Updated integration tests checking equality to accept a tolerance factor. Now, rather than checking if the actual value is within `epsilon` of the expected value, we check if the actual value is within `tolerance_factor * expected_value` of the expected value.
- Fixed the default executable path in unit tests `conftest.py`.
- Updated `build_lbann.sh` to update Ray's spack options.

In order to better isolate work, future work will be split into multiple pull requests:
- Update unit tests to also run with the debug version of compilers.
- Update integration and unit tests to run on all compilers (currently only implemented for performance tests).
- Implement nightly testing with lower data percentage and weekly testing with higher data percentage for some large data-sets.
- Add permanent archive log files that will record how an actual value (say training run time or objective function) changes on each build.


bamboo/unit_tests/test_unit_lbann_invocation.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_unit_two_models_bad(cluster, exe, dirname) :
+    sys.stderr.write('TESTING: run lbann with two models but no optimizer or reader; lbann should throw exception\n')
+    if cluster in ['catalyst', 'surface']:
+        return_code = os.system('srun %s --exit_after_setup --model={prototext/model_mnist_simple_1.prototext,prototext/model_mnist_simple_1.prototext}' % (exe))
+    else:
+        raise Exception("Unsupported Cluster %s" % cluster)
+    assert return_code != 0
+
+def test_unit_two_models_bad2(cluster, exe dirname) :
```
DylanMcKinney: missing a comma between exe and dirname here

bamboo/unit_tests/test_unit_checkpoint.py
```diff
@@ -1,31 +1,40 @@
...
-    if cluster in ['catalyst', 'surface']:
-        return_code_nockpt = os.system('%s %s --model=%s/model_zoo/tests/model_lenet_mnist_ckpt.prototext --optimizer=%s/model_zoo/optimizers/opt_sgd.prototext --reader=%s/model_zoo/data_readers/data_reader_mnist.prototext --num_epochs=2' % (slurm_cmd, exe, dirname, dirname, dirname))
-    else:
-        raise Exception("Unsupported Cluster %s" % cluster)
+    command = tools.get_command(
+        cluster=cluster, executable=exe, num_nodes=1, num_processes=1,
+        dir_name=dirname,
+        data_filedir_ray='/p/gscratchr/brainusr/datasets/MNIST',
+        data_reader_name='mnist', model_folder='tests',
+        model_name='lenet_mnist_ckpt', num_epochs=2, optimizer_name='sgd')
```
DylanMcKinney: Think you accidentally removed the following line here:

"return_code_nockpt = os.system(command)"

bamboo/compiler_tests/conftest.py
```diff
@@ -2,9 +2,15 @@
```
DylanMcKinney: I think this configuration file needs the same lines from unit tests' which determine what cluster we are on.

bamboo/common_python/tools.py
```diff
@@ -0,0 +1,207 @@
...
+                raise Exception('Invalid usage: model_path is set but so are model folder and model_name')
+            elif model_folder != None:
+                raise Exception(
+                    'Invalid usage: model_path is set but so is model folder')
+            elif model_name != None:
+                raise Exception(
+                    'Invalid usage: model_path is set but so is model_name')
+        else:
+            if (model_folder != None) and (model_name != None):
+                option_model = ' --model=%s/model_zoo/models/%s/model_%s.prototext' % (dir_name, model_folder, model_name)
```
DylanMcKinney: prototext models that are in the 'tests' subdir are incorrect here. Generates model_zoo/models/tests/mnist_partitioned_io.prototext when it should generate model_zoo/tests/mnist_partitioned_io.prototext

bamboo/unit_tests/test_unit_checkpoint.py
forsyth2: Fixed.

bamboo/compiler_tests/conftest.py
forsyth2: Fixed.

DylanMcKinney: need to import re

PR #230, opened 2018-02-22
removing use_cudnn and replacing with disable_cuda, which defaults to…

davidHysom:
… false.

The models: autoencoder_cifar10/model_conv_autoencoder_cifar10.prototext
and autoencoder_mnist/model_conv_autoencoder_mnist.prototext contain:
  disable_cuda: true; remaining model files do not contain this field.
You can pass --disable_cuda or --disable_cuda=1 on the command line.
Logic was changed in model_zoo/lbann.cpp and model_zoo/lbann2.cpp for instantiating
cudnn::cudnn_manager


DylanMcKinney:
Looks good to me!

PR #235, opened 2018-02-28
Distributed Checkpoint support

DylanMcKinney:
This PR adds distributed checkpointing. Distributed checkpointing saves and restores matrices and model parameters on a per rank basis. If specified these parameters will be saved/loaded from node local ssd. Distributed checkpoint can be used alongside shared checkpointing. 

**Edit 04/18:**
The scope of this PR ballooned out a little bit. When working on distributed checkpoint I noticed quite a few bugs in shared. Those fixes are included here as well. With this pull request.

- Both checkpoint variants now fully work for the models I have tested (lenet and alexnet), when training with any optimizer, and using any checkpoint interval. 
- Checkpoint should also now be useable with non-sequential models. 
- The two ckpt variants can be used at the same time, and will restart from whichever of the two saved a checkpoint most recently. 
- Examples of the callback usage can be found in the model_zoo/tests directory 

With this pull request I also added support for loading weights saved during a checkpoint with lbann2. 
- To load weights using lbann2 point to a directory with the command line option --ckpt_dir=<checkpoint dir>.
- This will check layer/weight names from the newly setup model with the checkpoint weights, and load any matches for training. 

Aside from the functionality this PR includes some general clean-up/refactoring. 
I will squash these commits down before merging.


JaeseungYeom:
I found a bug in GPU data copying I checked in earlier. Some layer such as the fully connected layer does not use GPU even after allocating data arrays on GPU. The behavior is eventually controlled by m_using_gpus. So, in such a case, we should not copy GPU data to host for dumping states. I will check in a fix for this.

DylanMcKinney:
@JaeseungYeom Sounds good. If you can please check it into my ckpt-dist branch. I merged the lbann2 support into this branch so that it will be merged with this PR.

DylanMcKinney:
@JaeseungYeom , can you confirm these functionalities are working on GPU? I have tested CPU extensively, and remember you mentioning that you were testing on GPU. If those tests look good this PR should be ready to merge.

JaeseungYeom:
I will look. Today, I am a little occupied by attending a hackathone. I also wanted to ask about what happens if I use checkpointing schemes for both step and epoch together. I had some trouble after restarting step-based one. When it reached the end of epoch, it crashed.

JaeseungYeom:
I am also in the middle of updating lbann_comm broadcast interfaces (in other PR) so that it can be easier to replace native mpi calls for broadcasting.

JaeseungYeom:
Was it intentional to change optimizer's checkpoint interfaces private?

DylanMcKinney:
I will test checkpointing on step and epoch at the same time again. I had this tested and working, but may have done some refactoring which broke it. I will also transfer the mpi functions to use the lbann_comm versions.

bamboo/common_python/tools.py
```diff
@@ -17,8 +17,8 @@ def get_command(cluster, executable, num_nodes=None, partition=None,
                 exit_after_setup=False, mini_batch_size=None,
                 model_folder=None, model_name=None, model_path=None,
                 num_epochs=None, optimizer_name=None, optimizer_path=None,
-                processes_per_model=None, output_file_name=None,
-                error_file_name=None, return_tuple=False):
```
forsyth2: Don't remove `error_file_name`.

DylanMcKinney:
@bvanessen do you want to put eyes on this before I merge? I believe everything works as we have discussed it should, and it will not break anything.

JaeseungYeom:
I think there might be an issue when validation percent is set to 0% (i.e., disabled).
When restarting, it was looking for 'validate' file which does not exist, and crashed.

DylanMcKinney:
I forgot you could set validation percent to 0. I will make some safeguards surrounding that case

davidHysom:
Not too long ago I modified lbann proto to *not* instantiate a validation

reader if validation_percent is zero:

  if (readme.role() == "train" && readme.validation_percent() > 0.) {

This should no longer cause a crash.


DylanMcKinney:
I believe the cause of the crash is unrelated to that. I save certain parameters during a validation phase callback. If validation percent is set to 0 then these parameters do not get saved, which causes the restart to crash because it still attempts to load them.

JaeseungYeom:
As far as I can tell, the checkpointing is working not only for CPU-only platforms but also for GPU-enabled ones. I can confirm that Siamese model runs on CPU-only platforms are reproducible with/without checkpointing. However, base LBANN runs using GPUs are not reproducible, and making it difficult to verify checkpointing on GPU-enabled platforms. 
The test I did involves the steps as below:
- Save states from CPU-only runs
- Restart some middle states on a GPU-enabled platform.
- Right after restarting, checkpoint the states into a different directory
- Finally compare the *.bin state files on the original directory and the new directory.
The files were identical byte-to-byte. Restarting uploads data into GPU memories, and checkpointing downloads data from GPU memories.
The reproducibility issue has nothing to do with checkpointing.

ndryden:
cuDNN does not, in general, guarantee bit-level equality between runs for some of its operations by default, notably pooling. That's probably why you're seeing different results.

JaeseungYeom:
Thanks for confirming that. Do you remember how much variation in validation accuracy you see with any model like Alexnet?

ndryden:
I don't, off the top of my head-- maybe @timmoon10 does?

mcneish1:
In https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html, many operations (not sure about all of what we use) can use different algorithms or flags to make them deterministic. That still leaves the .cu files in the lbann source which may or may not be.

PR #236, opened 2018-03-03
Accumulating gradients with non-blocking GPU allreduces

timmoon10:
This pull request uses Aluminum's NCCL backend to overlap GPU computation (namely back prop) and communication (namely gradient accumulation). It also exposes non-blocking allreduce functions and request objects even when Aluminum is not included (in this case it will just call a blocking allreduce).


include/lbann/comm.hpp
```diff
@@ -30,14 +30,53 @@
...
+#ifdef LBANN_HAS_NCCL2
+#define ALUMINUM_HAS_NCCL
+#endif // LBANN_HAS_ALUMINUM
 #include <allreduce.hpp>
-#endif
+#endif // LBANN_HAS_ALUMINUM
 
 namespace lbann {
 
+namespace Al {
```
timmoon10: The `lbann::Al` namespace provides wrappers for various Aluminum functionality.

include/lbann/comm.hpp
```diff
@@ -832,6 +870,27 @@ class lbann_comm {
     return is_world_rank_on_node(world_rank);
   }
 
+  #ifdef LBANN_HAS_CUDA
```
timmoon10: These functions allow the cuDNN manager to send the GPU list and CUDA stream list to the LBANN communicator. This is admittedly a hacky, short-term solution. In a future refactor, I think it would be worthwhile to eliminate the cuDNN manager and have the communicator manage the various GPU objects.

include/lbann/comm.hpp
```diff
@@ -864,24 +923,24 @@ class lbann_comm {
...
-   */
-  std::unordered_map<MPI_Comm, allreduces::MPICommunicator*> al_comms;
-
-  /**
-   * Get the Aluminum communicator associated with Elemental communicator c.
-   * This will create a new Al communicator if needed.
-   * @todo This currently only supports the Aluminum MPI backend.
+  using al_comms_key_type = std::pair<MPI_Comm, std::type_index>;
+  using al_comms_val_type = std::unique_ptr<allreduces::MPICommunicator>;
+  std::map<al_comms_key_type, al_comms_val_type> m_al_comms;
```
timmoon10: Changed to a `map` since it is non-trivial to make an `unordered_map` use `pair` as keys.

include/lbann/comm.hpp
```diff
@@ -864,24 +923,24 @@ class lbann_comm {
    */
   int threads_per_proc;
   /** Pre-allocated buffers for collectives. */
-  std::unordered_map<size_t, std::vector<uint8_t *>> collective_bufs;
+  std::map<size_t, std::vector<uint8_t *>> collective_bufs;
```
timmoon10: `m_al_comms` was changed to a `map`, so I figured I might as well change this one too.

include/lbann/utils/cudnn_wrapper.hpp
```diff
@@ -422,7 +422,27 @@ class cudnn_manager {
   // One GPU per single thread of one MPI rank is assumed
   std::vector<ncclComm_t> m_nccl_comm;
   ncclDataType_t nccl_datatype();
-#endif // #ifdef LBANN_HAS_NCCL2
+#endif // LBANN_HAS_NCCL2
```
timmoon10: Don't pay attention to this, I forgot to delete it from an earlier version.

src/comm.cpp
```diff
@@ -86,9 +86,7 @@ lbann_comm::~lbann_comm() {
     }
   }
 #ifdef LBANN_HAS_ALUMINUM
-  for (auto&& c : al_comms) {
-    delete c.second;
-  }
+  m_al_comms.clear();
```
timmoon10: `m_al_comms` now contains smart pointers.

src/optimizers/optimizer.cpp
```diff
@@ -76,9 +76,7 @@ optimizer& optimizer::operator=(const optimizer& other) {
   m_gradient_allreduce_needed = other.m_gradient_allreduce_needed;
   m_gradient_allreduce_started = other.m_gradient_allreduce_started;
   m_gradient_allreduce_finished = other.m_gradient_allreduce_finished;
-  #ifdef LBANN_NBALLREDUCE_GRADIENT
```
timmoon10: We no longer check for the `LBANN_NBALLREDUCE_GRADIENT` macro. LBANN will always use non-blocking allreduces if they are available.

bvanessen:
I haven’t recently. I know that there is a nvcc issue that Dylan has a temporary fix, using an older cmake.


Brian C. Van Essen

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Mar 5, 2018, at 8:10 AM, Nikoli Dryden <notifications@github.com<mailto:notifications@github.com>> wrote:


I fixed a build issue with OpenMPI that showed up on Ray, but I'm also seeing linker errors. Have you managed to successfully build on Ray?

ndryden:
A normal LBANN build works fine on Ray for me. The issue seems to stem from linking in Aluminum.

ndryden:
More specifically, here's the linker issue I see:
```
[ 99%] Linking CXX executable lenet_mnist
/usr/bin/ld: lenet_mnist: hidden symbol `cudaGetDevice' in /usr/tce/packages/cuda/cuda-8.0/lib64/libcudart_static.a(libcudart_static.a.o) is referenced by DSO
/usr/bin/ld: final link failed: Bad value
```

I also see the same thing on Surface.

If I had to guess, I'd say it's due to a static version of libcudart being linked in, while liballreduce (the Aluminum library) needs some symbols from it.

ndryden:
It looks like one workaround is to modify the Aluminum makefile to link the CUDA and NCCL libraries into liballreduce.so.

timmoon10:
I have `/opt/cudatoolkit-8.0/lib64` in my `LD_LIBRARY_PATH`, but this is probably not the best solution.

timmoon10:
Here is a scaling study showing epoch times on ImageNet-1K:
surface_scaling

ndryden:
I am not seeing good performance from this on Ray. I can't tell if it's hanging or just running very slow. Using NCCL directly works fine, so I wonder if there's an issue with Aluminum or using it in a non-blocking way (with multiple calls outstanding, too).

bvanessen:
Interesting that data parallel is not scaling that well.  I assume that this is because it is using a fixed 256 sample mini-batch and that there isn’t enough work to do, correct?

Brian Van Essen
(w) 925-422-9300
(c) 925-290-5470

On Mar 5, 2018, at 2:16 PM, Tim Moon <notifications@github.com<mailto:notifications@github.com>> wrote:


Here is a scaling study showing epoch times on ImageNet-1K:
[surface_scaling]<https://user-images.githubusercontent.com/4406448/37002765-987462d4-207f-11e8-8eae-bc7a3bcf9082.png>

timmoon10:
I think that's right. I'd wager we get very good weak scaling, but we need algorithmic changes to get good strong scaling.

bvanessen:
That is fine that we don’t get better strong scaling with a fixed mini-batch size and data parallelism.  Once you hit the minimum number of samples per GPU to saturate, anything less is bad.

Brian Van Essen
(w) 925-422-9300
(c) 925-290-5470

On Mar 6, 2018, at 4:56 PM, Tim Moon <notifications@github.com<mailto:notifications@github.com>> wrote:


I think that's right. I'd wager we get very good weak scaling, but we need algorithmic changes to get good strong scaling.

naoyam:
@timmoon10 Do you have performance results on Surface with model parallel and Aluminum/NCCL?

timmoon10:
I don't have anything with AlexNet right now. For model-parallel Resnet-50 we have 31.3 sec/epoch and for data-parallel we have 25.2 sec/epoch (on Surface gpgpu, ImageNet-10, Aluminum+NCCL).

ndryden:
I tentatively think there's either a bug in Aluminum or NCCL that's resulting in the (lack of) performance I'm seeing on Ray. Since it's probably orthogonal to this implementation, :shipit: at your convenience.

PR #238, opened 2018-03-06
GPU debugging

timmoon10:
This pull request avoids making invalid cuBLAS GEMM calls when the local mini-batch size is zero, which should fix a runtime warning. It also implements a convenience function to attach GPU matrices to the GPU workspace buffers.


timmoon10:
This PR has not been tested or reviewed, but deadlines are deadlines @andy-yoo

naoyam:
Looks good to me.

PR #239, opened 2018-03-07
learning rate change schedule "poly"

JaeseungYeom:
Implements the learning rate change schedule poly as in Caffe.
- It updates the learning rate as "new_learning_rate = base_lr * (1 - cur_iter/max_iter)^p".
- 'p' and 'max_iter' are parameters specific to this schedule.
- We use 'm_cur_global_lr' as the base_lr. However, this is updated per epoch while the poly policy changes the learning rate per iteration.
- Therefore, 'optimizer_schedule' computes the learning rate as "m_cur_global_lr * (m_lr / m_last_epoch_lr)", where 'm_lr' is the learning rate for the new iteration computed as "(1 - cur_iter/max_iter)^p", and 'm_last_epoch_lr' is 'm_lr' at the end of the last epoch.


src/callbacks/callback_learning_rate.cpp
```diff
@@ -212,6 +213,50 @@ float lbann_callback_linear_growth_learning_rate::global_schedule(model *m) {
...
+}
+
+/**
+ * Keep the record of the learning rate at the end of the current epoch.
+ */
+float lbann_callback_poly_learning_rate::global_schedule(model *m) {
+  const float amt = m_lr / m_last_epoch_lr;
+  m_last_epoch_lr = m_lr;
+  return m_cur_global_lr * amt;
+}
```
timmoon10: This should be deleted.

src/callbacks/callback_learning_rate.cpp
```diff
@@ -212,6 +213,50 @@ float lbann_callback_linear_growth_learning_rate::global_schedule(model *m) {
...
+/**
+ * Compute the learning rate for the next iteration.
+ */
+float lbann_callback_poly_learning_rate::optimizer_schedule(model *m, optimizer &opt) {
+  const uint64_t cur_iter = static_cast<uint64_t>(m->get_cur_step());
+  if (m_max_iter > cur_iter) {
+    m_lr = static_cast<float>(pow(1.0 - static_cast<double>(cur_iter)/m_max_iter, m_p));
+  }
+  const float amt = m_lr / m_last_epoch_lr;
+  return m_cur_global_lr * amt;
```
timmoon10: We might run into numerical issues when `cur_iter/m_max_iter` is close to `1.0`. How about:
```
const auto iter = std::min(m->get_cur_step(), m_max_iter);
const double remaining_iters = m_max_iter - iter;
const auto scale = std::pow(remaining_iters/m_max_iter, m_p);
return m_cur_global_lr * scale;
```

include/lbann/callbacks/callback_learning_rate.hpp
```diff
@@ -229,6 +229,42 @@ class lbann_callback_linear_growth_learning_rate :
...
 
+/**
+ * Decrease the learning rate by polynomial policy
+ * base_lr*(1 - i_cur/i_max)^p, where
+ * base_lr is the initial learning rate, i_cur is the current iteration,
+ * i_max is the maximum iteration, and p is a parameter.
+ */
+class lbann_callback_poly_learning_rate : public lbann_callback_learning_rate {
+ public:
+  lbann_callback_poly_learning_rate(double p, uint64_t n_epochs, uint64_t max_iter);
```
timmoon10: We should document that `n_epochs` is not used when `max_iter` is non-zero.

include/lbann/callbacks/callback_learning_rate.hpp
```diff
@@ -229,6 +229,42 @@ class lbann_callback_linear_growth_learning_rate :
...
+  /// The exponent to compute new learning rate in poly policy
+  double m_p;
+  /// The number of epochs for training
+  uint64_t m_num_epochs;
+  /// The maximum number of iterations until which the learning rate changes
+  uint64_t m_max_iter;
+  /// The current rate to scale the base learning rate
+  float m_lr;
+  /// The learning rate scale used at the end of the last epoch
+  float m_last_epoch_lr;
```
timmoon10: `m_lr` and `m_last_epoch_lr` aren't needed if this review's changes are adopted.

src/callbacks/callback_learning_rate.cpp
```diff
@@ -29,6 +29,7 @@
 #include "lbann/callbacks/callback_learning_rate.hpp"
 #include <limits>
 #include <utility>
+#include <cmath> // pow
```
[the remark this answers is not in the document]
JaeseungYeom: I will change pow to std::pow. Both are actually the same under cmath.

src/callbacks/callback_learning_rate.cpp
JaeseungYeom: I am changing 1-iter/m_max_iter to (m_max_iter-iter)/m_max_iter.

include/lbann/callbacks/callback_learning_rate.hpp
JaeseungYeom: Adding some comments about it.

src/callbacks/callback_learning_rate.cpp
JaeseungYeom: One thing that was not clear to me was whether optimizers are supposed to operate exclusively.

src/callbacks/callback_learning_rate.cpp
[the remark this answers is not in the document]
timmoon10: I think we could get rid of the cmath header if we use `std::pow`. In any case, this is a nitpick.

src/callbacks/callback_learning_rate.cpp
timmoon10: It looks like they can all interact via the static variable `m_last_epoch_lr`. This interface seems very fragile. I'd prefer if each learning rate were independent, but @ndryden might be doing some fancy stuff with other learning rate schedules.

ndryden:
I think all the learning rate schedules are possibly more complicated than necessary. This is mainly driven by us not having a good way to compose learning rate schedules and handle the combination of "global" schedules (ones that apply one learning rate change to every layer) and "local" schedules (ones that apply a different learning rate to each layer).

PR #240, opened 2018-03-08
External data reader

mcneish1:
Work in progress.

TODO:
 * ~Finish Pilot2DataReader~
 * ~Grab by index instead of all data at once~
   * ~Support this for labels/responses in python side~
 * Port something with labels and something with responses to python, for verification
 * Clean up/documentation


mcneish1:
See #927

PR #242, opened 2018-03-10
Prototext parser refactor

timmoon10:
This refactor cleans up the prototext parser's handling of layers:
- The source code is moved to separate source files in the src/proto/factories directory.
- The layer factory function is templated by data layout to reduce code verbosity. If we template layers by device in the future, it will be easy to extend the layer factor function.
- This implementation has much better data encapsulation. In particular, it no longer relies on global variables.
- Added a utility function to parse space-separated lists.
- Added a macro to automatically add file and line number information to error messages.

I'm not sure if this is the right approach, so discussion is welcome. If it is, similar changes can be made to the prototext parser's handling of callbacks, objective functions, optimizers, weights, and models.


timmoon10:
I've expanded the pull request to include factory functions for callbacks, optimizers, objective functions, metrics, and models.

davidHysom:
I'd like your "#define  LBANN_ERROR(comm, message)" to be made more global. This is great, would same me and other much typing. Aside from that, let my know how I can clone your branch (I should know .. but I'm still fairly new to GIT).

timmoon10:
@davidHysom I think that would be a good idea. We should move it into base.hpp. Btw, do you know why we only threw exceptions on the master process? We could simplify the error macro if we threw exceptions on every process. It would also help catch problems that don't arise on all processes.

davidHysom:
I agree that all procs should throw exceptions. I just don't like all the printout when working with hundreds of processors. But, I always use the --stack_trace_to_file option. Unsure if that should be the default, but I'm in favor of it.

samadejacobs:
lbann2 will another good smoke test for prototext refactoring.

PR #253, opened 2018-03-22
Ds4triplet

JaeseungYeom:
data_store extension for data_reader_triplet.
- modification to data_reader_multi_images to make this extension easier while reusing the code as much as possible.  The requirement for data_reader_triplet is mostly the same as for data_reader_multiple_images.
  The only difference is the sample type, pair<vector<string>m, label_t>, returned by reader->get_sample(idx).  The label_t is int for multi_images and uint8_t for triplet. 
- addition of data_store_triplet inheriting data_store_multi_images
- override generic_data_reader::setup_data_store() for each derived class such that we do not need to include all the data_reader headers and associated data_store headers in the implementation file of the base class.
- data_store for triplet at least pass the data_store setup() but still crashes

### DIFF
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -761,13 +761,13 @@ class generic_data_reader : public lbann_image_preprocessor {
   }
   
   /// returns the data store, which may be a nullptr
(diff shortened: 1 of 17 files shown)


JaeseungYeom:
Merged into David's branch

PR #256, opened 2018-03-23
Debugging LTFB

timmoon10:
This fixes several problems I found in the LTFB callback:
- We used to get an infinite recursion under certain circumstances (models contain callbacks and the LTFB callback contains a model). To avoid this, LTFB maintains a list of weights rather than a model instance. We'll need to think of a workaround when we want LTFB with multiple model types.
- LTFB correctly resets the execution mode after evaluating the model on the validation set.
- The weights class synchronizes GPUs after CPU-GPU memory transfers.
- The effective mini-batch is only equal to the global mini-batch size when inter-model communication is enabled.
- We now throw an error if inter-model communication and LTFB are enabled at the same time.

LTFB on LeNet and AlexNet now appears to run reasonably.

### DIFF
--- a/include/lbann/layers/io/input/generic_input_layer.hpp
+++ b/include/lbann/layers/io/input/generic_input_layer.hpp
@@ -33,6 +33,7 @@
 #include "lbann/io/data_buffers/partitioned_io_buffer.hpp"
 #include "lbann/io/data_buffers/distributed_io_buffer.hpp"
 #include "lbann/models/model.hpp"
+#include "lbann/callbacks/callback_imcomm.hpp"
 
 namespace lbann {
 class generic_input_layer : public io_layer {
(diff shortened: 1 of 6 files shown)


include/lbann/layers/io/input/generic_input_layer.hpp
```diff
@@ -168,15 +169,20 @@ class generic_input_layer : public io_layer {
...
-    // batch size for the neural network
+    // Determine model mini-batch size and effective mini-batch size
+    // Note: If inter-model communication is activated, the effective
+    // mini-batch is equal to the global mini-batch size.
     mini_batch_size = get_current_mini_batch_size();
-    this->m_model->set_current_mini_batch_size(mini_batch_size);
+    int effective_mini_batch_size = mini_batch_size;
+    for (auto&& cb : this->m_model->get_callbacks()) {
+      if (dynamic_cast<lbann_callback_imcomm*>(cb) != nullptr) {
+        effective_mini_batch_size = get_current_global_mini_batch_size();
```
ndryden: Since there _should_ be only one instance of `lbann_callback_imcomm` in any model, we can `break` after finding it.

include/lbann/layers/io/input/generic_input_layer.hpp
```diff
@@ -168,15 +169,20 @@ class generic_input_layer : public io_layer {
    * averaged across models. */
   void fp_setup_data(int mini_batch_size) override {
 
-    // Use the predetermined size of the mini-batch to set the current
-    // batch size for the neural network
+    // Determine model mini-batch size and effective mini-batch size
+    // Note: If inter-model communication is activated, the effective
+    // mini-batch is equal to the global mini-batch size.
```
ndryden: Not a criticism of this, more a general comment: Do we need to deal with this elsewhere?

I feel like at least part of this comes down to us abusing the model replication framework for LTFB...

samadejacobs:
There may be a problem in how we are computing iteration per epoch, it seems global instead of per-model mini-batch size is used, setting appropriate iteration per epoch is needed for LTFB tournament (correctness).

samadejacobs:
The issue reported was fixed by setting data_set_per_model=true in input layer, thanks @bvanessen

PR #266, opened 2018-03-28
Optimizing GPU L2 regularization term.

timmoon10:
I am attempting to reduce the time spent in forward prop for the L2 regularization term on GPU. Changes:
- Using the cuBLAS `dot` routine instead of `nrm2`, which uses a numerically stable algorithm that is likely excessive for our use case.
- Running cuBLAS kernels asynchronously w.r.t. the host.
- Using all available GPUs.

### DIFF
--- a/include/lbann/objective_functions/weight_regularization/l2.hpp
+++ b/include/lbann/objective_functions/weight_regularization/l2.hpp
@@ -65,15 +65,17 @@ class l2_weight_regularization : public objective_function_term {
 
  private:
 
-  /** Reference to cuDNN manager. */
-  cudnn::cudnn_manager* m_cudnn = nullptr;
-
   /** Holds intermediate terms for each weights. */
   std::vector<EvalType> m_sqsums;
   /** Holds an allreduce request for each weights. */
   std::vector<Al::request> m_allreduce_reqs;
   /** Whether an allreduce is needed for each weights. */
   std::vector<bool> m_allreduce_started;
+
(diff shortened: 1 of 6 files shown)


include/lbann/objective_functions/weight_regularization/l2.hpp
```diff
@@ -64,12 +64,17 @@ class l2_weight_regularization : public objective_function_term {
   void compute_weight_regularization() override;
 
  private:
+
+  /** Reference to cuDNN manager. */
+  cudnn::cudnn_manager* m_cudnn = nullptr;
```
bvanessen: Shouldn't this be wrapped in #ifdef LBANN_HAS_CUDNN

src/objective_functions/weight_regularization/l2.cpp
```diff
@@ -88,51 +85,83 @@ void l2_weight_regularization::setup(model& m) {
   for (size_t i = 0; i < m_weights.size(); ++i) {
     m_allreduce_reqs.push_back(std::move(Al::request()));
   }
+
+  // Get cuDNN manager
+  m_cudnn = nullptr;
```
bvanessen: Shouldn't this be wrapped in #ifdef LBANN_HAS_CUDNN

ndryden:
It appears that L2 regularization, in particular, remains the primary component of forward and backward prop runtime for the objective function. We should look at optimizing this further.

ndryden:
This looks good now, :shipit: .

PR #268, opened 2018-03-30
Freezer

JaeseungYeom:
To avoid gradient computation and weight updates with pre-trained layers, provide 'freeze' interface.
- layers and weights get new freeze(), unfreeze(), and is_frozen() methods.
- 'freeze: true' in the prototext block of a layer freezes the layer.
- The default is unfrozen.
- When a layer gets frozen, it also freezes the weights associated with it.
- If a weight is shared, and one layer relying on it is frozen and another is not, throw exception.
- Given a frozen layer, freeze layers that precede it if that makes senses for the particular model, such as sequential or siamese. For othe models, users can manually control the behaivor by indicating whether to freeze each layer in the model description prototext.

### DIFF
--- a/src/layers/layer.cpp
+++ b/src/layers/layer.cpp
@@ -38,7 +38,8 @@ namespace lbann {
 
 Layer::Layer(lbann_comm *comm)
   : m_comm(comm),
-    m_cudnn(nullptr) {
+    m_cudnn(nullptr),
+    m_frozen(false) {
 
   // Initialize layer name
   static int num_layers = 0;
@@ -85,6 +86,7 @@ Layer::Layer(const Layer& other) :
   m_model(other.m_model),
   m_using_gpus(other.m_using_gpus),
   m_cudnn(other.m_cudnn),
+  m_frozen(other.m_frozen),
 #ifdef LBANN_HAS_CUDNN
   m_mini_batch_size_per_gpu(other.m_mini_batch_size_per_gpu),
   m_max_mini_batch_size_per_gpu(other.m_max_mini_batch_size_per_gpu),
@@ -145,6 +147,7 @@ Layer& Layer::operator=(const Layer& other) {
   m_model = other.m_model;
   m_using_gpus = other.m_using_gpus;
   m_cudnn = other.m_cudnn;
+  m_frozen = other.m_frozen;
 #ifdef LBANN_HAS_CUDNN
   m_mini_batch_size_per_gpu = other.m_mini_batch_size_per_gpu;
   m_max_mini_batch_size_per_gpu = other.m_max_mini_batch_size_per_gpu;
@@ -272,6 +275,9 @@ void Layer::forward_prop() {
 }
 
 void Layer::back_prop() {
+  if (m_frozen) {
+    return;
+  }
   const auto bp_start = get_time();
 
   // Setup matrix data, e.g. input matrices
@@ -302,6 +308,7 @@ void Layer::back_prop() {
 }
 
 bool Layer::update() {
+  if (m_frozen) { return true; }
   // Apply any updates.
   const auto update_compute_start = get_time();
   const auto layer_done = update_compute();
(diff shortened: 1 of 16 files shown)


JaeseungYeom:
I see that by having get_optimizre() to return nullptr, I can make the code simpler. I was thinking about some callback that dynamically unfreezes/freeze under certain conditions. To unfreeze the layer specific optimizer, I think it is easier to keep the object and maintain the state of optimizer constructed at prototext parsing. Thoughts?

timmoon10:
That's a good point. This approach could also be useful for training the generator and discriminator in GANs (@samadejacobs).

src/layers/layer.cpp
```diff
@@ -272,6 +275,9 @@ void Layer::forward_prop() {
 }
 
 void Layer::back_prop() {
+  if (m_frozen) {
```
timmoon10: I am not a fan of this. It forces all layers "below" a frozen layer to be frozen as well. There is already support to exit backprop early if possible, so it just constrains functionality.

JaeseungYeom: Without it, bp_compute() will still call gemms and there will be GPU synchronizations.
model's backward_prop() will still call the the back_prop() of all the layers below, in case that one has multiple layers above and not all of them are frozen.

timmoon10: If you look in `model::backward_prop()`, the model stops calling `layer::backward_prop()` when it has all the gradients it needs. For a simple linear network, this will only call one unnecessary GEMM.

In any case, we need the ability to backprop through frozen layers if we want to use this functionality for GANs.

JaeseungYeom: I think there are three solutions here.
1. back to the original version: put "if (l->is_frozen()) continue" into the top of the body of the for-loop in 'model::backward_prop()', and remove the "if (m_frozen) return" in 'Layer::back_prop()'.
2. initialize 'all_gradients_computed' to "!l->is_frozen()" instead of true.
3. remove "if (m_frozen) return" in 'Layer::back_prop()'

Both option 1 and option 2 do not call extra gemm or synchronization, and still allows back_prop of layers below a frozen layer.
The former is the most straight forward but skips 'do_layer_backward_prop_begin/end_cbs()'.
The latter is the opposite with some overhead.
Option 3 allows back_prop of layers below a frozen layer with extra gemm, which are in my view side effects to clean up without knowing what you have in mind for future extension.
My preference would be option 2.
Other requirement I see is that 'get_bp_output()' should zero the input matrix. For this purpose, 'Layer::add_to_error_signal()' on a frozen layer can simply zero the 'm_error_signals' instead of calling El::Axpy().

Opinions/objections?

timmoon10: I would lobby for option 3. If I'm interpreting this functionality correctly, freezing weights just means that we don't need to compute gradient updates. However, an unfrozen layer below a frozen layer still needs the frozen layer's error signal to correctly compute its own gradients. We need this functionality for GANs, where we might train a generator network that outputs into a frozen discriminator network. Terminating or zeroing out the error signal of frozen layers will be incorrect for GANs, although it does save a substantial amount of computation in the Siamese networks.

JaeseungYeom: In that case, solution 2 and 3 need to be combined to avoid early termination. agreed?

timmoon10: Option 2 terminates backprop once it encounters a frozen layer, which isn't the behavior we want. With option 3, early termination only happens when all required gradients have been computed.

JaeseungYeom: Probably the confusion is coming from "!l->". It looks like double el(l)s, but it is boolean not and an el. So, "!l->is_frozen()" tells if the layer 'l' is not frozen. it behaves the same with a non-frozen layer and avoids breaking the loop with a frozen layer. With option 3 alone, 'get_optimizer()' returns nullptr if a layer is frozen. This will not allow changing 'all_gradients_computed' to false.
To make freezing dynamic, we need to make sure that it is ok with callbacks. 
Currently, freezing is done before setting up callbacks, which is ok for now.

PR #272, opened 2018-04-04
Made additional change to use cuda version 9.1 by default and make it…

andy-yoo:
… compatible with

new NCCL installation.


andy-yoo:
Please take a look and if it is OK with you to use cuda 9.1 on surface. We will need to do similar update for ray, but it is after we get binaries for NCCL 2.1

Issue #193, opened 2018-02-06
A proposal for a new objective function paradigm


timmoon10:
As we move beyond classical problems like ImageNet classification, we are starting to develop a need for more problem-specific objective functions. We currently have to implement a new derived class for every objective function term, so experimenting with objective functions is slow, bug-prone, and contributes to feature bloat. It can also lead to confusing situations. For instance, our usage of KL divergence in variational autoencoders is non-standard and the implementation is completely different from that Keras and Tensorflow. [Explanation: Typically KL divergence is calculated between two distributions, e.g. probability vectors for "ground truth" and "prediction." However, VAEs compute the KL divergence caused by adding noise. The inputs are not probability vectors, but rather the noise distribution parameters.]

I think there is better way. Instead of building and maintaining these complicated objective_function_term objects, we should instead support a collection of layers to perform simple entry-wise operations and column-wise reductions. For instance, the KL divergence term in VAEs can be computed with exponentiation and addition (both of which we already support), followed by a column-wise average. This will result in a 1xN activations matrix where each entry is the KL divergence corresponding to a mini-batch sample. This can be output to an objective_function_layer that will be responsible for extracting the values and initializing gradients for backprop.

This proposal would eliminate objective_function_term classes corresponding to loss functions and activation regularization. However, it would be a good idea to implement layers for common loss functions like cross entropy or mean squared error.

bvanessen:
I think that in general this is a good idea.  I would like to walk through what changes to the code would be required for this, and how it tracks with current deadlines.

samadejacobs:
Here ([link](https://raw.githubusercontent.com/cdoersch/vae_tutorial/master/net_drawings/vae_train_net.png)) will be a good example of the approach I believe you are alluding to. If that is true, then it seems to me that we already have some of the building blocks except reduction layer and refactoring objective classes to become layers. Two observations: (1) ground_truth layer is more like what you will call data_layer (2) If my assumption is true, then this approach is more Caffe-ish than it is Keras/Tensorflow. In Caffe, layer is the first class candidate and almost every function is implemented as layer and it is seems to me that is approach you are proposing here.

PR #276, opened 2018-04-05
Evaluation layers

timmoon10:
This pull request makes progress on #193. Major changes:
- Implemented `evaluation_layer`, which computes an average value across a mini-batch.
- Implemented `layer_term`, which is an objective function term that interfaces with an evaluation layer.
- Reproduced the KL divergence term in the MNIST VAE model. This involved implementing `reduction_layer` and `power_layer`.

This MNIST VAE model passes a gradient check.

### DIFF
--- a/include/lbann/layers/activations/power.hpp
+++ b/include/lbann/layers/activations/power.hpp
@@ -1,0 +1,69 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2016, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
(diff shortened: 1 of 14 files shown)


bvanessen:
When you say this layer centric design, what is the alternative that you want a comparison too?

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

samadejacobs:
Layer-centric is a word Tim and I use a lot! IMHO, layer-centric approach is much of implementing most operations (including simple mathematical operations) as "Layers", this is the original Caffe's approach, TensorFlow and Keras are kind of different (at least from user's view point, it doesn't mean TF approach is best).  I am in between as to what is best, though I feel like "Layers" are heavyweight for some mathematical operations and verbose (in prototext design) but then I like the flow and easy of maintenance (among other pros/cons).

PR #279, opened 2018-04-06
Choose cuDNN convolution algorithm empirically.

timmoon10:
Instead of using cuDNN's algorithm heuristics, this pull request performs an exhaustive performance test every time a layer encounters an unfamiliar mini-batch size.

@ndryden has observed that cuDNN's heuristics often choose suboptimal algorithms. However, I find that this pull request actually harms performance. Perhaps the performance benchmark is not robust against variance in kernel runtime. If this is the case, we should run the benchmark several times.


ndryden:
If this isn't helpful, then let's shelve this this idea for now. We can revisit this PR in the future if this turns out to be needed.

PR #282, opened 2018-04-09
Fix ResNet50 max-pooling padding

ndryden:
The max-pooling layer in ResNet should have padding 1.

It looks like the [original Caffe implementation](https://github.com/KaimingHe/deep-residual-networks) used 0 padding, but more recent implementations ([TensorFlow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/learn/resnet.py), [PyTorch](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py)) use padding 1.

This also avoids shrinking the image unnecessarily, and avoids going from even to odd dimensions.

### DIFF
--- a/model_zoo/models/resnet50/model_resnet50.prototext
+++ b/model_zoo/models/resnet50/model_resnet50.prototext
@@ -92,7 +92,7 @@ model {
     pooling {
       num_dims: 2
       pool_dims_i: 3
-      pool_pads_i: 0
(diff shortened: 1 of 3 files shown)


timmoon10:
Does this change the dimension fed into the mean pooling layer at the end?

ndryden:
It should still be 7x7.

PR #286, opened 2018-04-13
specifying the weight initialization method for each layer of siamese model

JaeseungYeom:
prototext updates


timmoon10:
Looks good to me!

PR #287, opened 2018-04-13
Data reader csv

davidHysom:
This PR primarily modifies data_reader_csv::load so that only one processor reads the file,
then bcasts required info to others. Note that this has no affect/interaction
with data_store_merge_features -- I'm only trying to make the load() method
more efficient. The change reduces load() time by a factor of 5.5 (measured
on 40 cores on surface. 

Note: there are a couple of additional TODO items
in data_reader_csv.cpp that this PR does not address.

Note: I used  "raw" MPI calls since lbann::comm doesn't contain all needed functionality
(e.g, broadcasting a vector)

Note: if pilot1 data gets "really large," we should convert the CSV files
to some more-efficient binary format. When *not* using data_store_merge_features
this could significantly reduce runtime. If using data_store_merge_features in-memory-mode, such a change would only affect data_store setup() time (one time cost).

### DIFF
--- a/src/data_readers/data_reader_csv.cpp
+++ b/src/data_readers/data_reader_csv.cpp
@@ -97,18 +97,25 @@ csv_reader::~csv_reader() {
 }
 
 void csv_reader::load() {
-  int rank = m_comm->get_rank_in_world();
-  MPI_Comm world_comm = m_comm->get_world_comm().comm;
+  bool master = m_comm->am_world_master();
   setup_ifstreams();
   std::ifstream& ifs = *m_ifstreams[0];
+  const El::mpi::Comm world_comm = m_comm->get_world_comm();
   // Parse the header to determine how many columns there are.
   // Skip rows if needed.
-  if (!rank) {
+  if (master) {
     skip_rows(ifs, m_skip_rows);
-  } 
-  MPI_Bcast(&m_skip_rows, 1, MPI_INT, 0, world_comm); 
+    m_comm->broadcast<int>(0, m_skip_rows, world_comm);
+  } else {
+    m_skip_rows = m_comm->broadcast<int>(0, world_comm);
+  }
 
-  if (!rank) {
+  //This will be broadcast from root to other procs, and will
+  //then be converted to std::vector<int> m_labels; this is because
+  //El::mpi::Broadcast<std::streampos> doesn't work
+  std::vector<int> index;
+
+  if (master) {
     std::string line;
     std::streampos header_start = ifs.tellg();
     // TODO: Skip comment lines.
@@ -160,7 +167,7 @@ void csv_reader::load() {
     // TODO: Skip comment lines.
     // Used to count the number of label classes.
     std::unordered_set<int> label_classes;
-    m_index.push_back(ifs.tellg());
+    index.push_back(ifs.tellg());
     int line_num = 0;
     while (std::getline(ifs, line)) {
       ++line_num;
@@ -171,7 +178,7 @@ void csv_reader::load() {
           "csv_reader: line " + std::to_string(line_num) +
           " does not have right number of entries");
       }
-      m_index.push_back(ifs.tellg());
+      index.push_back(ifs.tellg());
       // Extract the label.
       if (!m_disable_labels) {
         size_t cur_pos = 0;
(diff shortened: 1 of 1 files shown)


src/data_readers/data_reader_csv.cpp
```diff
@@ -97,127 +97,157 @@ csv_reader::~csv_reader() {
...
-    if (!m_disable_labels) {
-      if (m_label_col < 0) {
-        // Last column becomes the label column.
-        m_label_col = m_num_cols - 1;
-      }
-      if (m_label_col >= m_num_cols) {
+  if (!rank) {
+    skip_rows(ifs, m_skip_rows);
+  } 
+  MPI_Bcast(&m_skip_rows, 1, MPI_INT, 0, world_comm);
```
ndryden: You should be able to use the existing `lbann_comm::broadcast` method here, which supports scalars.

src/data_readers/data_reader_csv.cpp
```diff
@@ -97,127 +97,157 @@ csv_reader::~csv_reader() {
 }
 
 void csv_reader::load() {
+  int rank = m_comm->get_rank_in_world();
```
ndryden: It would be cleaner to use `m_comm->am_world_master`, I think.

src/data_readers/data_reader_csv.cpp
```diff
@@ -97,127 +97,157 @@ csv_reader::~csv_reader() {
...
+        throw lbann_exception(
+          "csv_reader: label classes are not contiguous");
+      }
+      m_num_labels = label_classes.size();
     }
-    m_num_labels = label_classes.size();
+    ifs.clear();
+  } // if (!rank)
+
+  MPI_Bcast(&m_num_cols, 1, MPI_INT, 0, world_comm);
```
ndryden: `lbann_comm::broadcast`

src/data_readers/data_reader_csv.cpp
```diff
@@ -97,127 +97,157 @@ csv_reader::~csv_reader() {
...
-    m_num_labels = label_classes.size();
+    ifs.clear();
+  } // if (!rank)
+
+  MPI_Bcast(&m_num_cols, 1, MPI_INT, 0, world_comm); 
+  m_label_col = m_num_cols - 1;
+
+  //bcast the index vector
+  int index_size = m_index.size(); 
+  MPI_Bcast(&index_size, 1, MPI_INT, 0, world_comm);
```
ndryden: `lbann_comm::broadcast`

src/data_readers/data_reader_csv.cpp
```diff
@@ -97,127 +97,157 @@ csv_reader::~csv_reader() {
...
+  } // if (!rank)
+
+  MPI_Bcast(&m_num_cols, 1, MPI_INT, 0, world_comm); 
+  m_label_col = m_num_cols - 1;
+
+  //bcast the index vector
+  int index_size = m_index.size(); 
+  MPI_Bcast(&index_size, 1, MPI_INT, 0, world_comm); 
+  m_index.resize(index_size);
+  MPI_Bcast(m_index.data(), index_size*sizeof(std::streampos), MPI_BYTE, 0, world_comm);
```
ndryden: I think it would be cleaner to add a scalar-array version of `broadcast` to `lbann_comm` than to use `MPI_Bcast` directly.

src/data_readers/data_reader_csv.cpp
```diff
@@ -97,127 +97,157 @@ csv_reader::~csv_reader() {
...
+  MPI_Bcast(&index_size, 1, MPI_INT, 0, world_comm); 
+  m_index.resize(index_size);
+  MPI_Bcast(m_index.data(), index_size*sizeof(std::streampos), MPI_BYTE, 0, world_comm); 
+  m_num_samples = index_size - 1;
+
+  //optionally bcast the response vector
+  if (!m_disable_responses) {
+    m_response_col = m_num_cols - 1;
+    m_responses.resize(index_size - 1);
+    MPI_Bcast(m_responses.data(), (index_size - 1)*sizeof(DataType), MPI_BYTE, 0, world_comm);
```
ndryden: Update this also when we have the right `broadcast` support.

src/data_readers/data_reader_csv.cpp
```diff
@@ -97,127 +97,157 @@ csv_reader::~csv_reader() {
...
-  m_num_samples = m_index.size() - 1;
-  // End of file offset.
-  m_index.push_back(ifs.tellg());
-  // Seek back to the beginning.
-  ifs.clear();
-  ifs.seekg(0, std::ios::beg);
+
+  //optionally bcast the label vector
+  if (!m_disable_labels) {
+    MPI_Bcast(&m_num_labels, 1, MPI_INT, 0, world_comm);
```
ndryden: `lbann_comm::broadcast`

src/data_readers/data_reader_csv.cpp
```diff
@@ -97,127 +97,157 @@ csv_reader::~csv_reader() {
...
-  m_index.push_back(ifs.tellg());
-  // Seek back to the beginning.
-  ifs.clear();
-  ifs.seekg(0, std::ios::beg);
+
+  //optionally bcast the label vector
+  if (!m_disable_labels) {
+    MPI_Bcast(&m_num_labels, 1, MPI_INT, 0, world_comm); 
+    m_labels.resize(m_num_samples);
+    MPI_Bcast(m_labels.data(), m_num_samples, MPI_INT, 0, world_comm);
```
ndryden: Ditto.

davidHysom:
OK, I'll change the code to use lbann:comm. But I'm wondering what the argument is against using raw MPI? There's actually several places in our codebase (not all written by me) that uses MPI calls.

ndryden:
Yeah, I've tried to remove most of them, but some remain. The main reason to route everything through `lbann_comm` is so that it's easier to change how the communication is done and have everything in LBANN automatically take advantage of it. (E.g. switching from MPI to NCCL.)

PR #295, opened 2018-04-18
make broadcast more general

JaeseungYeom:
Make broadcast more general such that we can call it on a flat data structure or a memory buffer.
This will make checking point structure handling simpler.

### DIFF
--- a/src/data_readers/data_reader_csv.cpp
+++ b/src/data_readers/data_reader_csv.cpp
@@ -105,10 +105,8 @@ void csv_reader::load() {
   // Skip rows if needed.
   if (master) {
     skip_rows(ifs, m_skip_rows);
-    m_comm->broadcast<int>(0, m_skip_rows, world_comm);
-  } else {
-    m_skip_rows = m_comm->broadcast<int>(0, world_comm);
   }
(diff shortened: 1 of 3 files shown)


src/data_readers/data_reader_csv.cpp
```diff
@@ -105,10 +105,8 @@ void csv_reader::load() {
   // Skip rows if needed.
   if (master) {
     skip_rows(ifs, m_skip_rows);
-    m_comm->broadcast<int>(0, m_skip_rows, world_comm);
-  } else {
-    m_skip_rows = m_comm->broadcast<int>(0, world_comm);
   }
+  m_comm->broadcast<int>(0, m_skip_rows, world_comm);
```
davidHysom: I think this is much cleaner than having separate calls for root and other procs. Possibly add  world_broadcast and model_broadcast, so you don't have to pass in a communicator (similar to what I did for bcasting vectors)

JaeseungYeom: We have similar but I will make the interfaces consistent by making the arguments pass-by-reference.
However, I am not sure if it is ok to use the pointer to such variable as is or should I copy-construct a heap-allocated variable and pass the pointer to MPI calls.

PR #302, opened 2018-04-25
Data store comms

davidHysom:
mostly: replacing native MPI calls with lbann::comm calls.


davidHysom:
Thanks. I already did the merge, but will make corrections per your comments.

I think lbann::comm is getting confusing: we have very many version of, e.g, broadcast.

Possibly we should simplify? Possibly get rid of all model_XX and world_XX ??

PR #250, opened 2018-03-20
Feature hydrogen with GPU device Support

bvanessen:
I have integrated the branch of Hydrogen with GPU support into LBANN.  Where possible classes have been properly templated to handle either GPU or CPU allocated Elemental matrices.

Specifically, the layer classes are setup for either allocation.  However, there is no GPU specific code for the fully connected layers and the convolutional layers still use their internal matrix definitions.

Additionally, the I/O layers and call back layers are intended for CPU allocations at the time being.  We need to see if some of them should be allowed to have GPU allocations.

The initial goal of this PR is to have the same CPU-centric execution with the new version of Hydrogen.  Once this is working we can slowly migrate all of the objects to have GPU specific code.

PR #305, opened 2018-04-27
Visual Attention

timmoon10:
This PR adds support for visual attention, namely an approximation to DRAM (Ba 2015). Major changes:
- New layers: crop, log, discrete random, categorical random.
- GPU implementations of Hadamard and tanh layers.  LSTMs can now reside entirely on GPUs.
- Recurrent models now only support objective functions in the objective-functions-as-layers paradigm (#193).
- The CUDA and cuBLAS error-checking macros now synchronize GPUs before and after each call to thoroughly check for errors.


timmoon10:
This is ready to merge, but I would prioritize #250 first.

src/callbacks/callback_dump_activations.cpp
```diff
@@ -34,23 +32,26 @@ namespace lbann {
...
 void lbann_callback_dump_activations::on_forward_prop_end(model *m, Layer *l) {
 
-  //Skip if we are interested in saving inferences at (a) given layer(s)
-  if(!m_layer_names.empty()) return;
+  // Skip if we are interested in saving inferences at (a) given layer(s)
+  if (!m_layer_names.empty()) { return; }
 
-  // Skip target layers
-  if (dynamic_cast<generic_target_layer*>(l) != nullptr) {
-    return;
```
samadejacobs: Why eliminate this check?

include/lbann/layers/activations/tanh.hpp
```diff
@@ -64,7 +64,6 @@ class tanh_layer : public entrywise_activation_layer {
     // Activate GPU if needed
     if (cudnn != nullptr && T_layout == data_layout::DATA_PARALLEL) {
```
bvanessen: I believe that the guard for only using the GPU if it is in data parallel mode should go away.

include/lbann/layers/transform/crop.hpp
```diff
@@ -140,15 +142,15 @@ class crop_layer : public transform_layer {
   protected:
 
   void fp_compute() override {
-    if(this->m_using_gpus) {
+    if (this->m_cudnn != nullptr) {
```
bvanessen: I think that you should switch to a predicate like:

if(this->using_gpus()) {

to be more consistent with other part of the code.

include/lbann/layers/transform/crop.hpp
```diff
@@ -140,15 +142,15 @@ class crop_layer : public transform_layer {
...
       fp_compute_gpu();
     } else {
       fp_compute_cpu();
     }
   }
 
   void bp_compute() override {
-    if(this->m_using_gpus) {
+    if (this->m_cudnn != nullptr) {
```
bvanessen: if(this->using_gpus()) {

include/lbann/layers/transform/hadamard.hpp
```diff
@@ -96,159 +74,46 @@ class hadamard_layer : public transform_layer {
...
-    std::vector<const Mat*> local_inputs;
-    for (int i = 1; i < num_parents; ++i) {
-      local_inputs.push_back(&get_local_prev_activations(i));
-    }
-    auto& local_output = get_local_activations();
-
-    // Compute entrywise product
-    const int local_height = local_output.Height();
-    const int local_width = local_output.Width();
-    #pragma omp parallel for collapse(2)
```
bvanessen: Is it safe to assume that the Elemental Hadamard is also parallelized with OMP for the CPU?

samadejacobs:
Additional note here, log layer is currently placed within activation folder. It could as well a type of transform layer, entry wise layer, or loss layer. We may to have at some point make room for (and clearly distinguish) all these layer types as part of objective function as layer refactoring #154

src/callbacks/callback_dump_activations.cpp
timmoon10: We now iterate through all the activations rather than explicitly dumping the first activation. Target layers have no children, so it doesn't have any activations to dump.

PR #306, opened 2018-04-28
modifying cuda and cudnn error condition messages to be more informative

davidHysom:
I'm still getting "CUDA error" when running on surface, so decided to revise the error messages to be mor informative.

### DIFF
--- a/include/lbann/utils/cudnn_wrapper.hpp
+++ b/include/lbann/utils/cudnn_wrapper.hpp
@@ -63,17 +63,16 @@
 
 // Error utility macros
 #ifdef LBANN_HAS_CUDNN
-#define FORCE_CHECK_CUDA(cuda_call)                                     \
-  do {                                                                  \
-    const cudaError_t cuda_status = cuda_call;                          \
-    if (cuda_status != cudaSuccess) {                                   \
-      cudaDeviceReset();                                                \
-      std::stringstream err;                                            \
-      err << __FILE__ << " " << __LINE__ << " :: "                      \
-          << "CUDA erro; status: " << cuda_status << " err string: "    \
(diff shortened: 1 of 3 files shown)


include/lbann/utils/cudnn_wrapper.hpp
```diff
@@ -67,20 +67,24 @@
   do {                                                                  \
     const cudaError_t cuda_status = cuda_call;                          \
     if (cuda_status != cudaSuccess) {                                   \
-      std::cerr << "CUDA error: " << cudaGetErrorString(cuda_status) << "\n"; \
-      std::cerr << "Error at " << __FILE__ << ":" << __LINE__ << "\n";  \
       cudaDeviceReset();                                                \
-      throw lbann::lbann_exception("CUDA error");                       \
+      std::stringstream err;                                            \
+      err << __FILE__ << " " << __LINE__ << " :: "                      \
+          << "CUDA erro; status: " << cuda_status << " err string: "    \
```
timmoon10: Spelling and formatting OCD.

davidHysom: This lbann_exception is about to be thrown:/usr/workspace/wsb/hysom/DATA_STORE/lbann/src/utils/cudnn_wrapper.cpp 349 :: CUDA erro; status: 30 err string: unknown error

This is in the cudnn_manager ctor. I'd really like to get this working so I can get some additional results to Brian for the LDRD review(s).  For pilot1 (merge_features) there's only a very slight execution time advantage to using data_store. However, that's running without cuda. I suspect I could show some advantage if I could run with cuda/, since file-reading/computation ratio would be more favorable.

PR #310, opened 2018-04-30
Add Docker definition file, update singularity definition file

DylanMcKinney:
In this PR I am adding a Docker definition file, and instructions on its use. I am also updating the very dated singularity definition file. I updated the README to direct to the containers subdirectory (renamed from singularity) for container build information. I have tested the docker container locally on my Mac and the build works. The singularity build completes without error, and ran at one point. However LC's singularity still isn't functioning properly, so I have not tested this one as recently.


DylanMcKinney:
@forsyth2 if you could just check the READMEs for spelling/any confusing bits that would be great.

PR #311, opened 2018-04-30
Metric that interacts with evaluation layers.

timmoon10:
This PR implements a "layer metric" that simply reads off the value from an evaluation layer.

This is related to the objective-function-as-layers refactor (issue #193).

### DIFF
--- a/src/metrics/layer_metric.cpp
+++ b/src/metrics/layer_metric.cpp
@@ -66,7 +66,7 @@ EvalType layer_metric::evaluate(execution_mode mode,
 
   const auto& start = get_time();
   auto&& eval = dynamic_cast<evaluation_layer<data_layout::DATA_PARALLEL>*>(m_evaluation_layer);
-  EvalType total_value = eval->get_value(true);
+  EvalType total_value = eval->get_value(true) * mini_batch_size;
   if (m_unit == "%") { total_value *= 100; }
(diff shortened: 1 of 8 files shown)


ndryden:
This is a digression, but why are we using `EvalType` for timing? `get_time` returns a `double` and if the clock has a nanosecond period then we're actually losing precision (not that I'd necessarily trust our timing measurements down to the nanosecond).

src/metrics/layer_metric.cpp
```diff
@@ -0,0 +1,79 @@
...
+EvalType layer_metric::evaluate(execution_mode mode,
+                                int mini_batch_size) {
+
+  // Check if evaluation layer pointer has been setup
+  if (m_evaluation_layer == nullptr) {
+    LBANN_ERROR("attempted to evaluate metric without setting a target layer");
+  }
+
+  const auto& start = get_time();
+  auto&& eval = dynamic_cast<evaluation_layer<data_layout::DATA_PARALLEL>*>(m_evaluation_layer);
```
ndryden: Are we not supporting model-parallel distributions? What if if the cast fails?

timmoon10:
It feels nicer to me to use a consistent datatype for all of our reporting, but I don't think it makes much difference whether times are reported in `double` or `EvalType`.

PR #321, opened 2018-05-05
flexible JAG reader

JaeseungYeom:
Make the JAG data reader more flexible to accommodate new modeling modes such as Cycle GAN.
Instead of requiring the modeling mode to choose independent/dependent variable types, allow users to directly specify the types.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -326,7 +326,7 @@ if (LBANN_HAS_ALUMINUM)
 endif ()
 
 if (LBANN_HAS_CONDUIT)
-  target_link_libraries(lbann PUBLIC ${CONDUIT_LIBRARIES})
+  target_link_libraries(lbann PUBLIC CONDUIT::CONDUIT)
 endif ()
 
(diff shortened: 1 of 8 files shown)


samadejacobs:
Are the other/better naming than dependent/independent? Not sure, just asking. Is it require of user to specify dependent (Y) mode? what will break if a dependent mode is not specified? 

As a said note, I noticed that fetch_label is not implemented. I have an implementation of fetch_label to get cycle GAN to work, this may go away with data reader refactoring.

JaeseungYeom:
If a user does not specify dependent, it will be left as the default, which is simulation input.
Similarly, if a user does not specify independent, the default, 'image' is used.
Suggest me a pair of names to replace independent/dependent. The major change here is to allow users to choose which source of data to use for which purpose, that was previously done by `set_model_mode()`. The rests are to use `m_independent`/`m_depdendent` instead of `m_model_mode`.

samadejacobs:
Not sure of a better name! I think it is good to merge now.

PR #325, opened 2018-05-10
Cycle gan

samadejacobs:
Continuing support for (Cycle) GAN implementation. This PR among other things include:
1. Added sigmoid_binary_cross_entropy_with_logits layer
2. Added capability to replace model weights by name
3. Merge_feature data reader without label input file
4. Support model training at subepoch (multiple minibatches) level
5. Minor clean up etc


include/lbann/layers/activations/sigmoid_bce_with_logits.hpp
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  DataType activation_derivative(DataType z) const override {
+    DataType res;
+    if( z >=0 ) {
+      DataType log_term_derivative = std::exp(-(z)) / (DataType(1) + std::exp(-(z)));
+      if(m_ground_truth_label) res = DataType(0) - log_term_derivative;
+      else res = DataType(1) - log_term_derivative;
+    }else {
+      DataType log_term_derivative = std::exp(-(std::abs(z))) / (DataType(1) + std::exp(-(std::abs(z))));
```
timmoon10:
```
DataType log_term_derivative = - 1 / (std::exp(z) + 1)
```

include/lbann/layers/activations/abs.hpp
```diff
@@ -0,0 +1,56 @@
...
+  std::string get_type() const override { return "abs"; }
+  data_layout get_data_layout() const override { return T_layout; }
+
+ protected:
+  DataType activation(DataType x) const override {
+    return std::abs(x);
+  }
+
+  DataType activation_derivative(DataType x) const override {
+    return DataType(1);
```
timmoon10: The derivative should be -1 if x < 0.

include/lbann/layers/activations/l2_loss.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include "lbann/layers/activations/activation.hpp"
+
+namespace lbann {
+
+/** Half the L2 loss of a tensor without the sqrt.
+ * (x^2)/2
+ * @todo relocate to loss layer as part of github issue 154
+ */
+template <data_layout T_layout>
+class l2_loss_layer : public entrywise_activation_layer {
```
timmoon10: This is a special case of the `power_layer` added in #305. I think a dedicated L2 loss layer would be useful, but it should take both the prediction and the ground truth as inputs.

samadejacobs: The implementation mimics TF (https://www.tensorflow.org/api_docs/python/tf/nn/l2_loss), but yes, we can refactor to take both prediction and ground truth. We could also add scaling factor to power_layer for general purposes that would be useful for this.

PR #328, opened 2018-05-11
create tarball of files on local ssd, to enable faster startup time in subsequent runs

davidHysom:
This PR adds the capability to create tarballs of files on local ssd, and copy to e.g, lscratchX; the intent is to reduce startup time in subsequent runs. This will hopefully save much time for large data sets, e.g, imagenet22K. Currently each file is individually copied to local store.


src/data_store/generic_data_store.cpp
```diff
@@ -293,4 +311,57 @@ void generic_data_store::init_minibatch() {
...
+}
+
+void generic_data_store::create_dirs(std::string s) {
+  if (s.back() != '/') {
+    s += '/';
+  }
+  size_t i = s.find('/', 1);
+  while (i != std::string::npos) {
+    std::string s2 = s.substr(0, i);
+    const int dir_err = mkdir(s2.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
```
JaeseungYeom: The utility function make_dir() in src/io/file_io.cpp makes sure that only the master rank creates a directory and synchronizes. However, if there are large number of directories to create, this function is not efficient. 
In that case, you can use this function (or the similar existing function create_dir() in src/utils/file_utils.cpp) to create directories, but have only the master create directories and make sure to use barrier after that.

src/data_store/generic_data_store.cpp
```diff
@@ -293,4 +311,57 @@ void generic_data_store::init_minibatch() {
...
+      throw lbann_exception(err.str());
+    }
+    i = s.find('/', i+1);
+  }
+}
+
+std::string generic_data_store::run_cmd(std::string cmd, bool exit_on_error) {
+  std::array<char, 128> buffer;
+  std::string result;
+  std::shared_ptr<FILE> pipe(popen(cmd.c_str(), "r"), pclose);
```
JaeseungYeom: If you have already tested this on ray, then disregard this comment. I had some issue with passing string::c_str() to call a system function on ray. For some reason, I had to use the real c-string (null-terminated) in a character array.

src/data_store/data_store_image.cpp
```diff
@@ -636,4 +639,146 @@ void data_store_image::fetch_data() {
   }
 }
 
+void data_store_image::write_file_sizes() {
+if (m_master) std::cerr << "XXX data_store_image::write_file_sizes; count: " << m_file_sizes.size() << "  role: " << m_reader->get_role() << "\n";
```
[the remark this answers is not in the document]
davidHysom: thanks; thought I'd removed that.

src/data_store/data_store_image.cpp
```diff
@@ -636,4 +639,146 @@ void data_store_image::fetch_data() {
...
+    err << __FILE__ << " " << __LINE__ << " :: "
+        << "failed to open " << s.str() << " for reading";
+    throw lbann_exception(err.str());
+  }
+  size_t idx;
+  size_t len;
+  while (in >> idx >> len) {
+    m_file_sizes[idx] = len;
+  }
+if (m_master) std::cerr << "XXX data_store_image::read_file_sizes; count: " << m_file_sizes.size() << "  role: " << m_reader->get_role() << "\n";
```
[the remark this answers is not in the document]
davidHysom: thanks; thought I'd removed that.

src/data_store/generic_data_store.cpp
```diff
@@ -76,6 +76,18 @@ generic_data_store::generic_data_store(generic_data_reader *reader, model *m) :
   if (opts->has_bool("verbose") && opts->get_bool("verbose")) {
     m_verbose = true;
   }
+
+  if (opts->has_string("use_tarball")) {
+    m_dir = m_reader->get_local_file_dir();
+  }
+
+  if (m_comm->get_num_models() != 1) {
```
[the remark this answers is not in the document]
davidHysom: I'm not sure I understand your question, but I'll try to answer. I assume all models use the same data set (that may be a bad assumption). The problem is that there are many places where the data_store code may break if there is more than one model. Since all (or almost all) of our model_zoo models use a single model (procs_per_model: 0 in the prototext file), for an initial implementation I assumed there is a single model (hope that makes sense).

src/data_store/generic_data_store.cpp
davidHysom: Didn't know about that; I'll check it out. Thanks.

src/data_store/generic_data_store.cpp
davidHysom: I've only worked on catalyst. " I had some issue with passing string::c_str() to call a system function on ray ..." Thanks for the heads-up; and that seems weird.

src/data_store/generic_data_store.cpp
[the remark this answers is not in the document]
samadejacobs: We have multi-model cases (e.g., LTFB) where data store is of importance. If this is an initial implementation may be it is ok to make assumption of a single model.

src/data_store/generic_data_store.cpp
davidHysom: modified to copy cmd to c-style string

PR #335, opened 2018-05-16
Siamse patch list tool

JaeseungYeom:
This PR is to create an offline tool that enables selection of Siamese patch samples to run.
For example, if we want to selectively use samples from 30th to 65th out of total 100 samples, we need to create a list that only contains those samples. The sample list is packed in a numpy-based data structure compressed by using dictionaries. Each entry of a sample consists of kyes to multiple dictionaries from which values are extracted and concatenated to reconstuct the file path.

This tool helps creating a list of selected samples identified by a range from an existing list file.
This tools will probabaly only be used by myself, and barely touches the LBANN code base.
I have tested it for my use cases. Also find with not pushing this into the main branch depending on what people think.

### DIFF
--- a/include/lbann/data_readers/offline_patches_npz.hpp
+++ b/include/lbann/data_readers/offline_patches_npz.hpp
@@ -81,6 +81,11 @@ class offline_patches_npz {
   /// Return the label of idx-th sample
   label_t get_label(const size_t idx) const;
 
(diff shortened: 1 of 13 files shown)


davidHysom:
Seems OK to me. Yes, I think you should check it in -- and hopefully include some documentation so others know how to run it.

JaeseungYeom:
I will add a documentation after merging.

PR #356, opened 2018-05-19
GPU performance optimization

timmoon10:
This PR whittles away at performance bottlenecks when I run Resnet-50 with GPUs on Surface gpgpu:
- Stop allocating unnecessary memory for forward prop and backward prop inputs. This should hopefully halve our GPU memory requirements.
- Use the CUB memory pool when it is convenient to repeatedly allocate/deallocate GPU matrices, e.g. when a matrix often switches between owning data and being a matrix view.
- Pipeline cuBLAS calls when computing the L2 regularization term and divide work amongst GPUs (restoring functionality from #266). This implementation also only requires one allreduce instead of an allreduce for each set of weights.

Note that this requires the bugfixes in https://github.com/LLNL/Elemental/pull/6.

### DIFF
--- a/include/lbann/layers/regularizers/dropout.hpp
+++ b/include/lbann/layers/regularizers/dropout.hpp
@@ -51,11 +51,15 @@ class dropout : public regularizer_layer {
     : regularizer_layer(comm),
       m_keep_prob(keep_prob) {
 
-  #if defined(LBANN_HAS_CUDNN) && !defined(LBANN_SEQUENTIAL_CONSISTENCY)
     // Initialize GPU memory if using GPU
-    /// @todo GPU implementation of dropout with sequential consistency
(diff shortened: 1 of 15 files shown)


include/lbann/layers/regularizers/dropout.hpp
```diff
@@ -54,10 +54,7 @@ class dropout : public regularizer_layer {
   #if defined(LBANN_HAS_CUDNN) && !defined(LBANN_SEQUENTIAL_CONSISTENCY)
```
bvanessen: The guard for && !defined(LBANN_SEQUENTIAL_CONSISTENCY) should probably go away, or we need to prevent dropout from being marked as a not legal for GPUs when SEQUENTIAL consistency is requested.  Otherwise this causes a runtime error.

Issue #320, opened 2018-05-05
Hints for layer output dimensions


samadejacobs:
Something akin to [tf.like_zero(tensor)](https://www.tensorflow.org/api_docs/python/tf/ones_like) and [tf.like_one(tensor)](https://www.tensorflow.org/api_docs/python/tf/zeros_like), a good use case will be (Cycle)GAN. This will generalize previous modification to MNIST reader in #296 and also important for data_reader refactoring. One major concern here is that we may have to decouple target layer from input layer (again?)

timmoon10:
How about we generalize? We add functionality so that layers can take the dimensions of another layer. To emulate tf.like_zero, you would construct a constant_layer (with value 0) and set it to have the dimensions of the other layer. I'm thinking about this approach because it could also replace our current functionality where you can set the fully-connected layer output to match the data readers. It could also be useful in recurrent networks.

samadejacobs:
Yes you are right, constant_layer will serve that purpose for both like_zero and like_=one . I have actually been using constant_layer as described except that I manually specify "num_neurons".

timmoon10:
Closed with #703.

Issue #297, opened 2018-04-20
Weight initialization broken


samadejacobs:
Weight initialization within layer definition no longer work as user would expect due to recent changes in layer/weight refactoring. We should clean up the prototext files to reflect appropriate changes.

timmoon10:
I have some other proposals if we refactor the prototext definition:
- Redo the field numbering to be more systematic. It's a mess and adding a new message type involves hunting for a number that isn't already taken.
- Change to a Caffe-ish approach to specify layer type. In particular, instead of defining a new prototext type for each new layer, the type should be determined by a string, something like:
```
layer {
  name: "conv1"
  type: "Convolution"
  convolution_layer_params {
    etc: "etc"
  }
}
```

davidHysom:
While there's some merit to your proposal (second bullet) I believe this means there would be a single Layer (in lbann.proto); so that Layer would need to include a field for every parameter for every layer (in the lbann sense). That's not necessarily a bad thing, or code bloat -- but I think it complicates writing new model files. Because for every layer in your model you would have to remember (or look up somewhere) which of the many fields in Layer you need to specify. An advantage of the present schema is that you can look at lbann.proto, and immediately know which fields are pertinent to which layers.


BTW, I'm not partial to the present schema because I did the original design. I'm sure it can be improved. My design process was, "huh; there's lot of different ways to do this. Um ... OK, that one seems like it should work OK. Yeah. Let's go with that."


________________________________

timmoon10:
In the print callback, the `interval` option does nothing.

PR #357, opened 2018-05-21
Fc neurons

davidHysom:
added functionality: the fully connected layer preceeding a softmax layer no longer needs to specify num_neurons; instead, use the field: 'num_neurons_is_num_labels: true'

### DIFF
--- a/bamboo/README.md
+++ b/bamboo/README.md
@@ -67,7 +67,7 @@ To run a specific test file: `python -m pytest -s <test_file>.py`.
 
 To run a specific test: `python -m pytest -s <test_file>.py -k '<test_name>'`.
 
-Currently, it is not possible to run a test with a different executable. We plan on adding support for this.
+Most integration and unit tests allow for running a test with a different executable. The convention is to have a similarly structured test replacing `_<compiler_name>` with `_exe`. These tests are set to be skipped in Bamboo, but can be run locally. There should be a line above the test that gives the command to run the test locally.
 
 # Helpful Files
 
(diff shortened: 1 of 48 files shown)


timmoon10:
The `num_neurons_is_num_labels` option for the fully-connected layer duplicates the `num_neurons_from_data_reader` option for layers. I'm also rather uneasy that this approach only deals the fully-connected layer and doesn't handle tensor dimensions. A more general approach would be functionality so that the base layer class can take the dimensions of another layer (see #320). I don't think we should make widespread changes to our prototexts without some more discussion.

davidHysom:
Sam: changes to the data store should not have been in this PR.
Tim: I didn't know about "num_neurons_from_data_reader;" I wish someone had mentioned that when I first discussed my proposed changes several weeks ago.
I will remove this PR.

timmoon10:
I'm sorry for not noticing the conflict earlier. Our documentation also does need a lot of work, we also tripped up in #297... It may be worthwhile going through our prototext format, clearing out deprecated functionality, and designing a stable set of supported features.

davidHysom:
"The num_neurons_is_num_labels option for the fully-connected layer duplicates the num_neurons_from_data_reader option for layers"
After checking, this is different functionality. "num_neurons_from_data_reader" uses linearized_data_size.
"num_neurons_is_num_labels" set the number of neurons to be the number of labels.

JaeseungYeom:
Is this PR still active?

davidHysom:
There are several conflicts, both in the code and the above comments, so I am closing this PR.

PR #361, opened 2018-05-22
set num_neurons in FC to num_labels from data_reader

davidHysom:
This PR replaces #357, since that PR included changes to data_store classes that should not have been included. Please see Sam and Tim's comments on #357 before reviewing this PR.

This PR adds functionality for FC layers that precede softmax layers to set num_neurons to be the number of labels in the data reader. Previously, if the two numbers differ an exception was thrown, so making this auto-magic makes sense (to me). The new proto field in FullyConnected layer is: "num_neurons_is_num_labels: true" This field will over-ride the num_neurons field (if it exists).
Without this PR, we had to modify model_alexnet.prototext when switching between imagenet, imagement22K, etc.

I've only added "num_neurons_is_num_labels: true" to model_alexnet.prototext  
I propose we modify all applicable model_XX.prototext files, but Tim's comments (on PR 357) indicate more discussion is warranted prior to making those changes.

### DIFF
--- a/include/lbann/layers/learning/fully_connected.hpp
+++ b/include/lbann/layers/learning/fully_connected.hpp
@@ -157,6 +157,11 @@ class fully_connected_layer : public learning_layer {
 
   std::string get_type() const override { return "fully connected"; }
 
+  void set_num_neurons(int n) { 
+    m_num_neurons = n; 
+    this->m_neuron_dims.assign(1, this->m_num_neurons);
+  }
+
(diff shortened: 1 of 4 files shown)


timmoon10:
On second thought I think this PR's functionality would be useful. I was a little hesitant since it could be extended (#320), but that plan requires input layers emit both data and labels. That functionality is still in flux and it might be a while before we converge to a stable interface.

PR #362, opened 2018-05-22
Read the Docs support using Breathe+Sphinx

DylanMcKinney:
This PR updates our documentation to use RtD.  It includes configuration files necessary to convert our doxygen documentation to sphinx format, and sphinx formatted rst pages. 

Notable changes:

- Automatically rebuilds documentation with every commit, while avoiding the gh-pages scenario that bloated our repo. Generated documentation is stored on RtD servers, not anywhere in our repo.  
-  Pages are now (semi) automatically updated. Any file added into the subdirectories which have pages will be automatically added to that page in the documentation.  
-  Pages are pretty easy to add/modify. Rst is basically what you see is what you get, and any RtD project has a "show source" button on every page. So if you see some fancy formatting or something in a different project's documentation it's easy to see how its done. 
- RtD also has some really nice documentation versioning options I haven't messed with much.

**Note:** This PR does not flesh out the documentation; it's still fairly bare bones. With this structure in place and the relative simplicity of rst hopefully fleshing it out won't be too painful.

I have a RtD project tied to this PR, so to preview these changes you can look here: https://doxsphinx-test.readthedocs.io/en/latest/index.html
Currently it is just short descriptions of our main classes (callbacks, layers, metrics etc) followed by a dump of the documentation for each file in the subdir. Once this merges I should be able to just add our main develop to my RtD projects page and this should be all set to go.


DylanMcKinney:
Changes made, am I good to merge this?

PR #363, opened 2018-05-22
Deterministic cuDNN calls 

DylanMcKinney:
This PR adds the ability to switch to deterministic cuDNN calls if the LBANN_DETERMINISTIC_CUDNN flag is specified. By default this flag is also set when --debug is used with the lbann build script. When building with doubles **results** are reproducible, however dumping the weights shows that there is still tiny differences on a run to run basis.

This PR also updates the build script so that using the --datatype <float,double,etc> works again

**Edit:** As of 6/25 results are entirely reproducible regardless of datatype


davidHysom:
"however dumping the weights shows that there is still tiny differences on a run to run basis"
1) can you quantify "tiny?"
2) how many processors have you run with this? Do you still get differences with a single processor?
3) are weights save as ascii or binary? If ascii, I'm not surprised that there would be differences.
None of my questions preclude approving this PR -- I'm just curious.

DylanMcKinney:
I checked the dumped weights in binary before checking ascii. The binary weights came back as differing run to run. However since the summary stats were all coming back the same that didn't seem right so I checked with ASCII. To try to quantify, an example from one of the weight dumps:
**run one**: 0.02789953042392163
**run two**: 0.0278995304239216

So not exactly the same but not different enough to impact training I don't **think**. I haven't tried this with a single processor but I will. This was just on one pascal node

DylanMcKinney:
@timmoon10 I tagged you as a reviewer to check if the small changes I made in dropout.hpp and l2.cpp make sense. I noticed when building debug these two places failed and I think the changes I have are what we need.

include/lbann/layers/learning/base_convolution.hpp
```diff
@@ -428,8 +428,13 @@ class base_convolution_layer : public learning_layer {
...
     // Determine transposed convolution algorithm
     const size_t work_space_size = this->m_cudnn->get_work_space_size();
+    #ifndef LBANN_DETERMINISTIC_CUDNN
     cudnnConvolutionBwdDataAlgo_t transposed_convolution_cudnn_algorithm
       = CUDNN_CONVOLUTION_BWD_DATA_ALGO_0;
+    #else
+    cudnnConvolutionBwdDataAlgo_t transposed_convolution_cudnn_algorithm
+      = CUDNN_CONVOLUTION_BWD_DATA_ALGO_1;
+    #endif
     CHECK_CUDNN(cudnnGetConvolutionBackwardDataAlgorithm(this->m_cudnn->get_handle(),
```
timmoon10: I don't think the `LBANN_DETERMINISTIC_CUDNN` flag is having an effect right now since the algorithm choice is being overwritten by `cudnnGetConvolution<Forward/Backward>Algorithm()`. This function should only be called in the non-deterministic case.

DylanMcKinney: Oops, I didn't even notice that! When I tried moving the getConvAlgo call into the non_deterministic case I get a failure in the cudnnConvolutionBackwardData call that follows. The cuDNN error is CUDNN_STATUS_BAD_PARAM. I don't know why changing the algorithm would cause this. I read through what could cause it here: https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#cudnnConvolutionBackwardData 
but unfortunately I am still unsure. Any idea what else I need to change?

ndryden: We support 3D convolutions, but I don't believe any of our current models use it. Perhaps the right answer is to use a deterministic algorithm for 2D convolutions, and just print a warning or something at startup for 3D convolutions.

DylanMcKinney: Ok, good to know. I will make those additions

timmoon10:
Strange that the problem just went away. Otherwise, looks good to me!

DylanMcKinney:
Agreed! It leads me to maybe I was incorrectly resizing the workspace or something of that ilk

PR #364, opened 2018-05-23
Jag conduit multi-source data

JaeseungYeom:
This PR is to allow combining multiple data sources, such as images, scalars, simulation parameters, for independent/dependent variable, which is required by Cycle GAN
- m_independent and m_dependent are now the vector of variable_t type instead of just variable_t

Still need to work on:
- fetch_data()/fetch_response() will define view of El::Mat& X for writing each source into
- prototext support for specifying the list of variable type


JaeseungYeom:
will continue to work on the next PR. merging it now to avoid conflicts

Issue #351, opened 2018-05-16
GPU memory profiling


bvanessen:
Add callback that can calculate the amount of GPU memory being used during different phases of execution.  Also add a bamboo test to check our memory footprint on nightly tests (maybe weekly)

PR #372, opened 2018-05-29
Adds a callback to print GPU memory usage at the beginning of each epoch

naoyam:
This PR is to address #351. It simply prints out the size of the total allocated memory at the beginning of each epoch. For example, here's what it looks like with AlexNet on Pascal with 2 ranks:

```
[0] Epoch : stats formated [tr/v/te] iter/epoch = [451/51/2]
            global MB = [ 256/ 256/ 256] global last MB = [ 105  /  11  / 244  ]
             local MB = [ 256/ 256/ 256]  local last MB = [ 105+0/  11+0/ 244+0]
GPU memory usage at epoch 0 of model 0 at rank 1: 3463446528 bytes (3.23 GiB) used out of 17071734784 bytes (15.9 GiB)
GPU memory usage at epoch 0 of model 0 at rank 0: 3463446528 bytes (3.23 GiB) used out of 17071734784 bytes (15.9 GiB)
```
Let me know if the output looks OK.

### DIFF
--- a/include/lbann/callbacks/callback_gpu_memory_usage.hpp
+++ b/include/lbann/callbacks/callback_gpu_memory_usage.hpp
@@ -1,0 +1,55 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2016, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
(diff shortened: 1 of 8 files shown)


include/lbann/callbacks/callback_gpu_memory_usage.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#ifndef __LBANN_CALLBACKS_CALLBACK_GPU_MEMORY_USAGE_HPP_INCLUDED
+#define __LBANN_CALLBACKS_CALLBACK_GPU_MEMORY_USAGE_HPP_INCLUDED
+
+#include "lbann/callbacks/callback.hpp"
+
+#ifdef LBANN_HAS_CUDA
+
+namespace lbann {
+/** Callback hooks for gradient check. */
```
samadejacobs: Is this correct? I guess should be Callback hooks for GPU memory usage?

naoyam: Thanks. It was my mistake. I reverted the merge.

PR #374, opened 2018-05-30
 Add ability to create reader with role:validate in data reader prototext 

LukeJaffe:
This minor change will allow a user to create a separate reader with role:validate in the data reader prototext. Currently, the validation reader is created as a percentage of samples used for the train set; however, it is common for datasets to have explicit validation and test sets which are separate from the train set. This change will allow users to measure train, validation, and test performance in the typical fashion, and to create deterministic comparisons between different training scenarios (e.g., cross-framework benchmarking).

In src/proto/proto_common.cpp, a local flag (separate_validation) was added to test if the user has specified a separate validation set. A scan of all readers and their roles is done before the main reader loop to check for a reader with role:validate. This flag is tested later to avoid creating a validation set as a percentage of the train set. Otherwise, the process for creating a validation reader was simply copied from the train and test reader setup.

The change was tested both with and without a separate validation reader, and no issues were found.

Note: This is my first pull request to lbann, so please let me know if there is a mistake or anything could be done better.

### DIFF
--- a/src/proto/proto_common.cpp
+++ b/src/proto/proto_common.cpp
@@ -49,6 +49,18 @@ void init_data_readers(lbann::lbann_comm *comm, const lbann_data::LbannPB& p, st
   const lbann_data::DataReader & d_reader = p.data_reader();
   int size = d_reader.reader_size();
 
+  // A separate explicit validation set is created only if a reader with role "validate"
+  // is found in the list of data readers. Otherwise, a validation set is created as a
+  // percentage of data from the train set.
+  bool separate_validation = false;
+  for (int j=0; j<size; j++) {
+    const lbann_data::Reader& readme = d_reader.reader(j);
+    if (readme.role() == "validate") {
+        separate_validation = true;
+        break;
+    }
+  }
+
   for (int j=0; j<size; j++) {
     const lbann_data::Reader& readme = d_reader.reader(j);
     // This is a temporary measure until we individually setup data reader specific preprocessors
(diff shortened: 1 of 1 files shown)


JaeseungYeom:
You might want to update your branch and try it again.
I had the exact same error and it was resolved around 6 days ago.

LukeJaffe:
I see, I just merged with develop from today, rebuilt, and got the same error with 32 nodes. I'll investigate the source this week.

LukeJaffe:
I am still getting the categorical_accuracy.cpp 171 error, but for a 1 node use case now. Using lbann/develop commit 2e458386cbefe1b55c2556b77ca9c723e1cdf646, merged with this PR, with the attached prototexts, the error is occurring.

I have noticed that the error only triggers when the optimizer is SGD, but not for Adam, with this example.

Files to reproduce:
[validate_bug2.zip](https://github.com/LLNL/lbann/files/2100632/validate_bug2.zip)

JaeseungYeom:
Need the group 'brain' access permission set on the data files.

JaeseungYeom:
Anyway, the code looks fine to me. I do not believe the error has to do with this PR. I also understand that without resolving the issue this PR does not help your experiment as much as it can be.
In my opinion, you could probably just merge this PR as David already has approved, and track the issue on the issue thread posted.
The only thing I might add is the information on the amount of samples used is missing when separate validation set is used. It could probably be added after the main data reader parsing loop to print out the number of samples of each type reader if the reader is present when a separate validation set is used.

Issue #371, opened 2018-05-27
Potential synchronization error with Adam


ndryden:
Training ResNet-50 with Adam results in NaNs showing up. This seems similar to the prior error we had when Aluminum was not synchronized correctly, but the issue occurs without Aluminum.

I'm not seeing the error when using SGD, so the issue would seem to be related to Adam.

ndryden:
This issue occurs in a debug build as well.

ndryden:
This doesn't seem to be restricted to Adam-- I'm also seeing it with SGD now. I'm not seeing it with AlexNet, or on CPU, so I suspect the GPU implementation of something is messed up.

ndryden:
I'm not seeing the issue when I use the model-parallel FC and softmax layer in ResNet-50.

Also, #375 came from this, but doesn't seem to fix the issue.

ndryden:
Closed by #375.

PR #375, opened 2018-05-30
Fix softmax GPU cutoff kernels

ndryden:


ndryden:
Looks good to me, too. #371 seems fixed by this. Thanks Tim!

PR #378, opened 2018-05-30
Jag conduit data reader for cycle_gan

JaeseungYeom:
Continued improvements of jag_conduit reader for cycle_gan support.
- Implemented the create_datum_views() to support mixing multiple data sources for a variable.  A view of El::Mat is constructed using the linearized size of each data source and its order in the mix. Then, the data from each is copied to the corresponding view.
- Similar changes are applied to the original jag reader to allow interim testings with the current Sam's setup before deprecating it and moving on to the new conduit-based reader.
- Allow users to specify an array of data sources for the independent/dependent variable via the prototex interface.

### DIFF
--- a/src/data_readers/data_reader_jag_conduit.cpp
+++ b/src/data_readers/data_reader_jag_conduit.cpp
@@ -429,9 +429,14 @@ void data_reader_jag_conduit::check_input_keys() {
 
 #ifndef _JAG_OFFLINE_TOOL_MODE_
 void data_reader_jag_conduit::load() {
-  if(m_gan_labelling) m_num_labels=2;
-  std::cout << "JAG load GAN m_gan_labelling : label_value "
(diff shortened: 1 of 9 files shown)


src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -428,6 +429,10 @@ void data_reader_jag_conduit::check_input_keys() {
 
 #ifndef _JAG_OFFLINE_TOOL_MODE_
 void data_reader_jag_conduit::load() {
+  if(m_gan_labelling) m_num_labels=2;
+  std::cout << "JAG load GAN m_gan_labelling : label_value "
```
davidHysom: Do you want all processors to write to cout -- or is the printing just for use during development?

JaeseungYeom: Good point. I think I will add if (is_master()) {},

Issue #347, opened 2018-05-16
Error signals overwrite matrices rather than update them


bvanessen:
Note that this depends on moving to the layer-based objective functions

PR #383, opened 2018-06-01
Layers for visual attention.

timmoon10:
A `stop_gradient_layer` blocks error signals from passing through during backprop (compare with https://www.tensorflow.org/api_docs/python/tf/stop_gradient). I've also implemented entry-wise min and max layers.

@samadejacobs TensorFlow includes GANs as an example use-case for `stop_gradient`.

### DIFF
--- a/include/lbann/layers/transform/max.hpp
+++ b/include/lbann/layers/transform/max.hpp
@@ -1,0 +1,196 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2016, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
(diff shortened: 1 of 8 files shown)


include/lbann/layers/transform/stop_gradient.hpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+#endif
+  }
+
+  void fp_compute() override {
+    El::LockedView(get_activations(), get_prev_activations());
+  }
+
+  void bp_compute() override {
+    // El::Zero(get_error_signals());
```
samadejacobs: This assumes error_signals matrix/container
 is initialized to Zero? Remove commented out code if not needed.

include/lbann/layers/transform/max.hpp
```diff
@@ -0,0 +1,196 @@
...
+        LBANN_ERROR(err.str());
+      }
+    }
+  }
+
+  void fp_compute() override {
+    const int num_parents = get_num_parents();
+    auto& output = get_activations();
+    switch (num_parents) {
+    case 0: El::Zero(output); break;
```
samadejacobs: Can we enforce that min and max layers have at least one parent?

include/lbann/layers/transform/stop_gradient.hpp
timmoon10: This is for convenience once we get around to implementing #347.

include/lbann/layers/transform/max.hpp
timmoon10: That sounds reasonable. I'll add that restriction.

timmoon10:
Freezing layers completely halts the optimization process for a given set of weights. stop_gradient layers prevent certain error signals from affecting the optimization, but doesn't stop it entirely. If a given layer has only one output, it will compute zero gradients, which will affect the optimizer state. This is more useful when a layer has multiple outputs.

PR #386, opened 2018-06-01
report the value of the environment variable MV2_USE_CUDA

JaeseungYeom:
report the value of the environment variable MV2_USE_CUDA


ndryden:
A quick note: You should use `std::getenv` to use the `getenv` from `cstdlib`-- otherwise you're probably using the `stdlib.h` one.

JaeseungYeom:
Thanks for pointing that out.
I thought I was using std:: but apparently not.

PR #387, opened 2018-06-01
fix the name of the member variable, m_model_rank, of generic_data_reader

JaeseungYeom:
`m_model_rank`, as described in the inline comment, actually represents the rank in model.
So, I am changing to `m_rank_in_model`, which is consistent with the naming in lbann::comm.
In lbann::comm m_model_rank represents something different, which is the rank of the model.


ndryden:
:shipit:

Issue #360, opened 2018-05-22
Refactor cuDNN manager


timmoon10:
Now that we support GPU Hydrogen (#250), much of the GPU functionality in LBANN is deprecated:
- GPU matrices.
- GPU memory allocation and movement routines.
- CPU memory pinning.
- Multiple GPUs per process.

In addition, #250 started templating layers by device type. Our old approach was to enable/disable the GPU implementation based on whether the layer had a null pointer to a cuDNN manager. We should clean up the code and commit to the template approach to minimize the risk of bugs.

bvanessen:
I agree that we should use the Device type and get rid of the null pointer to cuDNN manager.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #394, opened 2018-06-04
cuDNN manager only manages cuDNN.

timmoon10:
The bulk of GPU functionality in the cuDNN manager can now be performed with Hydrogen and Aluminum. This cleanup is a step toward #360.


ndryden:
Overall I don't see any errors. Have you run any benchmarks to make sure there's no performance regressions?

timmoon10:
Quick profiling doesn't show significant regressions. I also find that the bulk of time in Hydrogen Gemms takes place in cudaMallocs and cudaFrees.

ndryden:
Okay, that's good.

Should be a separate thing, but we might want to take a look at the memory allocation in GEMMs then. I know it sets up several temporary matrices. Maybe we can reduce that with the CUB memory pool?

include/lbann/layers/regularizers/batch_normalization.hpp
```diff
@@ -305,7 +305,7 @@ class batch_normalization : public regularizer_layer {
                                   get_prev_activations().LDim(),
                                   m_mean->Buffer(),
                                   m_var->Buffer(),
-                                  this->m_cudnn->get_stream());
+                                  El::GPUManager::Stream());
```
naoyam: Isn't it a little abusing to use Hydrogen as a GPU manager?

timmoon10: An alternative approach is to implement a GPU manager that wraps around Hydrogen's GPU manager, similar to the old cuDNN manager. It seems easier to me to use Hydrogen's GPU manager though.

naoyam: I understand it's easier with the current design and there's nothing wrong with it, for now. So, I'm not advocating changing the interface.

However, since Hydrogen is a library for matrix data and computations, it doesn't seem to be a good idea to use it as a GPU resource manager too. In general, libraries should not own resources such as GPU device contexts and (externally visible) streams, but the main program should do. For example, cuDNN and CUBLAS need stream to be passed from their user code.

PR #397, opened 2018-06-06
Cycle gan

samadejacobs:
Support for evaluation on held-out test dataset and  other minor cosmetic clean up.

### DIFF
--- a/src/data_readers/data_reader_merge_features.cpp
+++ b/src/data_readers/data_reader_merge_features.cpp
@@ -45,6 +45,7 @@ data_reader_merge_features::data_reader_merge_features(
   m_data_size(other.m_data_size) {
   if(other.m_label_reader != nullptr) 
     m_label_reader = other.m_label_reader->copy();
+  else m_label_reader = nullptr;
 }
 
 data_reader_merge_features& data_reader_merge_features::operator=(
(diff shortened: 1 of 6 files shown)


src/data_readers/data_reader_merge_features.cpp
```diff
@@ -43,7 +43,8 @@ data_reader_merge_features::data_reader_merge_features(
   const data_reader_merge_features& other) :
   generic_compound_data_reader(other),
   m_data_size(other.m_data_size) {
-  m_label_reader = other.m_label_reader->copy();
+  if(other.m_label_reader != nullptr)
```
timmoon10: We should handle the case where `other.m_label_reader == nullptr`.

src/data_readers/data_reader_merge_features.cpp
```diff
@@ -53,7 +54,8 @@ data_reader_merge_features& data_reader_merge_features::operator=(
   if (m_label_reader) {
     delete m_label_reader;
   }
-  m_label_reader = other.m_label_reader->copy();
+  if(other.m_label_reader != nullptr)
```
timmoon10: We should handle the case where `other.m_label_reader == nullptr`.

src/data_readers/data_reader_merge_features.cpp
samadejacobs: Fixed, thanks!

PR #400, opened 2018-06-06
Update model setup test 

DylanMcKinney:
Fixes a couple of bugs:
-  Resnet models failing incorrectly due to lack of memory
-  Adds correct reader for the alexnet variations that were failing incorrectly (Siamese, triplet)

Also adds GAN models to this test. A number of models are still skipped because the shared account does not have access to the datasets. 

This PR will remove this test from quarantine. Before merging I am going to tag a few folks about models which are still failing to see how to fix that/if we should not include these models in this test.

### DIFF
--- a/bamboo/unit_tests/test_unit_check_proto_models.py
+++ b/bamboo/unit_tests/test_unit_check_proto_models.py
@@ -30,10 +30,10 @@ def skeleton_models(cluster, dir_name, executables, compiler_name):
                     data_filedir_default = '/p/lscratchf/brainusr/datasets/MNIST'
                     data_reader_name = 'mnist'
                 elif 'adversarial' in file_name:
-		    data_filedir_default = '/p/lscratchf/brainusr/datasets/MNIST'
+                    data_filedir_default = '/p/lscratchf/brainusr/datasets/MNIST'
                     data_reader_path = '%s/model_zoo/models/gan/mnist/adversarial_data.prototext' % (dir_name)
-		    data_reader_name = None
-		elif 'discriminator' in file_name:
+                    data_reader_name = None
+                elif 'discriminator' in file_name:
                     data_filedir_default = '/p/lscratchf/brainusr/datasets/MNIST'
                     data_reader_path = '%s/model_zoo/models/gan/mnist/discriminator_data.prototext' % (dir_name)
                     data_reader_name = None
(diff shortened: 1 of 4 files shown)


DylanMcKinney:
@davidHysom Do motif model prototexts require additional steps to run? Currently this test fails on the resnet50_motif model with `could not construct layer res2a`

davidHysom:
They don't require additional steps to run.

My best guess is that no one has touched the motif renet

since I wrote it, and lbann has undergone enough changes

that it no longer works. I can take a look at fixing it, though

I'm on vacation until Wednesday.

bamboo/unit_tests/test_unit_check_proto_models.py
```diff
@@ -24,15 +25,41 @@ def skeleton_models(cluster, dir_name, executables, compiler_name):
                 data_filename_train_default=None
                 data_filedir_test_default=None
                 data_filename_test_default=None
+                data_reader_path=None
                 if 'mnist' in file_name:
                     data_filedir_default = '/p/lscratchf/brainusr/datasets/MNIST'
                     data_reader_name = 'mnist'
+                elif 'adversarial' in file_name:
+		    data_filedir_default = '/p/lscratchf/brainusr/datasets/MNIST'
+                    data_reader_path = '%s/model_zoo/models/gan/mnist/adversarial_data.prototext' % (dir_name)
```
forsyth2: Indent is off on this line.

DylanMcKinney: strange I wonder why python wasn't complaining about that. good catch

DylanMcKinney:
To clarify, for motif prototext is the usage to first use expandmotifs.py to generate a runnable prototext file from the motif prototext?

davidHysom:
>To clarify, for motif prototext is the usage to first use expandmotifs.py to generate a runnable prototext file from the motif prototext?

Yes. However, expandmotifs.py is broken, due to changes in our build system. Longer answer: the script builds the python wrappers for protocal buffers -- and it expects the directory <lbann>/build/<host>.llnl.gov/bin/protoc to exist. Except it no longer does.

At one point I was going to port expandmotifs.py to c++, but never finished that.

I think we should discuss, perhaps at next Wed. mtg, if we want to maintain the motif functionality. If yes, it needs (perhaps considerable) repair.

Additionally, model_resnet50_motif.prototext contains "split" layers, but currently model_resnet50.prototext does not -- so even if expandmotifs.py worked it would not output a runnable prototext model.

DylanMcKinney:
Ok thanks, that is good to know. I think a discussion at the meeting about motifs going forward makes sense, and I will just add it to the skip list in this unit test in the mean time!

DylanMcKinney:
@samadejacobs I added discriminator and adversarial to this test but they are failing on ray and pascal with `layer "noise" is trying to use GPUs but has an invalid pointer to the cudnn object`. Have you experienced this error before?

Issue #391, opened 2018-06-04
batchnorm hangs or segfaults when using GPU


JaeseungYeom:
When running triplet Siamese model, it crashes at the first time it hits the bn layer, or hangs.
![lbann-triplet-bn](https://user-images.githubusercontent.com/12957608/40925733-388bb380-67cf-11e8-9762-5bdc5671b9f1.png)
I happened both on surface and pascal.
However, it runs fine on catalyst.
I also tried with Alextnet with BN on surface. it ran fine with 4 nodes (16 GPUs), but it crashed with 8 compute nodes (32 GPUs)

MV2_USE_CUDA was set to 1 on surface and pascal.
I compiled LBANN with the options --use-nccl and --with-aluminum.
The mpi libraries linked are as below:
/usr/global/tools/mpi/sideinstalls/chaos_5_x86_64_ib/mvapich2-2.3/install-gcc-4.9.3-cuda-9.1/lib/libmpicxx.so.12
/usr/global/tools/mpi/sideinstalls/chaos_5_x86_64_ib/mvapich2-2.3/install-gcc-4.9.3-cuda-9.1/lib/libmpi.so.12

ndryden:
I don't recall seeing this with ResNet-50 on Pascal this weekend, but I might have had a different configuration. Can you try without Aluminum and see if the problem persists? Another good check would be to try with a debug build.

JaeseungYeom:
I will try a couple of things and come back, such as not providing --with-aluminum to compiler options, or
using the ordinary MPI w/wo NCCL by commenting the lines as below in the build_lbann_lc.sh?
  if [ "${CLUSTER}" == "pascal" -o "${CLUSTER}" == "surface" ]; then
    MPI_HOME=/usr/global/tools/mpi/sideinstalls/${SYS_TYPE}/mvapich2-2.3/install-gcc-4.9.3-cuda-9.1
    export MV2_USE_CUDA=1
  fi

ndryden:
I would try these two configurations: `./build_lbann_lc.sh` (no special config), and `./build_lbann_lc.sh --debug`. Using `--use-nccl` without `--with-aluminum` will use the old NCCL code in LBANN, which I don't think is well-supported (and is being removed).

davidHysom:
I can replicate the error on catalyst, so will debug there.

________________________________

JaeseungYeom:
Compiled in debug mode without aluminum but with nccl, and ran alexnet (as is with LRN).
[...]

PR #405, opened 2018-06-08
Handling case where layers have no local data.

### LINKED ISSUE #390: cuDNN error when num_samples == 0
cuDNN throws an error in debug mode when a dimension passed to `cudnnSetTensorNdDescriptor` is 0. This can occur in `cudnn::set_tensor_cudnn_desc` (called from `Layer::fp_setup_data`) when a process has no local samples in data-parallel mode. For example, when the last mini-batch size is 1 and there is more than one process.

I'm not sure whether we can just skip setting up the tensor descriptor, or if that will cause other issues.

timmoon10:
This should fix #390 and #391. Some changes made along the way:
- The base layer class no longer maintains cuDNN tensor objects.
- Fixing errors in checknan callback.
- Experiment script has option for different LBANN builds (e.g. Release and Debug).


bvanessen:
I am concerned with this PR because it has created a lot of replicated code in every layer that uses cuda.  I am guessing that the motivation to have the layer class no longer maintain the cudnn manager data is to help with layers that don't have GPU implementations, but this does not seem like the best answer.  Perhaps we need to create a GPU layer in the hierarchy that maintains all of that common data.

naoyam:
Is it possible to separate bug fix commits from other commits and make them two separate PRs? The removal of cuDNN tensor descriptors from the layer class is not fixing the no-data problem, but as its change is scattered across the code, it will be much easier to understand the overall impact.

ndryden:
I agree. There also seems to be some other unrelated changes (e.g. refactoring the checknan callback) that could be split out. It's hard to review with this many changes (and, if a bug is introduced, harder to bisect the history).

timmoon10:
I think our problems in the no-local-data case (#390) are difficult to separate from the way we've handled cuDNN objects in the base layer class. The tensor descriptors are used in three ways:
- To describe data-parallel tensors (conv, pooling, softmax, LRN).
- To describe local matrices for entrywise operations (relu, tanh, dropout).
- They are not used.

Currently, we get away with this since all model-parallel layers fall into categories 2 and 3. But this is inelegant.

I agree it's unwieldy to have to duplicate code for these seven layers. It's also unwieldy to have to maintain and setup tensor descriptors that satisfy both cases, even in layers where it doesn't make much sense (e.g. when they have no inputs/error signals or multiple of them).

 These become more complicated when we need to add error checking and intricate hacks to avoid computation with no local data.

timmoon10:
It's unavoidable that this is going to be an ugly pull request. The solution to #390 is to add a bunch of error checking whenever GPU objects are used to avoid bad behavior when there is no local data. For cuDNN objects, there are two options:
- Maintain cuDNN objects in all layers that need them (conv, pool, softmax, LRN, relu, tanh, dropout).
- Maintain cuDNN objects in the base class in a manner that works for all cuDNN layers and doesn't break non-cuDNN layers.

Reasoning about option 2 will only get harder as we add more layers with a wider range of behaviors and that break our old assumptions. Right now, we already have the assumptions that cuDNN layers have exactly one parent and one child and that model-parallel layers are entry-wise operations. I agree option 1 is unwieldy since we have to duplicate code for our seven cuDNN layers (conv, pooling, softmax, LRN, relu, tanh, dropout). We could mitigate this by making similar layers inherit from virtual classes (`data_parallel_cudnn_layer`, `entrywise_cudnn_layer`), but I personally think duplicated code is easier to maintain than intricate inheritance patterns.

timmoon10:
To deal with the replicated code problem, I've implemented a `cudnn::layer_tensor_manager` class (with `cudnn::data_parallel_layer_tensor_manager` and `cudnn::entrywise_layer_tensor_manager` as derived classes).

JaeseungYeom:
I have a feeling that #390 and #391 are different issues. I misled you by providing a wrong reproducer. This PR fixes 390 but not sure about 391. I tried this PR with the latest main devel branch for Siamese Triplet, which I built for the Aluminum with the MPI backend. I have an error as below. This is not the same error I reported. After all the recent changes, I might be missing something.
segfault

JaeseungYeom:
I no longer have the issue. Thanks a lot.

PR #409, opened 2018-06-08
correct value type handling in jag_conduit data reader

JaeseungYeom:
Handle various numeric types that is not float64 in conduit data, and
avoid blindly treating non-numeric values as a numeric.


JaeseungYeom:
Since this does not affect the LBANN in general but only affects the jag reader, I am merging it unless there is objection.

PR #410, opened 2018-06-08
Updated the input layer to support reconstruction target layers

bvanessen:
Updated the input layer to have a data reader target mode that can be
set to classification, regression, or reconstruction mode.  Based on
the mode, the second output of the input layer will get the correct
data.  This allows the reconstruction layer to have two inputs, and
avoid having the original_layer pointer.  Note that the greedy
layerwise model is temporarily broken and needs to deal with adding a
split layer into the iterative model.

### DIFF
--- a/include/lbann/base.hpp
+++ b/include/lbann/base.hpp
@@ -122,6 +122,8 @@ enum class pool_mode {invalid, max, average, average_no_pad};
 /** returns a string representation of the pool_mode */
 std::string get_pool_mode_name(pool_mode m);
 
+enum class data_reader_target_mode {CLASSIFICATION, REGRESSION, RECONSTRUCTION};
+
 namespace lbann {
 
 // Forward-declaration.
--- a/include/lbann/io/data_buffers/generic_io_buffer.hpp
+++ b/include/lbann/io/data_buffers/generic_io_buffer.hpp
@@ -35,15 +35,22 @@ namespace lbann
 {
 class fetch_data_functor {
  public:
-  fetch_data_functor (bool is_input_layer, bool is_for_regression) :
-    _is_input_layer(is_input_layer), _is_for_regression(is_for_regression) {}
-  int operator() (CPUMat& samples, CPUMat& response, generic_data_reader* data_reader) const {
+  fetch_data_functor (data_reader_target_mode target_mode) :
+    _target_mode(target_mode) {}
+  int operator() (CPUMat& samples, CPUMat& responses, generic_data_reader* data_reader) const {
     int num_samples_fetched = data_reader->fetch_data(samples);
     int num_responses_fetched;
-    if (_is_for_regression) {
-      num_responses_fetched = data_reader->fetch_responses(response);
-    } else {
-      num_responses_fetched = data_reader->fetch_labels(response);
+    switch(_target_mode) {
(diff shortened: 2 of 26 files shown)


ndryden:
@timmoon10 Can you make that an issue so we can track it outside this PR?

samadejacobs:
We had previously discussed three possibilities for target_mode, classification (label), regression (response), and reconstruction (input data). The implementation assume classification/label as default. This assumptions fails and breaks cycle GAN code because the input layer/data does not have labels. In other words, cycle GAN is not covered in three possibilities of target_mode. How would we handle this scenario (no target or synthetic target data/label)?

bvanessen:
I will add a fourth mode for no input.


Brian C. Van Essen

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Jun 11, 2018, at 4:35 PM, Sam Ade Jacobs <notifications@github.com<mailto:notifications@github.com>> wrote:


We had previously discussed three possibilities for target_mode, classification (label), regression (response), and reconstruction (input data). The implementation assume classification/label as default. This assumptions fails and breaks cycle GAN code because the input layer/data does not have labels. In other words, cycle GAN is not covered in three possibilities of target_mode. How would we handle this scenario (no target or synthetic target data/label)?

samadejacobs:
Also, we need to make sure that the children of input layer are not restricted to data and labels/responses. I am not sure what ordering is expected of children of input layer, but now it seems (from a brief look at the code) that it expect child1 to be data, child2 label. We should have flexibility such that input layer could feed a (set of) learning layer(s) directly or through slice layer or a combination of both. To allow such flexibility, target layer (label, responses) may have to the last child (of input layer) where a target layer derived from input layer exists (it may not as in the case of "fourth mode").

bvanessen:
> 	• As a design question, would it be better if the data reader knew nothing about the training mode? An alternative scheme is to give the data reader a list of the data it should emit (classification emits data and labels, regression emits data and responses, reconstruction just emits data). This might be more cumbersome for practitioners with straightforward problems, but is more generalizable (e.g. data sequences for RNNs or for learning with multiple images simultaneously).

It turns out that the data reader does not actually know anything about the training mode, it is really a flag to the input layer for which data reader functions to call.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

bvanessen:
For the input layer output 1 is always the sample and output 2 is either the label, sample, or response.  As for multiple outputs I think that may be best regulated to a split layer, but we have mentioned before that we will discuss the right way to to do that.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

samadejacobs:
We can discuss in person about adding split layer. I don't think that burden should be shifted to the users. We already have split layer automatically added if needed, we can do the same here. In the meantime (if you already working on a fix), get_linearized_label_size() and get_linearized_response() in generic_input_layer.hpp should handle cases where there are no labels/responses. Right now it only handle regression, we should extend it to the "fourth mode".

bvanessen:
Okay, I will work on that.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #412, opened 2018-06-08
Summarize certain runtimes from every rank

ndryden:
This adds support to the summarizer to log scalars from every rank instead of just the model masters. This is now used to log mini-batch, forward, backward, update, and optimization time from every rank.

The old approach of only logging from the model masters is kept as well, for backwards compatibility with plotting scripts. The overhead of this is minimal since communication is only done at the end of an epoch and each rank just pushes a few values into a vector.


samadejacobs:
Is the old approach of logging from the model masters the default? If not, how do we choose the old approach?

ndryden:
A few things are now logged from every rank (mini-batch time, forward-prop time, etc.). Most things are as before.

Issue #401, opened 2018-06-07
Remove --use-nccl


ndryden:
I believe we now detect NCCL support from the Aluminum config, and no longer need the `--use-nccl` flag in `build_lbann_lc.sh` nor the associated checks in our `CMakeLists.txt`. @timmoon10 if you can confirm this, I'll remove it.

timmoon10:
We can't just remove the `--use-nccl` flag since we need to include NCCL headers to use Aluminum's NCCL backend. See https://github.com/ndryden/Aluminum/issues/27.

PR #414, opened 2018-06-11
Better integration of Aluminum into the CMake

benson31:
This eliminates the `FindALUMINUM.cmake` module. The `Aluminum` target is found by `AluminumConfig.cmake`, installed as part of `Aluminum`'s build procedure.

That module exposes the variables `AL_HAS_CUDA`, `AL_HAS_MPI_CUDA`, and `AL_HAS_NCCL`. Additionally, it uses an imported target, `AL::Al` that should be used for all linkage, e.g.,
```
target_link_libraries(mytarget PUBLIC AL::Al)
```
This will forward all appropriate includes and third-party dependencies for Aluminum into the LBANN target.


ndryden:
After this is merged, we can address #401.

PR #417, opened 2018-06-11
GPU model-parallel softmax

ndryden:
GPU-aware model-parallel softmax implementation. This could probably be optimized further, but right now just avoiding CPU<->GPU transfers by running this on-device is likely most useful.

Passes gradient checking for our softmax regression test and AlexNet.


src/layers/activations/softmax.cu
```diff
@@ -101,5 +196,70 @@ void bp_cutoff(int height, int width,
     cutoff);
 }
 
+void max_local_col_entry(int height, int width,
+                         const DataType * __restrict__ input,
+                         int input_ldim,
+                         DataType * __restrict__ workspace,
+                         cudaStream_t stream) {
+  if (width <= 0 || height <= 0) {
```
timmoon10: When `width > 0` and `height == 0`., `workspace` should be set with very small values.

src/layers/activations/softmax.cu
```diff
@@ -101,5 +196,70 @@ void bp_cutoff(int height, int width,
...
+}
+
+void exp_and_col_sum(int height, int width,
+                     const DataType * __restrict__ input,
+                     int input_ldim,
+                     DataType * __restrict__ output,
+                     int output_ldim,
+                     DataType * __restrict__ workspace,
+                     cudaStream_t stream) {
+  if (width <= 0 || height <= 0) {
```
timmoon10: `workspace` should be set to zero when `width > 0` and `height == 0`.

src/layers/activations/softmax.cu
ndryden: Done.

src/layers/activations/softmax.cu
ndryden: Done.

src/layers/activations/softmax.cpp
```diff
@@ -75,12 +78,88 @@ void softmax_layer<data_layout::MODEL_PARALLEL, El::Device::CPU>::bp_compute() {
...
+  const auto& local_input = get_local_prev_activations();
+  auto& local_output = get_local_activations();
+  auto& local_workspace = m_workspace->Matrix();
+
+  const El::Int local_height = local_input.Height();
+  const El::Int local_width = local_input.Width();
+
+  // Find the maximum entry in each local column.
+  if (local_width > 0 && local_height == 0) {
+    El::Fill(local_workspace, std::numeric_limits<DataType>::min());
```
timmoon10: `local_workspace` should be filled with `lowest()` to deal with the case where all inputs are negative.

src/layers/activations/softmax.cpp
```diff
@@ -75,12 +78,88 @@ void softmax_layer<data_layout::MODEL_PARALLEL, El::Device::CPU>::bp_compute() {
...
+      local_input.LDim(), local_workspace.Buffer(), El::GPUManager::Stream());
+  }
+  // Find the global max entry in each column.
+  m_comm->allreduce(*m_workspace, m_workspace->RedundantComm(), El::mpi::MAX,
+                    std::type_index(typeid(Al::nccl_backend)));
+
+  // Exponentiate activations and compute column sums.
+  // This subtracts by the column max for stability.
+  if (local_width > 0 && local_height == 0) {
+    El::Fill(local_workspace, std::numeric_limits<DataType>::min());
```
timmoon10: `local_workspace` should be filled with 0.

src/layers/activations/softmax.cpp
```diff
@@ -75,12 +78,88 @@ void softmax_layer<data_layout::MODEL_PARALLEL, El::Device::CPU>::bp_compute() {
...
+  }
+  // Compute the global sums for each column.
+  m_comm->allreduce(*m_workspace, m_workspace->RedundantComm(), El::mpi::SUM,
+                    std::type_index(typeid(Al::nccl_backend)));
+
+  // Divide activations by the column sums.
+  // This rounds small values to avoid denormalization.
+  softmax_cuda::div_by_col_sums_and_cutoff(
+    local_height, local_width, local_output.Buffer(),
+    local_output.LDim(), local_workspace.Buffer(), m_min_output,
```
timmoon10: Should use `local_workspace.LockedBuffer()` for safety.

src/layers/activations/softmax.cu
```diff
@@ -58,15 +58,110 @@ __global__ void bp_cutoff_kernel(int height, int width,
...
+  const lbann::DataType * __restrict__ input,
+  int input_ldim,
+  lbann::DataType * __restrict__ workspace) {
+  const int tid = threadIdx.x + blockIdx.x*blockDim.x;
+  const int num_threads = blockDim.x * gridDim.x;
+  for (int col = tid; col < width; col += num_threads) {
+    const int col_offset = col*input_ldim;
+    lbann::DataType max_entry = input[col_offset];
+    for (int row = 1; row < height; ++row) {
+      max_entry = max(max_entry, input[row + col_offset]);
```
timmoon10: Try using `fmax()`.

src/layers/activations/softmax.cu
```diff
@@ -58,15 +58,110 @@ __global__ void bp_cutoff_kernel(int height, int width,
...
+  const lbann::DataType * __restrict__ workspace,
+  const lbann::DataType cutoff) {
+  const int tid = threadIdx.x + blockIdx.x*blockDim.x;
+  const int num_threads = blockDim.x * gridDim.x;
+  for (int col = tid; col < width; col += num_threads) {
+    const int col_offset = col*output_ldim;
+    const lbann::DataType scale = lbann::DataType(1) / workspace[col];
+    for (int row = 0; row < height; ++row) {
+#ifdef LBANN_ENABLE_SOFTMAX_CUTOFF
+      output[row + col_offset] = max(scale*output[row + col_offset], cutoff);
```
timmoon10: Try using `fmax()`.

src/layers/activations/softmax.cpp
```diff
@@ -75,12 +78,88 @@ void softmax_layer<data_layout::MODEL_PARALLEL, El::Device::CPU>::bp_compute() {
...
+  // Local matrices.
+  const auto& local_input = get_local_prev_activations();
+  auto& local_output = get_local_activations();
+  auto& local_workspace = m_workspace->Matrix();
+
+  const El::Int local_height = local_input.Height();
+  const El::Int local_width = local_input.Width();
+
+  // Find the maximum entry in each local column.
+  if (local_width > 0 && local_height == 0) {
```
timmoon10: We can simplify these if statements to just `local_height == 0` since the functions in `softmax_cuda` return immediately when `width == 0`.

Issue #213, opened 2018-02-10
OpenCV JPEG parser is slow with Lustre


timmoon10:
I've found an interesting performance regression while profiling AlexNet (ImageNet-10, 256 mini-batch size; 1 node on Surface gpgpu). It takes ~35 sec/epoch when we get data from Lustre and ~25 sec/epoch when data is cached in `/tmp`. I expect the first epoch to be slower with Lustre data, but previous experiments showed similar timings for subsequent epochs since the files would reside in cache.

The culprit is [`cv::imread()`](https://github.com/opencv/opencv/blob/601e3aaf988b8450634b931c10900633622bad04/modules/imgcodecs/src/loadsave.cpp#L625), which takes 25.5% of the runtime with Lustre data and 1.5% with cached data. The worst increase shows up in [`cv::ExifReader::getExif()`](https://github.com/opencv/opencv/blob/601e3aaf988b8450634b931c10900633622bad04/modules/imgcodecs/src/exif.cpp#L119), which extracts an Exif tag from the JPEG file header and takes 15.3% of the runtime. This is necessary since the Exif header determines the picture orientation, but it is wonky that parsing the header takes more time than loading the entire image. The parse function repeatedly calls `std::istream::read` for two bytes at a time, so I have some hypotheses:
- LC systems are no longer caching as well as before. Every time we read two bytes, we have to get it remotely.
- This is caused by the Meltdown patch. Reading two bytes from a file stored in cache involves system calls that are slowed down by the patch.

We have some options:
- Data stores (#197) should bypass this issue since their behavior should be similar to caching data in `/tmp`. However, from what I understand, the initial implementation requires that we be able to store the entire data set on the compute nodes.
- Disabling Exif parsing brings back most of the performance, although not all. [`cv::findDecoder()`](https://github.com/opencv/opencv/blob/601e3aaf988b8450634b931c10900633622bad04/modules/imgcodecs/src/loadsave.cpp#L185), [`cv::JpegDecoder::readHeader()`](https://github.com/opencv/opencv/blob/601e3aaf988b8450634b931c10900633622bad04/modules/imgcodecs/src/grfmt_jpeg.cpp#L207), and [`cv::JpegDecoder::readData()`](https://github.com/opencv/opencv/blob/601e3aaf988b8450634b931c10900633622bad04/modules/imgcodecs/src/grfmt_jpeg.cpp#L389) together take up 10% of runtime. This is at best a temporary solution since it would involve throwing away the picture's orientation data.
[...]

PR #421, opened 2018-06-12
use cv::imdecode() instead of cv::imread()

JaeseungYeom:
Add an interface, load_decode(), using cv::imdecode() to load image data instead of using cv::imread().
It basically reads the entire file from
This is to avoid reading the image header in determining the decoder directly from the file but to allow doing so from the memory.
This would help mitigating the issue #213 in case of sub-optimal filesystem cache operations.

### DIFF
--- a/include/lbann/data_readers/cv_utils.hpp
+++ b/include/lbann/data_readers/cv_utils.hpp
@@ -111,12 +111,7 @@ class cv_utils {
    *  return type. Avoiding the extra access to the underlying filesystem may
    *  result in a better performance.
    */
-  static cv::Mat load_decode(const std::string& img_file_path, int flags);
-  /**
-   * Additionally, this measures the time to load the image data from a file to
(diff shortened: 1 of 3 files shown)


timmoon10:
Avoiding the image header stuff in `imread()` would be very nice. Are any of our images going to have the wrong orientation without the Exif data? Do you know if there are any other differences?

Also, is there a reason why `LBANN_IMREAD` is a macro? I think it would be better practice if we called a function which could internally choose between `imread()` and `imdecode()`.

JaeseungYeom:
It does not bypass any image operations.
It is just a way to use a different version of overloaded findDecoder(). With imread(), findDecoder() opens a file to read the header. On the other hand, with imdecode(), it reads a memory buffer.
If no one find any issue, we can get rid of the macro. I have it as a quick way to switch back in case that something unforeseen comes up although I do not expect any.

PR #422, opened 2018-06-13
updates to build_lbann_lc.sh to support Aluminum build

benson31:
This adds Aluminum to the superbuild invocation and eliminates the `--with-nccl` flag. The aluminum build is controlled by flags `--with-aluminum` to enable the build, `--aluminum-with-mpi-cuda` to build the MPI-CUDA backend for Aluminum, and `--aluminum-with-nccl` to build the NCCL backend for Aluminum.

I encountered a problem where git needed my github credentials to pull Aluminum, which was not great in the context of the fast-moving superbuild, so I manually set `Aluminum_URL` to use the SSH protocol, and that was fine (I have a key setup). This will go away when Aluminum is officially released.

### DIFF
--- a/scripts/build_lbann_lc.sh
+++ b/scripts/build_lbann_lc.sh
@@ -67,6 +67,8 @@ WITH_CUDA=
 WITH_TOPO_AWARE=ON
 INSTRUMENT=
 WITH_ALUMINUM=OFF
+ALUMINUM_WITH_MPI_CUDA=OFF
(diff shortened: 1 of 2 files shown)


JaeseungYeom:
You have --aluminum-with-mpi-cda instead of --aluminum-with-mpi-cuda.

benson31:
Good catch. Fixed in #424.

PR #423, opened 2018-06-13
Update checkpoint tests 

DylanMcKinney:
`test_unit_checkpoint` was only really broken (and subsequently skipped in bamboo) because a hydrogen fix had not yet merged. This is now resolved, and a minor path issue for distributed ckpt test is fixed by this PR. 

`lbann2_reload` was broken in a couple ways: 

- dump weights callback failure due to the specified directory not existing before attempting the weight write. 

- it appears that by default lbann2 does not go through model training when weights are loaded from saved ckpt files. This causes the test  to fail, as the initial training weights are never dumped by the callback, and we check these weights against what would be loaded in during a standard lbann2 execution. I have resolved this by having the 'skip training' option as a command line option rather than default behavior.

### DIFF
--- a/model_zoo/lbann2.cpp
+++ b/model_zoo/lbann2.cpp
@@ -77,7 +77,7 @@ int main(int argc, char *argv[]) {
 
     // When using checkpoint states, skip training as those could be the result
     // of checkpointing by steps.
-    if (!opts->has_string("ckpt_dir")){
+    if (!opts->has_string("no_model1_train")){
       model_1->train( pb_model.num_epochs() );
     }
     // Evaluate model 1 unless it is set to skip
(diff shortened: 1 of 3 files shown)


model_zoo/lbann2.cpp
```diff
@@ -77,7 +77,7 @@ int main(int argc, char *argv[]) {
 
     // When using checkpoint states, skip training as those could be the result
     // of checkpointing by steps.
-    if (!opts->has_string("ckpt_dir")){
+    if (!opts->has_string("no_model1_train")){
```
DylanMcKinney: @JaeseungYeom I believe you had set this in a previous commit to, by default, not train model1 when loading from ckpt. Would it work for you to have this as a command line option `--no_model1_train` instead of the default behavior?

JaeseungYeom: That would be a better option.

model_zoo/lbann2.cpp
```diff
@@ -298,13 +298,16 @@ bool load_model_weights(std::string ckpt_dir, model * m){
   sprintf(latest, "%s/last.shared.checkpoint", ckpt_dir.c_str());
```
JaeseungYeom: Instead of finding out the last ckpt state under ckpt_dir, would it be beneficial to use ckpt_dir directly. This would give users an option to choose which state to begin with?

DylanMcKinney: I have it find the latest checkpoint only because that's the default usage of checkpoint/restart. I see how giving users the option could be beneficial. Do you mean rather than `--ckpt_dir=my_ckpt/` and then deciding which epoch/step to load from instead have a user give `--ckpt_dir/shared.model.0.epoch.#.step.#` and actually specify which checkpoint they would like to load from?

JaeseungYeom: yes

DylanMcKinney: I think that change makes sense, I will add it.

PR #431, opened 2018-06-14
Removing deprecated KL divergence objective function term.

timmoon10:
This has been deprecated as of #276, which implements the VAE KL divergence term as a series of simple layers outputting to an evaluation layer. See #193 for a litany of complaints regarding the old KL divergence term.


timmoon10:
I have some old data when #276 was implemented. On the JAG VAE model, we get similar objective function values with the VAE KL divergence term and using evaluation layers:
jag_vae
The MNIST VAE model in the model zoo (model_zoo/models/autoencoder_mnist/vae_mnist.prototext) also gets reasonable looking results.

Epoch | Input | Output
-------|-------|-------
1 |  images_input_epoch1 | images_output_epoch1
10 | images_input_epoch10 | images_output_epoch10
20 | images_input_epoch20 | images_output_epoch20

PR #434, opened 2018-06-14
initial implementation of jag_io class 

davidHysom:
Initial commit for jag_io class. This class provides functionality  to convert the *.bundle (aka, hdf5) data into our (my) proprietary format. It also provides functionality to read in data from our format, and pass the data to data_store_jag_conduit (note: data_store_jag_conduit doesn't yet exist). 

You should think of this class as : data_on_disk -> jag_io -> data_store_fag_io -> data_reader_jag_conduit.

### DIFF
--- a/include/lbann/data_store/jag_io.hpp
+++ b/include/lbann/data_store/jag_io.hpp
@@ -1,0 +1,193 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2016, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+// may not use this file except in compliance with the License.  You may
+// obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the license.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef __JAG_IO_HPP__
+#define __JAG_IO_HPP__
+
+#include "lbann_config.hpp"
+
+#ifdef LBANN_HAS_CONDUIT
+#include "conduit/conduit.hpp"
+#include "conduit/conduit_relay.hpp"
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+
+namespace lbann {
+
+class jag_io {
+ public:
+
+ /**
+  * NOTE: some methods below take variables "string node_name" while others
+  *       take "string key." My convention is that "node_name" indicated
+  *       a fully qualified name, i.e, it begins with the sample id:
+  *          0/field_1/field_2
+  *       on the other hand, "key" does not contain the sample id:
+  *          field_1/field_2
+  */
+
+/// WARNING! CAUTION! BE ADVISED!
+/// cut-n-paste from data_reader_jag_conduit; this is
+/// fragile -- probably best to but these in a small
+/// file that both reader and store can then include
+/// @todo
+using ch_t = double; ///< jag output image channel type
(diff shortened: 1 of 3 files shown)


JaeseungYeom:
Does it mean that 'data_reader_jag_conduit' now never see the raw conduit data, and offload all the conduit format handling to jag_io?
Or does the jag_io rely on 'data_reader_jag_conduit' to extract information to pack?

include/lbann/data_store/jag_io.hpp
```diff
@@ -0,0 +1,193 @@
...
+  jag_io();
+
+  //! copy ctor
+  jag_io(const jag_io&) = default;
+
+  //! operator=
+  jag_io& operator=(const jag_io&) = default;
+
+  //@todo: this causes a compile error ... why?
+  //jag_io * copy() const { return new jag_io(*this); }
```
JaeseungYeom: I do not have any error with clang/4.0.0 when I uncomment this line.

davidHysom: I got the error with gnu; from the error message it appears to be related to the conduit headers. I just tried compiling with clang on catalyst, but the build failed early on. I've sent Tom the error messages. I've started work on data_store_jag_conduit, and ran into the same compile error with copy().

JaeseungYeom: By any chance, are you using the conduit not built for the target platform?

davidHysom:
I'm using the one you sent me for catalyst. But if you compiled that with clang,

and I'm using gnu, that could possibly be a problem. Below is what I think is the

relevant portion of the error message.


/usr/tce/packages/gcc/gcc-7.1.0/include/c++/7.1.0/bits/stl_construct.h:75:7: error: use of deleted function 'std::basic_ifstream<_CharT, _Traits>::basic_ifstream(const std::basic_ifstream<_CharT, _Traits>&) [with _CharT = char; _Traits = std::char_traits<char>]'
     { ::new(static_cast<void*>(__p)) _T1(std::forward<_Args>(__args)...); }
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/workspace/wsb/hysom/DATA_STORE/lbann/build/gnu.Release.catalyst.llnl.gov/install/include/El/core.hpp:23:0,
                 from /usr/workspace/wsb/hysom/DATA_STORE/lbann/build/gnu.Release.catalyst.llnl.gov/install/include/El.hpp:14,
                 from /usr/workspace/wsb/hysom/DATA_STORE/lbann/include/lbann/base.hpp:32,
                 from /usr/workspace/wsb/hysom/DATA_STORE/lbann/include/lbann/data_store/generic_data_store.hpp:32,
                 from /usr/workspace/wsb/hysom/DATA_STORE/lbann/include/lbann/data_store/data_store_jag_conduit.hpp:31,
                 from /usr/workspace/wsb/hysom/DATA_STORE/lbann/src/data_store/data_store_jag_conduit.cpp:28:
/usr/tce/packages/gcc/gcc-7.1.0/include/c++/7.1.0/fstream:519:7: note: declared here
       basic_ifstream(const basic_ifstream&) = delete;



JaeseungYeom:
I copied the builds that I used to /p/lscratchh/brainusr/conduit
The one for catalyst is compiled with gcc-7.1.

JaeseungYeom:
Actually, I reproduced the error. The reason I was not seeing the error was that src/data_store/CMakeLists.txt does not include jag_io.cpp in the branch I checked out. So, I was never compiling your code.

src/data_store/jag_io.cpp
```diff
@@ -0,0 +1,386 @@
...
+  if (parent.find('/') != std::string::npos) {
+    // hack to discard keys that vary between samples;
+    // will fix later, when we have the samples we're going
+    // to actually use, and guidance as to which outputs
+    // we should use.
+    if (parent.find("outputs/scalars") == std::string::npos) {
+      m_keys.push_back( parent.substr(2));
+    }  
+  } 
+  conduit::Node nd2 = nd[parent.c_str()];
```
JaeseungYeom: c-string will automatically converted to std::string. So, there is no problem.
However, the interfaces expects "const std::string&" as below.
  Node             &operator[](const std::string &path);
  const Node       &operator[](const std::string &path) const;

JaeseungYeom:
One of the reason copy() does not work is because the default copy constructor is not working.
The reason is that the vector<ifsteram> member is copiable.
So, both the copy constructor and = operator needs to avoid copying m_data_streams.

src/data_store/jag_io.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+void jag_io::load(std::string base_dir) {
+
+  std::stringstream err;
+  std::string fn;
+  std::ifstream in;
+  std::string key;
+
+  // open the binary data file
+  m_data_streams.resize(omp_get_max_threads());
```
JaeseungYeom: This cannot be done.

JaeseungYeom:
Yes, that is because they store it as index_t type. They use this type pretty much for any integer based type.


Sent from my Verizon Wireless 4G LTE smartphone

-------- Original message --------

PR #438, opened 2018-06-15
ImageNet crops to 224x224

### LINKED ISSUE #428: ImageNet data reader crop size is wrong
The ImageNet data reader crops images to 227x227 after resizing them. It should crop them to 224x224 (blame says this is my fault...).

This is particularly important for ResNet, which expects the input to its average pooling layer to be 7x7 in our formulation.

I'm making this issue partly to bring attention to this, since the change may result in different accuracies in our tests (cc @DylanMcKinney @forsyth2).

ndryden:
Fixes #428.

### DIFF
--- a/model_zoo/data_readers/data_reader_imagenet.prototext
+++ b/model_zoo/data_readers/data_reader_imagenet.prototext
@@ -19,8 +19,8 @@ data_reader {
 
       cropper {
         disable: false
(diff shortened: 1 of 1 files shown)


ndryden:
More generally, different models require different data augmentation pipelines. It might be worthwhile to move all of that out.

Issue #407, opened 2018-06-08
resizing image without preserving aspect ratio


JaeseungYeom:
Another request from Luke to preserve the context from the original image even if distorting the aspect ratio.

JaeseungYeom:
Done at https://github.com/JaeseungYeom/lbann/tree/simple_resize

PR #442, opened 2018-06-18
Simple image resize

JaeseungYeom:
This PR implements the feature referred in issue #407.
It allows simple image resizing while preserving the context from the original image even if distorting the aspect ratio. My initial plan was to keep it in my private branch, but I decided to create a PR as merging will make Luke's work easier. The decision is up to the reviewer.

### DIFF
--- a/include/lbann/data_readers/cv_process.hpp
+++ b/include/lbann/data_readers/cv_process.hpp
@@ -37,6 +37,7 @@
 #include "cv_colorizer.hpp"
 #include "cv_decolorizer.hpp"
 #include "cv_cropper.hpp"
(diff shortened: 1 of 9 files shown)


ndryden:
What is adaptive interpolation?

JaeseungYeom:
It uses INTER_AREA when shrinking, and INTER_LINEAR when enlarging.
INTER_LINEAR is used as the default if not adaptive. Basically, it is to adopt the rule of thumb suggestion at:
https://docs.opencv.org/3.1.0/da/d54/group__imgproc__transform.html#ga47a974309e9102f5f08231edc7e7529d

PR #443, opened 2018-06-18
removing irrelevant "validation_percent" in prototext of the testset data reader

JaeseungYeom:
- print out the number of samples being used when a separate validation data reader is used
- removing "validation_percent: 1.0" in the test set data reader prototext because it does not do anything but creates confusion.  This is only used with training data readers when there is no separate validation reader is specified.


JaeseungYeom:
I will merge this soon since no one has objected, and it is easy to revert.

PR #445, opened 2018-06-18
updates per Sam and Tim's comments and requests.

davidHysom:
This PR attempts to update viz.py to deal with evolving changes in our prototext
format (lbann.proto). Nikoli pointed out that we perhaps should use the python bindings 
for viz, instead of my rather hacky hand-parsing. The problem with
Nikoli's suggestion is that viz.py would then need to be updated whenever
layers and their attributes in lbann.proto were added or modified. If lbann
ever becomes relatively stable, then yes, we should do so.

Specific changes:

Add support to display all attributes of layers. Note: due to evolving
changes in lbann.proto, some attributes failed to be printed when the
cmd option full=1 was given

viz.py would crash unless there was a blank line at the end of properties.txt;
this has been fixed

viz.py would crash if there was a space missing before '{', eg, "layer{'
or "elu{". This has been fixed (... but its sort of a hack).

If a layer type did not appear in properties.txt viz.py would crash.
Now the layer is printed as a grey rectangle. (Hence, if you see a grey rectangle,
take it as a reminder that you should update properties.txt)

### DIFF
--- a/viz/layer.py
+++ b/viz/layer.py
@@ -15,6 +15,9 @@ class Layer :
       if line.find('{') != -1 :
         t = line.split()
         self._type = t[0]
+        j = self._type.find('{')
+        if j != -1 :
+          self._type = self._type[:j]
     for line in self._layer :
       if line.find('parents:') != -1 :
         t = line.replace('"', '')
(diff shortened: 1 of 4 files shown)


timmoon10:
You have to do something like `./viz.py something.prototext full=1`.

davidHysom:
Also, run viz.py with no params to see full usage message

PR #446, opened 2018-06-18
Trying to fix performance regression.

timmoon10:
@ndryden has traced a performance regression in Resnet50 to #405. @samadejacobs has also observed a performance regression on Catalyst.

Changes:
- Initialize cuDNN tensors as 4D tensors whenever possible. My profiling shows that cuDNN ReLU gets poor performance on 1D and 2D tensors, often taking longer than convolution. `cudnn::entrywise_layer_tensor_manager` constructs a 4D tensor by factoring the local matrix height into three factors.


timmoon10:
@ndryden reports that this PR brings us back to expected performance.

PR #461, opened 2018-06-21
add key filtering mechanisms to jag_conduit data reader

JaeseungYeom:
Add two mechanisms to filter out certain keys with the default mode, which selects all the variables. One is by the exact string matching, and the other is by prefix substring matching. Users can express such a negative selection in the prototext of the jag_conduit data reader. Especially, "jag_scalar_filters" and "jag_input_filters" rely on the former. "jag_scalar_prefix_filters" and "jag_input_prefix_filters" rely on the latter.

For example, the line below will screen out the jag scalar variables with either of the "key1" or "key2".
jag_scalar_filters: ["key1", "key2"]

Given the line below, any variable of the key that has a length no shorter than 27 and starts with the substring "image_[" is excluded.

jag_scalar_prefix_filters: [ { key_prefix: "image_["  min_len: 27} ]

### DIFF
--- a/src/data_readers/data_reader_jag_conduit.cpp
+++ b/src/data_readers/data_reader_jag_conduit.cpp
@@ -163,6 +163,10 @@ void data_reader_jag_conduit::set_defaults() {
   m_scalar_keys.clear();
   m_input_keys.clear();
   m_uniform_input_type = false;
+  m_scalar_filter.clear();
+  m_scalar_prefix_filter.clear();
+  m_input_filter.clear();
+  m_input_prefix_filter.clear();
 }
 
 /// Replicate image processor for each OpenMP thread
@@ -266,7 +270,7 @@ void data_reader_jag_conduit::add_scalar_filter(const std::string& key) {
 }
 
 void data_reader_jag_conduit::add_scalar_prefix_filter(const prefix_t& p) {
-  m_scalar_prefix_filter.push_back(p);
+  m_scalar_prefix_filter.push_back((p.first.length() > p.second)? prefix_t(p.first, p.first.length()) : p);
 }
 
 void data_reader_jag_conduit::add_input_filter(const std::string& key) {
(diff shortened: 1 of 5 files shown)


src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -280,7 +326,11 @@ void data_reader_jag_conduit::set_all_scalar_choices() {
   conduit::NodeConstIterator itr = n_scalar.children();
```
davidHysom: Consider altering to use:
const std::vector<std::string> &names = n_scalar.shild_names();
for (auto t : names) {
  ...
}
My reasoning is that data_store can supply something similar:
const std::vector<std::string> &names = m_data_store->child_names("0/outputs/scalars");

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -789,7 +871,7 @@ std::vector<data_reader_jag_conduit::scalar_t> data_reader_jag_conduit::get_scal
     const conduit::Node & n_scalar = get_conduit_node(scalar_key);
```
davidHysom: FYI, I think the data_store version will be:
const char* data_store_jag_conduit::get_scalar(std::string scalar_key);
  - or possibly -
std::pair<const char*, uint> data_store_jag_conduit::get_scalar(std::string scalar_key);
where 'uint' is the number of bytes.
I'm thinking that data_store will return all values as const char*, which the data_reader can then cast them as desired. Let me know if you think this won't work, or something else works better.

JaeseungYeom:
I am making these changes. Thanks for the tip!

src/data_readers/data_reader_jag_conduit.cpp
JaeseungYeom: I see. It seems like there is no need for having separate get_scalar() and get_input().
We may be able to use std::pair<const char*, TypeID> instead. In which case, you need some special handling of CHAR8_STR_ID.

PR #462, opened 2018-06-21
HDF5 Checkpoint/Refactor

DylanMcKinney:
This PR adds an optional checkpoint mode which utilizes HDF5. If LBANN is built with HDF5 support this mode is used automatically, otherwise it falls back to the existing version. Since HDF5 does not require sequential read from file operations this checkpoint version can be easily extended to pick and choose which model parameters to reload. 

TODO:
- [x]  Fix report printed after save/reload for HDF5
- [x] Add recoverable failing with HDF5 (continue from checkpoint even if not all parameters load).
- [x] General clean up (addressing reviews etc)

Additionally this PR simplifies the existing checkpoint's read/write operations. Previously read/write functions were separated based on the datatype of the given parameter (eg: read_double). This PR transitions those functions to templated datatype agnostic functions (read/write_parameter). 

**Edit 6/28:** This PR is functionally complete, with results validated on Catalyst for all checkpoint modes, and Pascal (using deterministic lbann). If users build with a high enough version of HDF5 that version will be used. If not it will fall back to the original implementation. HDF5 must also be built with cpp support (easy way to attain this: `spack install hdf5%<compiler> +cxx` )

Future work will be adding the ability to load from saved checkpoints more granularly as users of this functionality need. Adding this support will be fairly trivial and can be done as use cases arise.


include/lbann/io/persist.hpp
```diff
@@ -115,6 +129,45 @@ class persist {
 
   bool write_datatype(persist_type type, const char *name, DataType  val);
   bool read_datatype (persist_type type, const char *name, DataType *val);
+#ifdef LBANN_HAS_HDF5   
+  template<typename T>
+  bool write_hdf5_parameter(H5::Group group_name, const char *name, T *val, H5::PredType hdf5_type) {
```
JaeseungYeom: "T *val" can be "const T *val"

include/lbann/io/persist.hpp
```diff
@@ -115,6 +129,45 @@ class persist {
...
+    H5::DataSpace dataspace = H5::DataSpace();
+    H5::Attribute attribute = group_name.createAttribute(name, hdf5_type, dataspace);
+    attribute.write(hdf5_type, static_cast<void*>(val));
```
JaeseungYeom: const void*

include/lbann/io/persist.hpp
```diff
@@ -115,6 +129,45 @@ class persist {
...
+  template<typename T>
+  bool read_hdf5_parameter(H5::Group group_name, const char *name, T *val) {
+    H5::Attribute attr = group_name.openAttribute(name);
+    H5::DataType type = attr.getDataType();
+    attr.read(type,val);     
+    return true;
+  }
+
+  template<typename T>
+  bool write_hdf5_array(H5::Group group_name, const char *name, std::vector<T> *val, H5::PredType hdf5_type) {
```
JaeseungYeom: const std::vector<T> *val

include/lbann/io/persist.hpp
```diff
@@ -94,12 +98,22 @@ class persist {
...
+  }
 
-  bool write_uint32(persist_type type, const char *name, uint32_t  val);
-  bool read_uint32 (persist_type type, const char *name, uint32_t *val);
+  bool write_int(persist_type type, const char *name, int val);
+  bool read_int(persist_type type, const char *name, int *val);
 
-  bool write_uint64(persist_type type, const char *name, uint64_t  val);
-  bool read_uint64 (persist_type type, const char *name, uint64_t *val);
+  bool write_long(persist_type type, const char *name, long  val);
```
JaeseungYeom: 'long' may not be 64-bit long.

include/lbann/io/persist.hpp
```diff
@@ -94,27 +98,80 @@ class persist {
...
+  
+  #ifdef LBANN_HAS_HDF5   
+  
+  H5::PredType cpp_to_hdf5(int val) { return H5::PredType::NATIVE_INT; } 
+  H5::PredType cpp_to_hdf5(double val) { return H5::PredType::NATIVE_DOUBLE; }
+  H5::PredType cpp_to_hdf5(float val) { return H5::PredType::NATIVE_FLOAT; }
+  H5::PredType cpp_to_hdf5(long val) { return H5::PredType::NATIVE_LONG; }
+  
+  template<typename T>
+  bool write_hdf5_parameter(H5::Group group_name, const char *name, T *val) {
```
JaeseungYeom: "const T *val"

include/lbann/io/persist.hpp
```diff
@@ -94,27 +98,80 @@ class persist {
...
+  template<typename T>
+  bool read_hdf5_parameter(H5::Group group_name, const char *name, T *val) {
+    H5::Attribute attr = group_name.openAttribute(name);
+    H5::DataType type = attr.getDataType();
+    attr.read(type,val);     
+    return true;
+  }
+
+  template<typename T>
+  bool write_hdf5_array(H5::Group group_name, const char *name, std::vector<T> val) {
```
JaeseungYeom: Probably, "const std::vector<T>& val" is better I think.

include/lbann/io/persist.hpp
```diff
@@ -94,27 +98,80 @@ class persist {
...
+    H5::DataSet ds = group_name.openDataSet(name);
+    H5::DataSpace dataspace= ds.getSpace();
+    ds.read(val.data(), H5::PredType::NATIVE_INT, dataspace);
+    return true;
+  }
 
-  bool write_datatype(persist_type type, const char *name, DataType  val);
-  bool read_datatype (persist_type type, const char *name, DataType *val);
+  H5::Group getGroup(std::string group_name);
+  bool write_hdf5_distmat(std::string group_name, const char *name, AbsDistMat *M, lbann_comm *comm);
```
JaeseungYeom: You can make comm pointer "const lbann_comm*" or simply replace it with "bool master" and pass comm->am_world_master()

DylanMcKinney:
Thank you for looking it over. I will make those clean up changes you recommended. Let me know if building this version fails. I have only tested with an HDF5 version I built with spack

JaeseungYeom:
When I tried to build on catalyst and surface using the build script scripts/build_lbann_lc.sh --with-hdf5, I am getting the error as below:
lbann/src/io/persist.cpp:382:23: error: 'class H5::H5File' has no member named 'exists'
   if(checkpoint_file->exists(group_name))

DylanMcKinney:
I see that when using the LC hdf5 version as well. Possibly a versioning thing, I will investigate

PR #463, opened 2018-06-21
Skip single-GPU allreduces

ndryden:
The allreduces should internally be skipping communication, but this also avoids unneeded stream syncs.


ndryden:
@naoyam Because right now the Aluminum NCCL backend runs on a separate stream, and so there needs to be a sync on the LBANN/Hydrogen stream to ensure the data is ready.

naoyam:
Then, why not making NCCL use the same stream as Hydrogen?

naoyam:
I talked with @ndryden, and realized that that would prevent overlapping of backprop computations and allreduces. So, a separate stream for allreduces would make everything proceed in a nonblocking way and still allow the overlaping.

PR #464, opened 2018-06-21
Remove cuDNN manager

timmoon10:
Previously, we determined whether a layer used a CPU or GPU implementation based on whether it had a null pointer to a cuDNN manger. This approach was deprecated with #250, which started templating layers by device type. The interaction between these two approaches has led to some interesting bugs.

This PR eliminates the cuDNN manager (closing #360). Utilities for CUDA and cuDNN are separated, so only layers that use cuDNN include the cuDNN utility header (I haven't tested, but we might be able to treat cuDNN as a truly optional GPU dependency). Instead of using the `cudnn::cudnn_manager` object, we now access a global cuDNN handle by calling `cudnn::get_handle()`.

Gradient checks come back clean with CPU and GPU runs and I don't see a performance regression on AlexNet on Surface gpgpu.


include/lbann/models/recurrent.hpp
```diff
@@ -26,6 +26,8 @@
 // recurrent .hpp .cpp - Recurrent neural network models
 ////////////////////////////////////////////////////////////////////////////////
 
+#if 0
```
naoyam: Why is this disabled?

model_zoo/lbann2.cpp
```diff
@@ -183,51 +183,78 @@ model * build_model_from_prototext(int argc, char **argv, lbann_data::LbannPB &p
...
-      std::cout << std::endl;
-    }
-
     // Save info to file; this includes the complete prototext (with any over-rides
     // from the cmd line) and various other info
     //save_session(comm, argc, argv, pb);
 
-    // Check for cudnn, with user feedback
-    cudnn::cudnn_manager *cudnn = nullptr;
+    // Report useful information
```
naoyam: Duplicate of lbann.cpp?

model_zoo/lbann_cycgan.cpp
```diff
@@ -192,51 +192,78 @@ model * build_model_from_prototext(int argc, char **argv, lbann_data::LbannPB &p
...
-      std::cout << std::endl;
-    }
-
     // Save info to file; this includes the complete prototext (with any over-rides
     // from the cmd line) and various other info
     save_session(comm, argc, argv, pb);
 
-    // Check for cudnn, with user feedback
-    cudnn::cudnn_manager *cudnn = nullptr;
+    // Report useful information
```
naoyam: Duplicate?

model_zoo/lbann_gan.cpp
```diff
@@ -185,51 +185,78 @@ model * build_model_from_prototext(int argc, char **argv, lbann_data::LbannPB &p
...
-      std::cout << std::endl;
-    }
-
     // Save info to file; this includes the complete prototext (with any over-rides
     // from the cmd line) and various other info
     //save_session(comm, argc, argv, pb);
 
-    // Check for cudnn, with user feedback
-    cudnn::cudnn_manager *cudnn = nullptr;
+    // Report useful information
```
naoyam: Duplicate?

src/models/recurrent.cpp
```diff
@@ -24,6 +24,8 @@
 // permissions and limitations under the license.
 ////////////////////////////////////////////////////////////////////////////////
 
+#if 0
```
naoyam: Why is this disabled?

src/proto/factories/model_factory.cpp
```diff
@@ -51,6 +51,7 @@ model* instantiate_model(lbann_comm* comm,
   if (type == "directed_acyclic_graph_model") {
     return new directed_acyclic_graph_model(comm, mini_batch_size, obj, opt);
   }
+#if 0
```
naoyam: Is this relevant to this PR?

include/lbann/models/recurrent.hpp
timmoon10: I forgot to uncomment, although I think we might want to eliminate the recurrent model. I've opened up issue #467.

model_zoo/lbann2.cpp
timmoon10: lbann2.cpp, lbann_cycgan.cpp, and lbann_gan.cpp are very similar to lbann.cpp. I know @samadejacobs is working on implementing the GAN stuff within the main LBANN pipeline. We should start thinking about how to replicate the functionality in lbann2.cpp so we can deprecate it.

samadejacobs: We should at least keep them for now until may be after SplashApp.

include/lbann/models/recurrent.hpp
naoyam: I think we should uncomment or address #467 first.

model_zoo/lbann2.cpp
timmoon10: I've opened #468 to discuss this.

include/lbann/models/recurrent.hpp
timmoon10: Recurrent models are removed in #471.

naoyam:
Looks good to me too.

PR #469, opened 2018-06-23
fix on the matrix view passing for numpy-based jag data reader

JaeseungYeom:
The same as the one applied in PR #458 for jag_conduit data reader


JaeseungYeom:
This should have been included in PR #458.

PR #470, opened 2018-06-23
Finer-grained NCCL streams

ndryden:
This does better synchronization between LBANN and Al/NCCL to avoid unnecessary stream synchronizations and improve overlap.

Specifically:
* `lbann_comm` keeps a small set of streams (currently five, arbitrarily-chosen number) to round-robin between for allreduces.
* When a GPU `nb_allreduce` is called, an event is used to synchronize between a NCCL stream and the LBANN stream. The allreduce is enqueued to run after the sync finishes. This avoids having to synchronize the LBANN stream to ensure data is ready.
* A separate event is enqueued after the allreduce, and is used to check for its completion. This avoids having to synchronize the entire stream (which would require waiting for other allreduces).
* Optimization steps are now done from end to start of the network. This should be roughly analogous to the order backprop is done in (assuming weights are added in the same order as layers). This better overlaps optimization with communication by waiting in the same order allreduces were started.

### DIFF
--- a/include/lbann/comm.hpp
+++ b/include/lbann/comm.hpp
@@ -56,18 +56,25 @@ using mpi_backend = ::Al::MPIBackend;
 #else
 using mpi_backend = lbann::Al::dummy_backend;
 #endif // LBANN_HAS_ALUMINUM
+using mpi_req_type = mpi_backend::req_type;
+static const mpi_req_type mpi_null_req = mpi_backend::null_req;
 /// @todo MPI-CUDA backend
 #if defined(LBANN_HAS_ALUMINUM) && defined(AL_HAS_NCCL)
 using nccl_backend = ::Al::NCCLBackend;
+// LBANN does its own synchronization on this.
+using nccl_req_type = cudaEvent_t;
+static const nccl_req_type nccl_null_req = (nccl_req_type) (-1);
 #else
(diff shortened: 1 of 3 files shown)


include/lbann/comm.hpp
```diff
@@ -1057,6 +1064,18 @@ class lbann_comm {
   using al_comms_key_type = std::pair<MPI_Comm, std::type_index>;
   using al_comms_val_type = std::unique_ptr<::Al::MPICommunicator>;
   std::map<al_comms_key_type, al_comms_val_type> m_al_comms;
+#ifdef AL_HAS_NCCL
+  /** Number of streams to round-robin between for NCCL allreduces. */
```
timmoon10: We should add a note explaining why a prime number is preferable to a nice, round number.

src/comm.cpp
```diff
@@ -88,6 +95,15 @@ lbann_comm::~lbann_comm() {
   }
 #ifdef LBANN_HAS_ALUMINUM
   m_al_comms.clear();
+#ifdef AL_HAS_NCCL
+  for (int i = 0; i < m_num_al_nccl_streams; ++i) {
+    CHECK_CUDA(cudaStreamDestroy(m_al_nccl_streams[i]));
```
timmoon10: Nitpick: we shouldn't throw exceptions in destructors.

PR #473, opened 2018-06-23
Cross entropy layer

timmoon10:
This PR implements a layer that takes predicted and ground truth distributions and outputs the cross entropy. This allows us to perform training with all major computation on GPUs, albeit without metrics. This is progress toward #193.

Note that this PR depends on #464, so the diff will be much less intimidating after that is merged.


include/lbann/layers/loss/cross_entropy.hpp
```diff
@@ -0,0 +1,139 @@
...
+    case data_layout::DATA_PARALLEL:
+      workspace.reset(new StarVCMat<Dev>(prediction.Grid(),
+                                         prediction.Root()));
+      break;
+    case data_layout::MODEL_PARALLEL:
+      workspace.reset(new StarMRMat<Dev>(prediction.Grid(),
+                                         prediction.Root()));
+      break;
+    default: LBANN_ERROR("invalid data layout");
+    }
```
ndryden: Why do we allocate a distributed matrix every time we do forward prop? That seems expensive.

include/lbann/layers/loss/cross_entropy.hpp
```diff
@@ -0,0 +1,139 @@
...
```
ndryden: Why do we allocate a new distributed matrix every time?

include/lbann/layers/loss/cross_entropy.hpp
timmoon10: It's not that bad since we only allocate memory equal to the local mini-batch size. It's less of a problem on GPUs since we use CUB's memory pool. Since the layer deallocates the memory at the end of its computation, it's also (very very barely) more efficient with GPU memory.

timmoon10:
@samadejacobs From the perspective of the layer, there's no way to distinguish predictions or ground truth.

timmoon10:
@ndryden Gradient checks come back clean. I don't notice significant performance differences on AlexNet and Resnet-50 when I disable metrics.

model_zoo/models/alexnet/model_alexnet.prototext
```diff
@@ -56,15 +56,28 @@ model {
...
   layer {
     name: "data"
-    children: "conv1 target"
+    children: "images labels"
     data_layout: "data_parallel"
     input {
       io_buffer: "partitioned"
     }
   }
+  layer {
```
bvanessen: Why are there identity layers added here?  It seems like you can just take the data directly to the consuming layer.

model_zoo/models/alexnet/model_alexnet.prototext
```diff
@@ -267,15 +301,28 @@ model {
...
+    name: "cross_entropy_eval"
+    parents: "cross_entropy"
+    data_layout: "model_parallel"
+    evaluation {}    
+  }
   layer {
     name: "target"
-    parents: "prob data"
+    parents: "prob labels"
```
bvanessen: Ah, are the identity layers added to allow the implicit split layer to be defined?

model_zoo/models/alexnet/model_alexnet.prototext
```diff
@@ -267,15 +301,28 @@ model {
...
     data_layout: "data_parallel"
     target {}
-  }
+  }
```
bvanessen: What is the role of the target layer with the cross entropy layers.

include/lbann/layers/loss/cross_entropy.hpp
bvanessen: Understood that it is not that bad, but why aren't these matrices just setup in the setup_data function?

model_zoo/models/alexnet/model_alexnet.prototext
```diff
@@ -12,7 +12,7 @@ model {
   ###################################################
 
   objective_function {
-    cross_entropy {}
+    layer_term { layer: "cross_entropy_eval" }
```
bvanessen: Maybe I just didn't see it in the code review, but how is the layer term used now?

model_zoo/models/alexnet/model_alexnet.prototext
timmoon10: Yes, perhaps making them split layers would be a little clearer.

model_zoo/models/alexnet/model_alexnet.prototext
timmoon10: Accuracy is still computed with the target layer. If we don't need to report accuracy, the target layer can be removed.

model_zoo/models/alexnet/model_alexnet.prototext
timmoon10: The evaluation layer was introduced in #276:
```
objective_function {
  layer_term {
    layer: "name_of_an_evaluation_layer"
    scale_factor: 1.0 // defaults to 1.0
  }
}
```
In the future, I'd like to add some logic to the model class so it can automatically insert evaluation layers where they're needed rather than forcing the practitioner to add them.

PR #484, opened 2018-06-25
Ltfb gan

samadejacobs:
Some code refactoring (and new codes/extension) in prep for LTFB GAN and in simplifying GAN implementation:
1. Add callback to replace weights in training
2. Refactor select_from_list in callback factory to return list as vector. This generalize to condition where we need element to be in order  
3. Refactor binary cross entropy and mse objective function as layers, mostly copied previous implementations in objective/loss function class/dir, previous implementations may need to be deprecated as soon asap
4. Add single LBANN-model GAN model prototext for JAG, this uses replace weights callback internally


include/lbann/layers/transform/binary_cross_entropy.hpp
```diff
@@ -0,0 +1,162 @@
...
+    for (size_t i=0; i<this->m_parent_layers.size(); i++) {
+      s << this->m_parent_layers[i]->get_name() << " " << this->m_parent_layers[i]->get_type() << " ";
+    }
+    s << " dataLayout: " << this->get_data_layout_string(get_data_layout());
+    return s.str();
+  }
+
+  protected:
+void fp_compute() {
+  //@todo: check that the order is not flipped
```
timmoon10: The order looks reasonable to me.

include/lbann/layers/transform/binary_cross_entropy.hpp
```diff
@@ -0,0 +1,162 @@
...
+                                              pred_val);
+      #endif // LBANN_DEBUG*/
+      DataType grad_val = DataType(0);
+      if (true_val > DataType(0)) {
+        grad_val += - true_val / pred_val;
+      }
+      if (true_val < DataType(1)) {
+        grad_val += (DataType(1) - true_val) / (DataType(1) - pred_val);
+      }
+      gradient_local(row, col) = grad_val;
```
timmoon10: We should multiply this by the previous error signal to get the correct error signal:
dL/d(pred) = dL/d(output) * (-label/pred + (1-label)/(1-pred))

We should also compute the gradient w.r.t. the ground truth:
dL/d(label) = dL/d(output) * (-log(pred) + log(1-pred))

include/lbann/layers/transform/mean_squared_error.hpp
```diff
@@ -0,0 +1,133 @@
...
+namespace lbann {
+
+/** Mean squared error loss function as a layer.
+ */
+template <data_layout T_layout = data_layout::DATA_PARALLEL, El::Device Dev = El::Device::CPU>
+class mean_squared_error_layer : public transform_layer {
+ public:
+
+  mean_squared_error_layer(lbann_comm *comm,
+                 cudnn::cudnn_manager *cudnn = nullptr)
```
timmoon10: The cuDNN manager was removed in #464.

include/lbann/layers/transform/mean_squared_error.hpp
```diff
@@ -0,0 +1,133 @@
...
+  const int local_width = predictions_local.Width();
+
+  // Compute squared sum of errors
+  #pragma omp parallel for collapse(2)
+  for (int col = 0; col < local_width; ++col) {
+    for (int row = 0; row < local_height; ++row) {
+      const DataType true_val = ground_truth_local(row, col);
+      const DataType pred_val = predictions_local(row, col);
+      const DataType error = true_val - pred_val;
+      output_local(row,col) += error * error;
```
timmoon10: `output_local(row,col) = error * error / predictions.Height();`

I also think it would be helpful if we just output the sum of these terms, so that the output is a single neuron containing the MSE.

include/lbann/layers/transform/mean_squared_error.hpp
```diff
@@ -0,0 +1,133 @@
...
+  const El::Int local_width = gradient_local.Width();
+
+  // Compute gradient
+  const DataType scale = DataType(2) / height; 
+  #pragma omp parallel for collapse(2)
+  for (El::Int col = 0; col < local_width; ++col) {
+    for (El::Int row = 0; row < local_height; ++row) {
+      const DataType true_val = ground_truth_local(row, col);
+      const DataType pred_val = predictions_local(row, col);
+      gradient_local(row, col) = (pred_val - true_val) * scale;
```
timmoon10: We need to multiply this by the previous error signal:
dL/d(pred) = 2 * (pred - label) / num_neurons * sum(dL/d(output))

We should also compute the gradient w.r.t. the ground truth:
dL/d(label) = 2 * (label - pred) / num_neurons * sum(dL/d(output))

src/proto/lbann.proto
```diff
@@ -739,6 +746,8 @@ message Layer {
    StopGradient stop_gradient = 319;
    Max max = 320;
    Min min = 321;
+   BinaryCrossEntropyLoss bce = 322;
```
timmoon10: I think it would be much clearer if we named these "binary_cross_entropy" and ""mean_squared_error" instead of using initials.

samadejacobs:
New commits were made specifically in LTFB callback, @timmoon10 please review. I have decided to deal with the BCE and MSE implementations as a separate PR after this is merge. BCE and MSE will be moved to the newly created /layer/loss/ dir at that time @timmoon10 feedback on BCE and MSE will be taken care of.

include/lbann/callbacks/callback_ltfb.hpp
```diff
@@ -50,8 +53,15 @@ class lbann_callback_ltfb : public lbann_callback {
 
   /** Constructor.
    *  @param round_size The number of minibatches in each round.
+   *  @param metric_mode  The expectation for a good tournament metric: increasing/descreasing 
+   *  @todo pair metric_mode with eval_metric
+   *  @param eval_metric Tournament evaluation metrics
+   *  @param selected_weights set of weights to exchange
    */
-  lbann_callback_ltfb(int round_size, lbann_summary* summarizer = nullptr);
+  lbann_callback_ltfb(int round_size, std::string metric_mode,
```
timmoon10: I think it's more natural to specify the metric mode with an enum or bool.

include/lbann/callbacks/callback_ltfb.hpp
```diff
@@ -50,8 +53,15 @@ class lbann_callback_ltfb : public lbann_callback {
...
+                      std::unordered_set<std::string> eval_metrics,
```
timmoon10: Does LTFB make sense with multiple metrics?

samadejacobs: Not at the moment, but there is a future possibility for something like mean/max/min of multiple metrics

include/lbann/callbacks/callback_ltfb.hpp
samadejacobs: Ok

include/lbann/callbacks/callback_ltfb.hpp
timmoon10: Would be simpler to make the user manually compute the mean/max/min? We already have layers that can do these operations.

PR #486, opened 2018-06-26
Weighted sum layer

timmoon10:
The weighted sum functionality in the sum layer was mistakenly removed in 50b68b03255fdb1079435084fcbb2569d6e3547c. However, the incorrect implementation is nice for basic sums since it uses matrix views to eliminate an unnecessary memory copy during backprop. I think the best solution is to move the weighted sum functionality into its own layer and leave the optimized implementation for basic sums.

### DIFF
--- a/include/lbann/layers/transform/weighted_sum.hpp
+++ b/include/lbann/layers/transform/weighted_sum.hpp
@@ -24,33 +24,40 @@
 // permissions and limitations under the license.
 ////////////////////////////////////////////////////////////////////////////////
 
-#ifndef LBANN_LAYER_SUM_HPP_INCLUDED
-#define LBANN_LAYER_SUM_HPP_INCLUDED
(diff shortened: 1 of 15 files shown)


include/lbann/layers/transform/weighted_sum.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+  /** Scaling factors for weighted sum. */
+  std::vector<DataType> m_scaling_factors;
+
+ public:
+  weighted_sum_layer(lbann_comm *comm,
+                     std::vector<DataType> scaling_factors)
+    : transform_layer(comm),
+      m_scaling_factors(scaling_factors) {
+    this->m_expected_num_parent_layers = -1; // No limit on parents
```
samadejacobs: same concern as before: enforce num_parents >=2

PR #494, opened 2018-06-27
Compiler test update -- Pascal

DylanMcKinney:
This PR updates the bamboo compiler tests, primarily making changes needed to add pascal builds to bamboo. It also simplifies the spack build script paths when operating in bamboo. This allows us to update the script without needing to propagate the new version numbers/anything else to our various pytest configuration files. 

This should not be merged until I have successfully used spack to build on pascal. I am making behind the scenes spack environment changes to try to make this happen, and will remove the in progress tag upon completion.

**Edit 8/7:** I am going to circumvent pascal spack usage for now, as that was in a lousy state last time I tried. This PR instead adds testing for pascal using the executable built by the build script. This will be a good work around until spack+pascal gets figured out.

### DIFF
--- a/bamboo/common_python/tools.py
+++ b/bamboo/common_python/tools.py
@@ -105,9 +105,9 @@ def get_command(cluster,
 
         # Create run command
         if command_allocate == '':
-            command_run = 'srun'
+            command_run = 'srun --mpibind=off'
         else:
-            command_run = ' srun'
+            command_run = ' srun --mpibind=off'
         option_num_processes = ''
         if num_processes != None:
             # --ntasks => Specify  the  number of tasks to run.
--- a/bamboo/compiler_tests/test_compiler.py
+++ b/bamboo/compiler_tests/test_compiler.py
@@ -40,36 +40,38 @@ def test_compiler_intel18_debug(cluster, dirname):
     skeleton_intel18(cluster, dirname, True)
 
 def skeleton_clang4(cluster, dir_name, debug, should_log=False):
-    if cluster in ['catalyst', 'pascal', 'quartz']:
+    if cluster in ['catalyst', 'quartz']:
         spack_skeleton(dir_name, 'clang@4.0.0', 'mvapich2@2.2', debug, should_log)
-        build_skeleton(dir_name, 'clang@4.0.0', 'mvapich2@2.2', debug, should_log)
+        build_skeleton(dir_name, 'clang@4.0.0', debug, should_log)
     else:
         pytest.skip('Unsupported Cluster %s' % cluster)
 
 def skeleton_gcc4(cluster, dir_name, debug, should_log=False):
-    if cluster in ['catalyst', 'pascal', 'quartz', 'ray']:
-        if cluster in ['catalyst', 'pascal', 'quartz']:
+    if cluster in ['catalyst', 'quartz', 'ray']:
+        if cluster in ['catalyst','quartz']:
             mpi = 'mvapich2@2.2'
+        elif cluster in  ['pascal', 'surface']:
+            mpi = 'mvapich2@2.2+cuda'
         elif cluster == 'ray':
             mpi = 'spectrum-mpi@2018.04.27'
         else:
             raise Exception('Unsupported Cluster %s' % cluster)
         spack_skeleton(dir_name, 'gcc@4.9.3', mpi, debug, should_log)
-        build_skeleton(dir_name, 'gcc@4.9.3', mpi, debug, should_log)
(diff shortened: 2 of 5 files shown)


bamboo/compiler_tests/test_compiler.py
```diff
@@ -85,7 +87,7 @@ def spack_skeleton(dir_name, compiler, mpi_lib, debug, should_log):
     debug_flag = ''
     if debug:
         debug_flag = ' -d'
-    command = '%s/scripts/spack_recipes/build_lbann.sh -c %s -m %s%s > %s 2> %s' % (
+    command = '%s/scripts/spack_recipes/build_lbann.sh -c %s -m %s %s > %s 2> %s' % (
```
forsyth2: `debug_flag` already includes the space

bamboo/common_python/tools.py
```diff
@@ -105,9 +105,9 @@ def get_command(cluster,
 
         # Create run command
         if command_allocate == '':
-            command_run = 'srun'
+            command_run = 'srun --mpibind=off'
```
DylanMcKinney: @forsyth2 why does this if/else logic do the same thing? I need to test this mpibind=off on catalyst before we merge this by the way. But we do need it (at least for now) on pascal/surface

forsyth2: It doesn't actually do the same thing. The `else` block puts a space in front

forsyth2: Because if there's a `salloc` command before, we need a space before `srun`.

bamboo/compiler_tests/test_compiler.py
```diff
@@ -107,17 +109,19 @@ def build_skeleton(dir_name, compiler, mpi_lib, debug, should_log):
...
     if cluster == 'ray':
         architecture += '_gpu_cuda-9.2.64_cudnn-7.0'
-    elif cluster in ['pascal', 'surface']:
-        architecture += "_gpu"
-    os.chdir('%s/bamboo/compiler_tests/builds/%s_%s_%s_%s_openblas_%s/build' % (dir_name, cluster, compiler, architecture, mpi_lib, build_type))
+    elif cluster == 'pascal':
+        architecture += '_gpu_cuda-9.1.85_cudnn-7.1'
+    elif cluster == 'surface':
+        architecture += '_gpu'
+    os.chdir('%s/bamboo/compiler_tests/builds/%s_%s_%s/build' % (dir_name, cluster, compiler, build_type))
```
forsyth2: Do we even need `architecture` anymore then? It looks like it's not used now.

bamboo/common_python/tools.py
DylanMcKinney: wow totally missed that subtlety. makes sense.

bamboo/compiler_tests/test_compiler.py
DylanMcKinney: Probably not. I will remove it before merging. There are probably several commented out or otherwise not used blocks that I will get rid of pre-merge

forsyth2: As discussed, I guess we'll comment out the `architecture` block.

bamboo/integration_tests/conftest.py
```diff
@@ -4,37 +4,28 @@ def pytest_addoption(parser):
...
-        default_exes['gcc4_debug'] = '%s/bamboo/compiler_tests/builds/%s_gcc-4.9.3_ppc64le_gpu_spectrum-mpi-2018.04.27_openblas_debug/build/model_zoo/lbann' % (default_dirname, cluster)
+    if cluster in ['ray']:
+        default_exes['gcc4'] = '%s/bamboo/compiler_tests/builds/%s_gcc-4.9.3_rel/build/model_zoo/lbann' % (default_dirname, cluster)
+        default_exes['gcc4_debug'] = '%s/bamboo/compiler_tests/builds/%s_gcc-4.9.3_debug/build/model_zoo/lbann' % (default_dirname, cluster)
 
     if cluster == 'surface':
         default_exes['gcc4'] = default_exes['default']
 
+    if cluster == 'pascal':
+        default_exes['gcc7'] = default_exes['default']
```
forsyth2: As discussed, let's make this `gcc4` not `gcc7`.

DylanMcKinney:
@forsyth2 Ok it sounds like nothing is glaringly wrong here. I am going to merge this, reactivate the pascal test on bamboo (which I had turned off due to generating 8 spack test failures), and kick off a clean build/nightly run on bamboo

PR #495, opened 2018-06-27
support for partitioning datasets amongst models.

davidHysom:
INITIAL COMMIT! NOT TESTED, NOT WORKING!
This PR adds functionality for partitioning a dataset amongst the models, i.e, each model gets a subset of the data. This is intended to support LTFB, etc. My plan is to make this general to all data_readers, by placing all modifications in data_reader_generic. This does not modify (or work with) data_store classes; that issue will be dealt with in a follow-on PR.


src/data_readers/data_reader.cpp
```diff
@@ -60,7 +60,15 @@ void generic_data_reader::setup() {
   shuffle_indices();
 }
 
+std::ofstream debug;
```
samadejacobs: debug? I suspect this will be removed before merge?

src/data_readers/data_reader.cpp
```diff
@@ -322,8 +331,150 @@ int generic_data_reader::get_next_position() const {
   }
 }
 
+void generic_data_reader::select_subset_of_data_partitioned() {
+/*
+char b[1024];
```
samadejacobs: cleanup

src/io/data_buffers/distributed_io_buffer.cpp
```diff
@@ -181,6 +181,10 @@ void lbann::distributed_io_buffer::calculate_num_iterations_per_epoch(int num_mo
     max_mini_batch_size = data_reader->get_num_data();
   }
```
samadejacobs: We should check somewhere here and in the partitioned_io_buffer that the max_mini_batch_size is properly computed for each model and that it is not aggregate (sum) for all models. The global/max_mini_batch_size should be as specified by the user in the prototext/cmd line.

PR #500, opened 2018-07-02
callback to synchronize selected layers

JaeseungYeom:
A callback to synchronize selected layers in particular prop steps.
It helps get a closer look at the performance of particular layers, especially marking the selected layers especially by `cudaProfilerStart()` and `cudaProfilerStop()` to profile GPU activities of those.
This inherits sync_layers callback.

The prototext interface is as below:
```
  callback {
    sync_selected {
      layer_to_sync {
        name: "conv1_head0"
        prop: Forward  # Forward, Backward or Both
      }
      layer_to_sync {
        name: "fc7"
        prop: Backward
      }

      async_gpus: false
      async_mpi: false

      cuda_profiler_setup {
        # set no_init to true to skip calling cudaProfilerInitialize(), which will avoid conflict with
        # an external profiler such as nvprof while still profiling selectively using the cudaProfilerStart()
        # and cudaProfilerStop() instrumented
        no_init: true  
        config_file: ""
        output_dir: "nvprof"
        output_mode: KeyValuePair # KeyValuePair or CSV
      }
    }
  }
```

### DIFF
--- a/include/lbann/callbacks/callback_sync_layers.hpp
+++ b/include/lbann/callbacks/callback_sync_layers.hpp
@@ -64,7 +64,7 @@ class lbann_callback_sync_layers : public lbann_callback {
   void on_forward_prop_end(model *m, Layer *l) override;
   void on_backward_prop_end(model *m, Layer *l) override;
     
- private:
+ protected:
   /** Whether to synchronize GPUs. */
   bool m_sync_gpus;
   /** Whether to do a global synchronization. */
@@ -72,7 +72,7 @@ class lbann_callback_sync_layers : public lbann_callback {
   /** Whether to only synchronize after the input layer. */
   bool m_only_input;
 
-  void do_sync(Layer *l);
+  virtual void do_sync(Layer *l);
 };
 
 }  // namespace lbann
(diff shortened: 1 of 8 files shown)


JaeseungYeom:
This is basically trying to provide the same benefit as sync_layers, but at the layers that users only have interest in. This is especially useful if you want to avoid large amount of profiling data or slow profiling that you do not really care about. Of course the added synchronization does not show what exactly happens. If profiling as is generates sufficient information to diagnose, this would not be needed in the first place. In my experience there are many non-ideal cases. For example, if there is a load imbalance issue hidden by other cost that overlaps, it may show misleading results. When you suddenly see something unexpected, you need means to check through assumptions quickly.

About merging with the profiler, I thought about that but I decided to simplify this PR, and rather deal with it in another PR. I also see the value of generalizing the profiler rather than nvprof only. score-p instrumented for all the backend would be very useful.
Using the currently profiler based on NVTX, users still cannot avoid generating data for uninterested region while they can label the code that would show up in the visual profiler. In addition, users need to manually instrument the code by inserting new lines and recompiling as opposed to typing a few lines in the prototext interface. Another reason was that I was not sure if the current way of push and pop based region marking would suffice was that if there is any chance of overlapping regions, we need the method based on start and stop by maintaining the start region id. This is simple to add but requires to change current sync interface parameter and change the unordered_set to unordered_map.

PR #503, opened 2018-07-03
Aluminum MPI-CUDA support

ndryden:
This adds support for Aluminum's MPI-CUDA backend to `lbann_comm`. The new Aluminum host-transfer allreduce will be used automatically for small-message allreduces in place of NCCL allreduces. This otherwise works the same way as the NCCL integration.

Passes gradient checking, ResNet-50 looks good.


src/comm.cpp
```diff
@@ -157,8 +157,16 @@ void lbann_comm::allreduce(AbsDistMat& m,
 #ifdef LBANN_HAS_GPU
   if (m.GetLocalDevice() == El::Device::GPU) {
 #ifdef AL_HAS_NCCL
-    // Force GPU matrices to use NCCL.
+    // We require NCCL for GPU matrices.
```
timmoon10: Aren't we able to handle the case where we have MPI-CUDA without NCCL, even if it's suboptimal?

src/comm.cpp
```diff
@@ -207,8 +225,16 @@ void lbann_comm::nb_allreduce(AbsDistMat& m,
...
```
timmoon10: Aren't we able to handle the case where we have MPI-CUDA without NCCL, even if it's suboptimal?

ndryden:
I decided to require NCCL for now to slightly simplify the code in LBANN.

I was also having some issues with the ring allreduce in the MPI-CUDA backend, so I decided to avoid those. Eventually, I plan to have Aluminum's backends handle the selection of the appropriate algorithm, which should simplify this code.

PR #506, opened 2018-07-06
Warning macro

timmoon10:
I've added an `LBANN_WARNING` macro that is helpful for non-fatal warnings and printf debugging. It reports the MPI rank, file, line number, and a message to the standard output. For instance:
```
some_function();
LBANN_WARNING("debug1", *m_comm);
horrible_bug_on_rank_1();
LBANN_WARNING("debug2", *m_comm);
another_function();
```
will output something like:
```
LBANN warning on rank 1 (lbann/some/file.cpp:42): debug1
LBANN warning on rank 0 (lbann/some/file.cpp:42): debug1
LBANN warning on rank 0 (lbann/some/file.cpp:44): debug2
srun: error: surface176: task 1: Segmentation fault (core dumped)
```

### DIFF
--- a/include/lbann/utils/exception.hpp
+++ b/include/lbann/utils/exception.hpp
@@ -36,14 +36,25 @@
 #include <exception>
 
 // Macro to throw an LBANN exception
-#define LBANN_ERROR(message)                                            \
+#define LBANN_ERROR(message)                                    \
+  do {                                                          \
+    std::stringstream ss_LBANN_ERROR;                           \
+    ss_LBANN_ERROR << "LBANN error "                            \
+                   << "(" << __FILE__ << ":" << __LINE__ << ")" \
+                   << ": " << (message);                        \
+    throw lbann::lbann_exception(ss_LBANN_ERROR.str());         \
+  } while (0)
+
+// Macro to print a warning to standard error stream.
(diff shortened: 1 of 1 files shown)


include/lbann/utils/exception.hpp
```diff
@@ -36,14 +36,25 @@
...
-                   << " (" << __FILE__ << ":" << __LINE__ << ")"        \
-                   << ": " << (message);                                \
-    throw lbann::lbann_exception(ss_LBANN_ERROR.str());                 \
-  } while(0)
+    std::stringstream ss_LBANN_WARNING;                                 \
+    ss_LBANN_WARNING << "LBANN warning "                                \
+                     << "on rank " << comm.get_rank_in_world() << " "   \
+                     << "(" << __FILE__ << ":" << __LINE__ << ")"       \
+                     << ": " << (message) << std::endl;                 \
+    std::cerr << ss_LBANN_WARNING.str();                                \
```
ndryden: If we're using `std::cerr`, we shouldn't need a `stringstream` and can just use `<<` on it directly.

timmoon10: We need to construct the string first to avoid race conditions.

PR #512, opened 2018-07-10
GPU evaluation layer

timmoon10:
This PR adds a GPU implementation of the evaluation layer. To simplify interactions with the evaluation layer (e.g. in objective function layer terms and layer metrics), the templated evaluation layer class inherits from an abstract evaluation layer class.

Gradient checks come back clean.


timmoon10:
We previously hard-coded the evaluation layer to run on CPU, so this PR will come into effect automatically. model_mnist_conv_graph.prototext is used in a gradient checking unit test.

PR #513, opened 2018-07-10
Feature omp tasking

bvanessen:
Hoisting the entire train_minibatch function to have a OpenMP parallel region.  Switching from omp parallel for to omp taskloop default(shared) constructs.


bvanessen:
I am still checking that we get the same answer and looking at performance numbers.

bvanessen:
* Yes, I have a separate branch that has the background I/O.  I wanted to keep this change as independent as possible.
* I did the conversion to omp critical in as direct / simple manner as possible.  They should be rewritten to use local variables and then sum at the end.
* We could use pragmas but I am not sure that it provides much benefit and it would have many permutations for the different nesting levels and private variables.

src/utils/summary.cpp
```diff
@@ -518,20 +534,27 @@ DataType lbann_summary::local_min(const Mat& mat) const {
   const El::Int ldim = mat.LDim();
   const DataType * __restrict__ mat_buf = mat.LockedBuffer();
   auto min = std::numeric_limits<DataType>::max();
+  int nthreads = omp_get_num_threads();
+  std::vector<EvalType> local_min(nthreads, std::numeric_limits<DataType>::max());
```
ndryden: This needs to be `DataType`, not `EvalType`, or it will fail to compile.

src/utils/summary.cpp
```diff
@@ -541,20 +564,27 @@ DataType lbann_summary::local_max(const Mat& mat) const {
   const El::Int ldim = mat.LDim();
   const DataType * __restrict__ mat_buf = mat.LockedBuffer();
   auto max = std::numeric_limits<DataType>::min();
+  int nthreads = omp_get_num_threads();
+  std::vector<EvalType> local_max(nthreads, std::numeric_limits<DataType>::min());
```
ndryden: This needs to be `DataType`, not `EvalType`, or it will fail to compile.

src/utils/summary.cpp
bvanessen: Fixed.

src/utils/summary.cpp
bvanessen: Fixed.

bvanessen:
@naoyam @ndryden @benson31 I have updated this PR to just be a refactoring into an LBANN_OMP_PARALLEL_FOR macro so that it will be easier to try out OMP taskloops, but it is not married to taskloops.  Furthermore, if OMP taskloops were enabled, the code cannot be built right now without OMP 5.0 support for reduction operators in the task / taskloop constructs.  I have a plan for how to support this in OMP 4.5 with a reduction data type, but that is not ready yet.  So for now, this PR is about refactoring the OMP interface into something easier to manipulate.

src/models/model.cpp
```diff
@@ -861,29 +861,38 @@ bool model::train_mini_batch() {
...
-    m->evaluate(execution_mode::training,
-                get_current_mini_batch_size());
-  }
 
-  // Update step
-  update_weights();
-  const bool finished = update_layers();
+  bool finished;
+
+#pragma omp parallel
```
ndryden: What impact is this going to have when we are not using taskloops? It seems like in that case we would need to enable nested parallelism, as this runs everything inside a `parallel` region and `LBANN_OMP_PARALLEL_FOR` uses the `omp parallel for` parallel loop construct.

bvanessen:
@benson31 Can you please take a look at this.  The new plan for this is to be a simple refactoring that would allow the exploration of taskloops in the future, but it is not enabled in the current set of changes.

bvanessen:
Okay.  Thanks.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

src/models/model.cpp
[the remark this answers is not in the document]
bvanessen: This has been disabled for the normal case.

PR #520, opened 2018-07-14
Model automatically adds evaluation layers

timmoon10:
The practitioner can now set up objective function layer terms and layer metrics to point to arbitrary layers. During the model setup phase, evaluation layers are automatically inserted as children of the original layers. Gradient checks pass and I don't see significant behavior changes in AlexNet and Resnet-50.

Note that there is ambiguous behavior if the original layer outputs different output to different children, e.g. the input layer outputs data samples and labels to its two children. The practitioner is still responsible for constructing an unambiguous model prototext.

### DIFF
--- a/include/lbann/models/model.hpp
+++ b/include/lbann/models/model.hpp
@@ -393,6 +393,14 @@ class model {
  private:
   /** Search layer graph and add all connected layers. */
   void add_connected_layers();
+  /** Insert evaluation layers where needed.
+   *  If an objective function layer term or a layer metric
+   *  corresponds to a layer that is not an evaluation layer, an
+   *  evaluation layer is added as a child of the original layer and
+   *  set as the corresponding layer to the layer term or layer
+   *  metric.
+   */
+  void add_evaluation_layers();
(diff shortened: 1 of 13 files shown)


include/lbann/models/model.hpp
```diff
@@ -393,6 +393,14 @@ class model {
  private:
   /** Search layer graph and add all connected layers. */
   void add_connected_layers();
+  /** Insert evaluation layers where needed.
```
samadejacobs: It seems term layers and metric layers are collectively called evaluation layer. I have no major problem with this, but something to note if we describing LBANN concept to the "practitioner or in documenting user API (though it seems this distinction is internal to the code?)

timmoon10: Right now the terminology is a little messy. A "layer term" is an objective function term and a "layer metric" is a metric; they are not layers. There is a special layer type, the "evaluation layer", that the layer term and layer metric can interact with.

In the future, I would be a fan of wiping out our objective functions and metrics and rewriting so that "objective function term" becomes identical to "layer term" and "metric" becomes identical to "layer metric".

PR #523, opened 2018-07-16
multi readers loading same data wastes time and memory

davidHysom:
attempting to add functionality: when using multiple models, as in cycle_gan,
all model's data readers load the same data (numpy and conduit). This multiplies
load time and memory by a factor of three. Attempting to hack so that only one
reader loads the data, and passes a reference to the other readers.


davidHysom:
This PR is no longer relevant, since we no longer load conduit data during calls to data_reader::load()

Issue #345, opened 2018-05-16
Total GPU memory footprint


bvanessen:
Examine how big the footprint is and are there some easy optimizations if networks are not able to fit on systems that they should.

timmoon10:
#356 should reduce GPU memory requirements by about half.

timmoon10:
I've heard complaints that some models don't fit on GPU, even though they did in the past. I suspect that making the CUB memory pool the default GPU memory allocator (https://github.com/LLNL/Elemental/pull/14) increased our memory usage. One solution is to avoid the CUB memory pool for GPU matrices that are only allocated/deallocated once, e.g. layer input/output tensors and weights.

ndryden:
Perhaps we can use #452 or something derived from it to identify where this is going wrong?

PR #531, opened 2018-07-20
Optimizations for GPU memory

timmoon10:
### Summary
Miscellaneous work (mostly in the base layer class) to optimize our GPU memory footprint without losing performance. See #345.

### Changelog
- GPU layers allocate tensor memory with the CUDA API, not the CUB memory pool.
- Layers that output tensor views (e.g. identity, reshape, split, sum) do not allocate memory for them.
- Added a virtual `construct_matrix()` function in the base layer class that derived classes can override to change matrix distributions and memory modes.
- Split up `fp_setup_data()` in the base layer class into `fp_setup_inputs()` and `fp_setup_outputs()`. Derived classes can override `fp_setup_outputs()` to avoid having to allocate unnecessary memory.
- Split up `bp_setup_data()` in the base layer class into `bp_setup_gradient_wrt_outputs()` and `bp_setup_gradient_wrt_inputs()`.
- `setup_data()` in the base layer class calls `fp_setup_inputs()` and friends to reduce code duplication.
- Fixed some bugs involving matrix alignment.
- Dropout no longer outputs views during evaluation. This gave a tiny performance improvement during evaluation at the expense of code complexity. Something something premature optimization.
- Slice and concatenation layers output views during forward prop and back prop, when possible.

### Results
Experiment results with Resnet-50 (1 Surface gpgpu node, 4 procs/node; mini-batch size 32; ImageNet-10):

Commit | GPU 0 usage at epoch 0 | GPU 0 usage at epoch 9 | Epoch 0 time | Average epoch time (excluding epoch 0)
--------|--------------|--------------|----------|--------------
3982aaa (develop) | 4.68 GiB | 5.79 GiB | 89.8s | 70.9s
1c5f057 | 3.93 GiB | 4.71 GiB | 84.7s | 70.6s
fa6bdf4 | 3.16 GiB | 4.26 GiB | 97.3s | 71.1s

I suspect the memory difference between epoch 0 and epoch 9 is because the convolution workspace is allocated during the first convolution layer's forward prop.

### Validation
- Gradient checks on model_zoo/tests/model_mnist_conv_graph.prototext come back clean.
- Reasonable results with AlexNet on ImageNet-10 (15.2%/59.5% top-1/top-5 training accuracy, 30%/76.2% top-1/top-5 validation accuracy, 12.5s training time for epoch 0).


ndryden:
What does the memory usage look like at epoch 1? I mostly want to make sure that the increase in memory usage really is just allocating convolution workspaces and other similar things, and that we don't have a memory leak.

Can we run some larger-scale performance tests (multi-node, ImageNet-1k) to make sure that there are no performance regressions?

timmoon10:
The memory usage at epoch 1 is the same as at epoch 9.

timmoon10:
Experiment results with Resnet-50 (16 Pascal nodes, 2 procs/node; mini-batch size 256; ImageNet-1K):

Commit | GPU 0 usage at epoch 0 | GPU 0 usage at epoch 1 | Epoch 0 time | Epoch 1 time
--------|--------------|--------------|----------|--------------
325e5b7c (develop) | 4.65 GiB | 5.78 GiB | 21.7 min | 19.7 min
92896429 | 3.25 GiB | 4.37 GiB | 24.2 min | 22 min

ndryden:
I'm a little concerned that this appears to result in a ~10% performance hit. Do we know where that is coming from? (It shouldn't be dropout, since that's not used in ResNet-50.)

timmoon10:
Repeating the last experiment a few times, it looks like the performance hit is real:

 Branch | Epoch 0 time | Epoch 1 time
-------|----------|----------
develop | 1112.15s | 1131.47s
develop | 1147.27s | 1061.5s
develop | 1146.46s | 1061.5s
gpu-memory-optimization | 1652.79s | 1491.19s
gpu-memory-optimization | 1571.84s | 1446.91s

I'll need to do some profiling to figure out what's going on.

timmoon10:
I've run the experiment several more times, but made sure to alternate between the develop branch and this PR:

Branch | Epoch 0 time | Epoch 1 time
-------|----------|----------
gpu-memory-optimization | 1430.66s | 1001.56s
develop | 1430.24s | 1001.55s
gpu-memory-optimization | 1523.79s | 1121.02s
develop | 1151.16s | 1030.9s
gpu-memory-optimization | 1072.83s | 1067.02s
develop | 1065.87s | 1035.5s
gpu-memory-optimization | 1045.94s | 1153.23s
develop | 1138.97s | 1151.65s

On average, this PR has an epoch 1 time that's ~30s slower. However, the sample standard deviation in these runs is ~60s. I didn't see anything unexpected when profiling with nvprof, so I'm inclined to blame the performance regression on weird variability on Pascal's network.

ndryden:
This agrees with my experiments.

include/lbann/layers/regularizers/dropout.hpp
```diff
@@ -201,7 +168,7 @@ class dropout : public regularizer_layer {
     // Do nothing if dropout is disabled
     const auto& mode = this->m_model->get_execution_mode();
     if (mode != execution_mode::training || m_keep_prob < EvalType(0)) {
-      El::LockedView(output, input);
+      El::Copy(input, output);
```
bvanessen: So just to clarify, you are reverting this back to a copy to simplify implementation despite a small performance loss.  Is that correct?

PR #535, opened 2018-07-23
Added thread local buffers to the data reader class so that it can

bvanessen:
decode image data into a temporary space that is declared once.  Also
updated the load_file function to load data with a single read rather
than using an interator.

### DIFF
--- a/include/lbann/utils/file_utils.hpp
+++ b/include/lbann/utils/file_utils.hpp
@@ -63,28 +63,7 @@ bool check_if_file_exists(const std::string& filename);
 bool check_if_dir_exists(const std::string& dirname);
 bool create_dir(const std::string output_dir);
 
-
-/// Load a file into a buffer
-template <typename T = std::vector<char> >
-inline bool load_file(const std::string filename, T& buf) {
-  std::ifstream file(filename, std::ios::binary);
-  if (!file.good()) {
-    return false;
-  }
-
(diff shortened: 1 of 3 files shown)


include/lbann/utils/file_utils.hpp
```diff
@@ -65,7 +65,7 @@ bool create_dir(const std::string output_dir);
 
 
 /// Load a file into a buffer
-template <typename T = std::vector<unsigned char> >
+template <typename T = std::vector<char> >
```
timmoon10: Do we need this template?

src/data_readers/data_reader.cpp
```diff
@@ -58,6 +58,13 @@ void generic_data_reader::setup() {
   set_initial_position();
 
   shuffle_indices();
+
+  m_thread_buffer.resize(omp_get_max_threads(), std::vector<char>());
+#pragma omp parallel
```
timmoon10: It seems simpler to loop through the buffers rather than launching an OpenMP section.

bvanessen: Good point.

include/lbann/utils/file_utils.hpp
bvanessen: No, the previous version of code made use of the template, but we don't need it.

PR #539, opened 2018-07-24
Stack trace refactor

timmoon10:
This PR does code cleanup and output prettying for the exception and stack trace reporting:
- Error messages and stack traces are printed in a unified block, which helps readability and reduces the  window for asynchronous printf jumbling.
- Signal handling functionality is enabled by default in main driver.
- Added a self-contained `stack_trace::get()` function that writes the stack trace to a string. This string is printed to `std::cerr` or to file elsewhere as needed, e.g. when an exception is caught.
- Separated declaration and implementation of `exception` class.
- Renamed `lbann_exception` to `exception` to reduce redundancy. Note that this class is in the `lbann` namespace, so it is clearly distinguished from `std::exception`. `lbann_exception` is retained as a typedef for the sake of backward compatibility.
- Added helper function `get_rank_in_world()` to safely determine MPI rank at any point in program.

Sample error message:
```text
Caught signal 11 (segmentation violation) on rank 1
Stack trace:
   0: lbann::stack_trace::get[abi:cxx11]()
   1: lbann::exception::exception(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)
   2: /usr/workspace/wsa/moon13/src/lbann/build/gnu.Release.catalyst.llnl.gov/lbann/build/liblbann.so(+0x9c0976) [0x2aaaab68f976] (could not find stack frame symbol)
   3: /usr/lib64/libpthread.so.0(+0xf680) [0x2aaaacb27680] (could not find stack frame symbol)
   4: lbann::Layer::Layer(lbann::lbann_comm*)
   5: lbann::generic_input_layer::generic_input_layer(lbann::lbann_comm*, int, std::map<execution_mode, lbann::generic_data_reader*, std::less<execution_mode>, std::allocator<std::pair<execution_mode const, lbann::generic_data_reader*> > >, bool, data_reader_target_mode)
...
   9: lbann::proto::construct_model(lbann::lbann_comm*, std::map<execution_mode, lbann::generic_data_reader*, std::less<execution_mode>, std::allocator<std::pair<execution_mode const, lbann:\
:generic_data_reader*> > > const&, lbann_data::Optimizer const&, lbann_data::Model const&)
  10: /usr/workspace/wsa/moon13/src/lbann/build/gnu.Release.catalyst.llnl.gov/lbann/build/model_zoo/lbann() [0x4046ee] (could not find stack frame symbol)
  11: __libc_start_main (demangling failed)
  12: /usr/workspace/wsa/moon13/src/lbann/build/gnu.Release.catalyst.llnl.gov/lbann/build/model_zoo/lbann() [0x4051c0] (could not find stack frame symbol)
```

### DIFF
--- a/include/lbann/comm.hpp
+++ b/include/lbann/comm.hpp
@@ -1186,6 +1186,12 @@ void lbann_comm::broadcast(const int root, T* data, const int count, const El::m
 template<>
 void lbann_comm::broadcast<std::string>(const int root, std::string& str, const El::mpi::Comm c);
 
+/** Get the current rank within MPI_COMM_WORLD.
+ *  This function is safe to call even if MPI has not initialized or
+ *  has been finalized. In either case it returns a negative value.
+ */
+int get_rank_in_world();
+
 } // namespace lbann
 
 #endif  // LBANN_COMM_HPP_INCLUDED
--- a/include/lbann/utils/exception.hpp
+++ b/include/lbann/utils/exception.hpp
@@ -22,17 +22,14 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
 // implied. See the License for the specific language governing
 // permissions and limitations under the license.
-//
-// lbann_exception .hpp .cpp - LBANN exception class
 ////////////////////////////////////////////////////////////////////////////////
 
-#ifndef LBANN_EXCEPTION_HPP_INCLUDED
-#define LBANN_EXCEPTION_HPP_INCLUDED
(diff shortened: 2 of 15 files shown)


davidHysom:
I'll take a close look tomorrow. You probably know the cmd line option "--stack_trace_to_file" causes each process that encounters an lbann_exception to print the trace to file. My thinking was, some procs in some models may encounter an exception, and others may not, so have each print to a separate file. However, I didn't want to make that the default behavior, since if running with thousands of procs you'd get ... thousands of files. That said, the current output is very messy when an exception is encountered and lbann is run with > 1 proc.

PR #543, opened 2018-07-25
Added a thread safe set of buffers for OpenCV decoding in the image

bvanessen:
data readers.  This buffer is pre-allocated and passed into the image
utilities for intermediate decoding.


src/data_readers/data_reader_image.cpp
```diff
@@ -136,6 +136,13 @@ void image_data_reader::load() {
   select_subset_of_data();
 }
 
+void image_data_reader::setup() {
+  generic_data_reader::setup();
+
+  using InputBuf_T = lbann::cv_image_type<uint8_t>;
+  m_thread_cv_buffer.resize(omp_get_max_threads(), cv::Mat(1, get_linearized_data_size(), InputBuf_T::T(1)));
```
ndryden: I'm not sure this is correct. The elements in `m_thread_cv_buffer` will be initialized by copying the input value. I believe the copy constructor for `cv::Mat` does not duplicate the underlying buffer, but instead creates a reference to it. (See [here](https://docs.opencv.org/3.4/d3/d63/classcv_1_1Mat.html#a294eaf8a95d2f9c7be19ff594d06278e).) So we probably want to explicitly create separate matrices for each thread.

include/lbann/data_readers/data_reader_imagenet.hpp
```diff
@@ -31,6 +31,7 @@
 
 #include "data_reader_image.hpp"
 #include "cv_process.hpp"
+#include "opencv_extensions.hpp"
```
JaeseungYeom: This might not be necessary although it does not hurt anything. cv_process.hpp already includes cv_transform.hpp which includes opencv_extensions.hpp

src/data_readers/data_reader_image.cpp
JaeseungYeom: m_thread_cv_buffer seems to reserve the buffer space for each thread. 
However, my concern is the hard-coded type uint8_t. The entire pipeline has been designed to handle various types. Now, this breaks that.

src/data_readers/cv_utils.cpp
```diff
@@ -102,9 +102,13 @@ cv::Mat cv_utils::lbann_imread(const std::string& img_file_path, int flags, std:
   const cv::Mat inbuf(1, buf.size(), InputBuf_T::T(1), buf.data());
 
   // decode the image data in the memory buffer
-  cv::Mat image = cv::imdecode(inbuf, cv::IMREAD_ANYCOLOR | cv::IMREAD_ANYDEPTH);
-
-  return image;
+  if(cv_buf != nullptr) {
```
JaeseungYeom: It might be the OpenCV's bug if it does not work without this conditional.
According to https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/loadsave.cpp +856, the returned Mat is the same as *cv_buf if it is not null.

src/data_readers/data_reader_image.cpp
JaeseungYeom: Since all the images we deal with using data_reader_image derivatives are of uint8_t, we can address this issue later. However, notice that there are other types of images. The image from jag simulation is floating point type although we do not use any data_reader_image derivative for that. I have dealt with X-ray images of 16bit/32bit integers in other project.

ndryden: `m_thread_cv_buffer` reserves a `cv::Mat` for each thread, yes. My concern is that all of those `cv::Mat`s share the same underlying buffer, unless this is reinitialized elsewhere and I've missed it.

JaeseungYeom: This loading creates a new buffer per image, and the subsequent preprocessing steps modifies the buffer in-place if possible. So, the buffers should rather be per minibatch item than per thread. If there is double buffering used, then per minibatch and per buffering stage.

ndryden: No, that's not the problem. Unless I've misunderstood something, this is what is going on: The problem is that every thread will share the same underlying buffer. The `std::vector` constructor being used (`vector(size_type count, const T& value)`) constructs the container by copying `value` once for each of the `count` entries. The copy constructor for `cv::Mat` does a shallow copy and just grabs a reference to the underlying buffer passed in. Hence, every element in that vector will point to the same underlying matrix.

JaeseungYeom: Oh, I see that!

src/data_readers/data_reader_image.cpp
```diff
@@ -136,6 +136,17 @@ void image_data_reader::load() {
...
+  generic_data_reader::setup();
+
+  using InputBuf_T = lbann::cv_image_type<uint8_t>;
+  auto cvMat = cv::Mat(1, get_linearized_data_size(), InputBuf_T::T(1));
+  m_thread_cv_buffer.resize(omp_get_max_threads());
+  for(int tid = 0; tid < omp_get_max_threads(); ++tid) {
+    m_thread_cv_buffer[tid] = cvMat.clone();
+  }
+}
+
```
bvanessen: Nikoli, I believe that this addresses your concern.  One question for you or Jae-Seung is if any allocation from the clone is properly cleaned up when the vector is destroyed.  I believe that it should.

PR #548, opened 2018-07-27
Added an optimization to use an asynchronous copy from the previous

bvanessen:
layer activations to the input matrices when copying from CPU to GPU
and the matrix distributions are the same.


bvanessen:
I have not been able to test this yet since there is a problem building the head of the develop branch.  I will be out tomorrow, so if any of you wants to pull it down and give it a try, feel free.

bvanessen:
I don't think that this is a premature optimization.  When looking at latest timing breakdowns, this cost is actually dominating the total runtime and improvements in the I/O time are hard to see with respect to this overhead.

timmoon10:
VTune is misleading with GPU code timing since all the time will appear during stream synchronizations. The cudaMemcpy looks expensive because it is waiting for backprop to finish. Looking at the GPU timings with nvprof shows that the CPU-GPU memory transfers are very small.

bvanessen:
Tim and Nikoli,  do you have any further thoughts on this PR.  I have run it with no performance regression and would like to get it merged knowing that it is not a current bottleneck, but may be one in the future.

ndryden:
Two comments:
1. @timmoon10 has added a wrapper around CUDA (`lbann::cuda::event_wrapper`) that should probably be used.
2. Someone should double-check this, but I think the current implementation introduces a race condition. As it stands in this PR, a layer doing an async copy will wait for the event, enqueue an asynchronous copy from a parent layer's output into its input, and then record the event. But the event does not prevent the parent layer from writing into its output before the copy completes. The event synchronization needs to occur before the _parent_ writes to its output, not before the copy begins.

PR #549, opened 2018-07-27
Update profiler

ndryden:
Various improvements to make the profiler callback more useful for LBANN, primarily focusing on nvprof.
* Optimizer regions are now annotated.
* Layer names and batch/epoch numbers are added to the appropriate regions.
* Fixed up annotating layers with different colors.
* Name the Hydrogen stream so this shows up in nvprof explicitly.
* Device synchronization is now optional.

When not using device synchronization (now the default), profiling will not perturb the application much, and can be used to understand CPU/GPU overlap pretty well. Using device synchronization is helpful when you want to understand the on-GPU performance of particular operations.

### DIFF
--- a/src/callbacks/profiler.cpp
+++ b/src/callbacks/profiler.cpp
@@ -33,22 +33,26 @@
 #include <scorep/SCOREP_User.h>
 #elif defined(LBANN_NVPROF)
 #include "nvToolsExt.h"
+#include "nvToolsExtCuda.h"
+#include "nvToolsExtCudaRt.h"
 #include "cuda_runtime.h"
 #endif
 
 namespace {
 #if defined(LBANN_SCOREP)
-static void prof_region_begin(const char *s, int c) {
+static void prof_region_begin(const char *s, int, bool) {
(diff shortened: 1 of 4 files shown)


src/callbacks/profiler.cpp
```diff
@@ -61,83 +65,124 @@ static void prof_region_begin(const char *s, int c) {
...
 }
 
 int lbann_callback_profiler::get_color(Layer *l) {
   const std::string &lname = l->get_type();
-  int idx = 4;
-  if (lname == "fully_connected") {
-    idx = 5;
-  } else if (lname == "convolution") {
+  int idx = 5;
+  if (lname == "fully connected") {
```
timmoon10: I don't think we should hardcode a colorscheme explicitly for Resnet-50. If we want all layers of the same type to have the same color, perhaps we can hash the layer type name.

ndryden: This wasn't really intended for ResNet-50 so much as just emphasizing commonly-used layers and layers that tend to be computationally expensive. (Admittedly, those and ResNet-50 tend to coincide...)

Issue #550, opened 2018-07-27
Support for distributed matrices with arbitrary alignment


timmoon10:
Reasons for supporting distributed matrices with arbitrary alignment:
- The concatenation and slice layers sometimes output matrix views. Forcing alignment with rank 0 can cause unnecessary data reshuffles.
- It makes our code more robust and easier to maintain, especially when it does weird stuff with views and submatrices.
- We'll need to get more explicit about our data layouts as we implement more intricate data layouts, e.g. for distributed convolution.

Problems:
- I think the base layer class has the machinery to handle arbitrary alignments (circa #524), but this hasn't been properly tested.
- The weights store values in matrices aligned to rank 0. This causes problems in the fully connected layer when the input and output tensors have unusual alignment.

PR #551, opened 2018-07-27
Align all matrices to rank 0.

timmoon10:
There are still some bugs that show up when distributed matrices have arbitrary alignment (#550), so this PR forces all distributed matrices to use the default alignment.


timmoon10:
This PR has been superseded by #557.

PR #557, opened 2018-07-31
Weights refactor

timmoon10:
### Goal
To allow weights to accommodate arbitrary matrix alignment (#550), obviating the need for #551 and closing #530.

### Changelog
- The weight matrix distribution must be fully specified before it is setup.
- Weights are initialized separately from layers. The `Layer::setup_data()` function now only sets the dimensions and matrix distribution of the weights.
- Weights store their initializers, optimizers, and matrices in `unique_ptr`s (#155).
- Renamed `fan_in_fan_out_initializer` to `variance_scaling_initializer` to match TensorFlow's terminology.

### Results
I reproduce @samadejacobs's correct results on the P1 Combo model (#530):
<details> <summary> 4 Pascal nodes, 2 procs/node </summary>

```text
Submitted batch job 73510
Mon Jul 30 18:09:07 PDT 2018


==============================================================
STARTING lbann with this command line:
/usr/workspace/wsa/moon13/src/lbann/build/gnu.Release.pascal.llnl.gov/lbann/build/model_zoo/lbann --model=/usr/workspace/wsa/moon13/src/lbann/model_zoo/models/candle/pilot1/combo.prototext --num_epochs=10 --disable_cuda=0 --optimizer=/usr/workspace/wsa/moon13/src/lbann/model_zoo/optimizers/opt_adam.prototext --reader=/usr/workspace/wsa/moon13/src/lbann/model_zoo/data_readers/data_reader_candle_pilot1_combo.prototext 

protobuf_utils::verify_prototext; starting verify for 1 models
	Max Parallel I/O Fetch: 8 (Limited to # Processes)

writing options and prototext to file: data.prototext

Hardware properties (for master process)
  Processes on node          : 2
  OpenMP threads per process : 18
  GPUs on node               : 1

Build settings
  Type     : Release
  Aluminum : NOT detected
  CUDA     : detected
  cuDNN    : detected
  CUB      : detected

GPU settings
  CUDA         : enabled
  cuDNN        : enabled
  MV2_USE_CUDA : 1

Model settings
  Models              : 1
  Processes per model : 8
  Grid dimensions     : 2 x 4

Loading file: //p/lscratchf/brainusr/datasets/cancer/pilot1/lm_genes/combo_x_train_lmg.0.txt
Set data filename: //p/lscratchf/brainusr/datasets/cancer/pilot1/lm_genes/combo_x_train_lmg.0.txt
[...]

PR #560, opened 2018-08-01
Manually setting thread affinity on Pascal.

timmoon10:
I recently saw a 50x performance degradation on Pascal when I built with Aluminum. Profiling shows that the slowdown happens in NCCL allreduce kernels, which end up taking >99% of the GPU runtime. Whatever the cause, I recovered expected performance when I manually set thread affinity to accommodate Pascal's weird topology (see the timing experiments in #557).


experiments/run_lbann_experiment.sh
```diff
@@ -269,13 +269,13 @@ esac
...
-        case ${HAS_GPU} in
-            YES|yes|TRUE|true|ON|on|1)
-                case ${CLUSTER} in
-                    surface|pascal|ray)
-                        MPIRUN="${MPIRUN} --nvidia_compute_mode=default"
-                        ;;
-                esac
+        case ${CLUSTER} in
+            surface|ray)
+                MPIRUN="${MPIRUN} --nvidia_compute_mode=default"
```
ndryden: Does Ray have `--nvidia_compute_mode`? I think the GPUs are always in default mode and there's no way to change it.

PR #563, opened 2018-08-02
Signal handler conforms to POSIX C standard.

### LINKED ISSUE #556: tmux crashes LBANN
The signal handler in LBANN crashes upon receiving signal 28 (window size change). It appears that tmux on Ray sends this signal occasionally (even when the window size hasn't actually changed), which causes LBANN to abort.

timmoon10:
The signal handler will only handle signals that cause abnormal termination according to the POSIX C standard (http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html). This should close #556 since signal 28 isn't in the POSIX C standard.

@ndryden brought up a good point that the POSIX C standard is not always followed strictly. For instance, the standard specifies that SIGTRAP causes a termination, but it's something used by gdb. That being said, I think it is unwise to choose which signals to handle based on folklore and I'm unable/unwilling to maintain such code. If there's a better spec (TODO: check the Linux spec), it is a perfectly fine design question about which one to follow.

### DIFF
--- a/include/lbann/utils/stack_trace.hpp
+++ b/include/lbann/utils/stack_trace.hpp
@@ -40,18 +40,19 @@ std::string get();
 
 /** Register signal handler.
  *  Initializes a signal handler that prints an error message and
- *  stack trace to the standard error stream when a signal is
- *  detected. If desired, it also writes to the file
- *  "stack_trace_rank<MPI rank>.txt". 
+ *  stack trace to the standard error stream when a fatal signal is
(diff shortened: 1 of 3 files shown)


timmoon10:
Looking like our signal handling matches Linux (https://github.com/torvalds/linux/blob/ef46808b79ebbf48587c556436e5281ec51d09b5/include/linux/signal.h#L292), except for the optional/deprecated signals they support.

Issue #91, opened 2017-06-28
Integer data type


timmoon10:
As of d2c414a1846b072ea6607c85a0254c5b1277d4e7, `int` is the standard integer data type. However, we may need 64-bit integers for very large matrices. Variables susceptible to overflow should be changed to `El::Int`. As a rule of thumb, indices into Elemental matrices should be `El::Int`s.

ndryden:
With these changes, I'm getting a lot of warnings about narrowing casts.

ndryden:
(Background/refresher: [fundamental C++ types](http://en.cppreference.com/w/cpp/language/types) and [fixed-width integer types](http://en.cppreference.com/w/cpp/types/integer))

Just to add some context for this discussion, these are our main sources/issues with integers, as far as I can recall:
* Elemental defines `El::Int` to be either `int` or `long long` depending on our compile-time flags. (There is also `El::Unsigned`.) We want to use `El::Int` when interacting with Elemental matrices.
* The C++ STL (e.g. `std::vector`) tends to use `size_t` for just about any size-related quantity. This can lead to us either narrowing a `size_t` or comparing signed/unsigned integers.
* cuDNN expects parameters to be `int`.
* MPI expects parameters to be `int`.

Ideally, we should come up with a consistent use of integers that satisfies all of these.

PR #567, opened 2018-08-03
Replacing El::Int with an IntType typedef.

timmoon10:
This is preparatory work toward using a consistent integer datatype (see Issue #91). This pull requests adds the `IntType` type, which is just a typedef of `El::Int`. Then it find-replaces every instance of `El::Int` into an `IntType`. There are no other changes.

The next step is to change the `int` type into `IntType`, which will be much more non-trivial.

### DIFF
--- a/include/lbann/base.hpp
+++ b/include/lbann/base.hpp
@@ -72,6 +72,9 @@ using StarMRMat  = El::DistMatrix<lbann::DataType, El::STAR, El::MR  , El::ELEME
 using DistMat         = MCMRMat<El::Device::CPU>;
 using Mat        = El::Matrix<lbann::DataType, El::Device::CPU>; // Temporarily define as CPUMat
 
+/** Integer type. */
+using IntType = El::Int;
+
 // Datatype for model evaluation
 // Examples: timing, metrics, objective functions
 using EvalType = double;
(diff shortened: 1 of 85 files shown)


ndryden:
Just to link them together (since Github apparently didn't do so automatically), see my review comments in #568 for some philosophical ramblings about this.

bvanessen:
I haven’t had a chance to review this yet but we may want to have multiple integer types to optimize width. Let’s discuss this later this week.


Brian C. Van Essen

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Aug 10, 2018, at 5:34 PM, Nikoli Dryden <notifications@github.com<mailto:notifications@github.com>> wrote:


Just to link them together (since Github apparently didn't do so automatically), see my review comments in #568<https://github.com/LLNL/lbann/pull/568> for some philosophical ramblings about this.

timmoon10:
The consensus opinion seems to be that using `IntType` consistently in our interfaces is a good idea, but internal integer datatypes should be dictated by the implementation requirements (`El::Int` for accessing Elemental matrices, `size_t` for STL objects, int for tight loops). In any case, this is not the best time to mess with types and we will defer this code debt to a later time.

PR #568, opened 2018-08-03
 Replacing int with IntType in communicator header.

timmoon10:
### Purpose
This is work toward using a consistent integer datatype (see Issue #91) and a followup to PR #567.

### Questions
- @ndryden Will these changes affect the quantizer?
- @davidHysom Some of the data stores call `all_gather`, which now takes `IntType` vectors instead of `int` vectors. I've kludged until it compiles, but I'm not sure if these changes will work. What command do you usually test your data store with?

### Validation
- Gradient checks on model_zoo/tests/model_mnist_conv_graph.prototext come back clean. (09a1574)
- Reasonable results with AlexNet on ImageNet-10 (14.7%/58.5% top-1/top-5 training accuracy, 7.88 training objective function, 30.0%/76.2% top-1/top-5 validation accuracy, 7.24 validation objective function, 15.4s training time for epoch 0 on 1 Pascal node). (09a1574)
- Observed a performance regression with Resnet-50 on ImageNet-1K (average of 1976 sec/epoch with 16 Pascal nodes, excluding first epoch). The mini-batch times have a large variance (mean 0.40s, stdev 0.20s), so this needs to be investigated more carefully. (09a1574)


timmoon10:
Rerunning my performance tests (Resnet-50 on ImageNet-1K; 16 Pascal nodes), the performance regression goes away. I get 1062 sec/epoch on the develop branch (d9923f8) and 1087 sec/epoch on this PR (3327a45).

PR #570, opened 2018-08-07
Fix build failure on Surface

timmoon10:
The TOSS 3 update on Surface broke the build script, so this attempts to fix some of the issues. LBANN now builds correctly on Surface, but crashes in MVAPICH during runtime:
```text
[surface146:mpi_rank_0][cudaipc_init_dynamic] [src/mpid/ch3/channels/mrail/src/gen2/ibv_cuda_ipc.c:486] cuda failed with 11
```
The corresponding CUDA error is `cudaErrorInvalidValue`, so some API call is being fed something wrong.

Of course, it may be worthwhile to completely abandon Surface now that it its reaching the end of its life. If we take this option, we should scrub it from our build scripts, experiment scripts, and documentation to make clear we no longer support it.

### DIFF
--- a/experiments/run_lbann_experiment.sh
+++ b/experiments/run_lbann_experiment.sh
@@ -271,10 +271,10 @@ case ${SCHEDULER} in
         MPIRUN="srun --nodes=${NUM_NODES} --ntasks=${NUM_PROCS}"
         case ${CLUSTER} in
             surface|ray)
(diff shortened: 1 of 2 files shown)


timmoon10:
AlexNet now runs correctly on Surface, provided you use @ndryden's custom build of MVAPICH (see #572 and the mailing list).

PR #303, opened 2018-04-25
DenseNet Model

forsyth2:
Add DenseNet model to LBANN's model zoo.

PR #574, opened 2018-08-08
Reconcile redundant weight values after each epoch

timmoon10:
Many of our weights are stored redundantly, especially for data-parallel layers, and we perform computation independently during the update step. This is very nice for reducing communication, but it opens the door for numerical errors to accumulate. Pessimistically, training a model on ImageNet-1K involves ~5000 update steps per epoch, ~100 epochs, and ~1e-7 numerical error at each step. I think it would be good practice to periodically make sure all of our processes actually have the same weight values. This is roughly as expensive as a backprop step, but it only happens once per epoch.

Hopefully this will improve the divergence we're seeing in DenseNet (see #303), but it's far from certain. We got reasonable results on AlexNet and Resnet-50, so numerical error in these cases were relatively small or not disruptive.

### Validation
- Gradient checks on model_zoo/tests/model_mnist_conv_graph.prototext come back clean. (c46d86d)
- Reasonable results with AlexNet on ImageNet-10 (14.8%/58.8% top-1/top-5 training accuracy, 7.88 training objective function, 27.7%/75.4% top-1/top-5 validation accuracy, 7.23 validation objective function, 19.1s training time for epoch 0 on 1 Pascal node). (c46d86d)
- Observed a performance regression with Resnet-50 on ImageNet-1K (average of 1230 sec/epoch with 16 Pascal nodes, excluding first epoch). I need to run experiments on the develop branch to see whether this regression is meaningful. (c46d86d)

### DIFF
--- a/include/lbann/models/model.hpp
+++ b/include/lbann/models/model.hpp
@@ -344,6 +344,11 @@ class model {
   virtual void update_weights();
   /** Update layers step. */
   virtual bool update_layers();
+  /** Reconcile weight values.
+   *  If weight values are duplicated across multiple processes, they
+   *  are set to the average across the processes.
+   */
+  virtual void reconcile_weight_values();
 
   ////////////////////////////////////////////////////////////
   // Callbacks
--- a/include/lbann/weights/weights.hpp
+++ b/include/lbann/weights/weights.hpp
@@ -181,6 +181,17 @@ public:
   /** Set an entry in the weight matrix. */
   void set_value(DataType value, int row, int col);
 
+  /** Reconcile weight values.
+   *  If weight values are duplicated across multiple processes, they
+   *  are set to the average across the processes.
+   */
+  void reconcile_values();
+  /** Asynchronously reconcile weight values.
+   *  If weight values are duplicated across multiple processes, they
+   *  are set to the average across the processes.
+   */
+  void reconcile_values(Al::request& req);
+  
   // -----------------------------------------------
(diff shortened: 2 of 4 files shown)


ndryden:
Before merging this, I would be interested in seeing some experiments to identify whether this sort of divergence is actually occurring. That way we can know whether our divergence in DenseNet is actually being (at least partly) caused by this, or if we need to keep looking. Something like `sum(abs(W1 - W2))` for every pair of weight matrices might be a good metric to start with.

If this _is_ a problem, it would be good to know!

timmoon10:
Rerunning my performance tests (Resnet-50 on ImageNet-1K; 16 Pascal nodes), the performance regression goes away. I get 1094.96 sec/epoch on the develop branch (d9923f8) and 1095.07 sec/epoch on this PR (c46d86d).

timmoon10:
Some of my experiments (Resnet-50; ImageNet-!K, 16 Pascal nodes) with the develop branch (d9923f8) suggest that this type of numerical error is not a major issue. As far as I can tell, redundantly stored weights are bit-perfect identical between processes. Perhaps the gradients are small enough that any numerical errors get truncated off when added to the weights.

That being said, I think this PR is a good idea for our peace of mind. @ndryden has observed that ring allreduces yield large numerical errors when the number of processes is large. This PR offers some protection if the error is large enough to survive an optimization step but not large enough to cause divergence after a few iterations.

PR #575, opened 2018-08-08
Use the exported conduit targets in the FindCONDUIT module

benson31:
Modify the FindCONDUIT module to work with the exported conduit targets. Hopefully addresses some of @JaeseungYeom build issues.

### DIFF
--- a/cmake/modules/FindCONDUIT.cmake
+++ b/cmake/modules/FindCONDUIT.cmake
@@ -3,61 +3,38 @@
 #   - CONDUIT_LIBRARIES
 #   - CONDUIT_INCLUDE_DIRS
 #
-# Also creates an imported target CONDUIT
 
-# Find the header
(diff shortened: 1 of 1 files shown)


JaeseungYeom:
It does not find the conduit header.

include/lbann/data_readers/data_reader_jag_conduit.hpp:35:31: fatal error: conduit/conduit.hpp: No such file or directory
 #include "conduit/conduit.hpp"

cmake/modules/FindCONDUIT.cmake
```diff
@@ -3,61 +3,38 @@
 #   - CONDUIT_LIBRARIES
 #   - CONDUIT_INCLUDE_DIRS
```
JaeseungYeom: It seems that CONDUIT_INCLUDE_DIRS  is removed.

benson31:
I was hoping the exported target would include the proper include paths, like they should. Apparently not. I'll add the include path search back first thing in the morning.

PR #579, opened 2018-08-12
Integrates autoencoder with cycle GAN in one training cycle

samadejacobs:
This PR integrates (variational) auto encoder with cycle gan in one training cycle. 
To achieve this, support for copying activations from one model to the other was added


model_zoo/lbann_cycgan.cpp
```diff
@@ -79,38 +83,71 @@ int main(int argc, char *argv[]) {
...
       model_1->copy_trained_weights_from(model3_weights);
+      
+      //Optionally evaluate on pretrained autoencoder
+      if(ae_model != nullptr && ae_proxy_model != nullptr){
+        if(master) std::cout << " Copy trained weights from autoencoder to autoencoder proxy" << std::endl;
+        auto ae_weights = ae_model->get_weights();
+        ae_proxy_model->copy_trained_weights_from(ae_weights);
+        if(master) std::cout << " Copy needed activations from cycle GAN" << std::endl;
+        auto model2_layers = model_2->get_layers();
+        ae_proxy_model->copy_activations_from(model2_layers);
```
timmoon10: I'm unsure if this will do what we're expecting. Every layer overwrites its activations during its forward prop step, so I suspect that the copies done here will be overwritten when `evaluate` is called.

timmoon10: This isn't entirely relevant here since we only evaluate on `ae_proxy_model`, but the `copy_activations_from` function can completely disrupt our automatic differentiation.

samadejacobs: @timmoon10 Changes have been made based on your feedback.

PR #590, opened 2018-08-20
enable partitioning 1B jag dataset amongst data readers

davidHysom:
To turn on the partitioning, include the cmd line option --jag_partitioned


davidHysom:
I believe this is no longer relevant, since we're loading data from file as needed (exactly what we do with imagenet).

bvanessen:
Okay.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #596, opened 2018-08-22
Temporarily move Bamboo tests from Spack to superbuild

DylanMcKinney:
Currently all our tests depend on spack builds completing to actually run. This PR modifies the compiler test to instead use the superbuild. This fixes the issue we see often where our build actually works fine, but fails in bamboo for spack specific reasons. I have been running these changes to test them out, and will continue to do so but they seem to work as intended. I will let someone else make the call on whether to merge this, as tomorrow is my last day. 

This also adds a clang build to the ray testing, as during the convo that spawned this pull request it was mentioned as a case of interest. I believe the ray clang compiler test was failing when I ran it, I will double check and post the build error in my weekly blog post.

### DIFF
--- a/bamboo/compiler_tests/test_compiler.py
+++ b/bamboo/compiler_tests/test_compiler.py
@@ -1,10 +1,14 @@
 import pytest
 import os, re, subprocess
 
-def test_compiler_build_script(cluster, dirname):
-    output_file_name = '%s/bamboo/compiler_tests/output/build_script_output.txt' % (dirname)
-    error_file_name = '%s/bamboo/compiler_tests/error/build_script_error.txt' % (dirname)
-    command = '%s/bamboo/compiler_tests/build_script.sh > %s 2> %s' % (dirname, output_file_name, error_file_name)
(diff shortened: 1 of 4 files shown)


bamboo/compiler_tests/test_compiler.py
```diff
@@ -1,10 +1,14 @@
 import pytest
 import os, re, subprocess
 
-def test_compiler_build_script(cluster, dirname):
-    output_file_name = '%s/bamboo/compiler_tests/output/build_script_output.txt' % (dirname)
-    error_file_name = '%s/bamboo/compiler_tests/error/build_script_error.txt' % (dirname)
-    command = '%s/bamboo/compiler_tests/build_script.sh > %s 2> %s' % (dirname, output_file_name, error_file_name)
+def build_script(cluster, dirname, compiler, debug):
+    if debug:
```
forsyth2: For clarity, I would pass `debug` as a bool, and then in this logic, do:
```
if debug:
    build = 'debug'
   debug_flag = '--debug'
```

PR #598, opened 2018-08-23
Refresh jag conduit

JaeseungYeom:
- Refreshes the existing jag_consuit data reader such that it can handle the latest JAG simulation data format.
- Handles multi-channel image data and adds the relevant prototext support
- Allows users to choose which emi images to use by listing the substring of the data path that represents the associated measurement view and the time index in the data reader prototext.
- Supports distributed data file ingestion, with which each data reader instance loads and consumes a unique set of data files without redistributing the samples after loading. Therefore, this suits best for a single epoch scenario. For multiple epoch, we would normally want the data exchange between data readers, which is not supported in this scheme. However, as the amount of local data increases, this scheme would provide a reasonable approximation.
- Allows shuffling of local samples
- Allow consistent shuffling of the list of data files among the data reader instances
- Fixes the bug in load_conduit(), which failed to populate the valid sample map with multiple data files
- Changes the type of valid sample index map from unordered_map<int,string> to vector<string> as I do not see the need for using a map here. However, if there is something I am missing, please let me know.
- updated the data reader and cycle gan model prototex files
- support select_subset_of_data(), and use_unused_index_set(). The options in the prorotext interface, 'validation_percent', 'absolute_sample_count', and 'percent_of_data_to_use', are applied to the amount of locally available valid samples. Therefore, the meaning of 'absolute_sample_count' is different from that of the base data reader as it represents the absolute local sample count to use for each data reader instance. In addition, an example to use a separate data reader for validation steps is provided in the 'data_reader_jad_conduit.prototext'.

The follow-up PRs that further enhances the capability and performance are merged into this PR.
- https://github.com/JaeseungYeom/lbann/pull/2
- https://github.com/JaeseungYeom/lbann/pull/3


src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -526,6 +551,95 @@ void data_reader_jag_conduit::check_input_keys() {
...
+  const size_t num_valid_samples = get_num_valid_local_samples();
+
+  MPI_Comm comm = m_comm->get_model_comm().comm;
+  const size_t my_rank = static_cast<size_t>(m_comm->get_rank_in_model());
+  const size_t num_ranks = static_cast<size_t>(m_comm->get_procs_per_model());
+
+  // Find the minimum of the number of valid samples locally available
+  unsigned long long n_loc = static_cast<unsigned long long>(num_valid_samples);
+  unsigned long long n_min = static_cast<unsigned long long>(num_valid_samples);
+  MPI_Allreduce(&n_loc, &n_min, 1, MPI_UNSIGNED_LONG_LONG, MPI_MIN, comm);
```
davidHysom: You could use: template <typename T>
  T allreduce(T snd, const El::mpi::Comm c, El::mpi::Op op = El::mpi::SUM)

JaeseungYeom: Thanks. I was looking for this.

JaeseungYeom:
Absolutely. Especially, for the channel-wise normalization scheme I am thinking of.

JaeseungYeom:
Now, a separate image can be generated per channel. Regarding the matter of having each reader  load a separate subset of data, I did not have to change the base generic_data_reader class. I only need to override a couple of methods. Of course, my approach is not ideal. However, I believe it is reasonable for single-epoch cases with a large number of samples, and as simple as it can be without requiring data exchange among data readers. Let me know if you see any particular issue with the approach.

davidHysom:
I think these will cause your reader to misbehave:
src/io/data_buffers/distributed_io_buffer.cpp
  data_reader->set_reset_mini_batch_index(m_comm->get_rank_in_model());
src/io/data_buffers/partitioned_io_buffer.cpp
  data_reader->set_sample_stride(num_parallel_readers_per_model);

JaeseungYeom:
The follow up PR is https://github.com/JaeseungYeom/lbann/pull/2

samadejacobs:
I am just getting to this PR, what is the update? I noted these two lines from your comment:

1. Supports distributed data file ingestion, with which each data reader instance loads and consumes a unique set of data files without redistributing the samples after loading. Therefore, this suits best for a single epoch scenario. For multiple epoch, we would normally want the data exchange between data readers, which is not supported in this scheme. However, as the amount of local data increases, this scheme would provide a reasonable approximation.
2. Allows shuffling of local samples

The two items (distributed data file ingestion and shuffling of local (per model) samples) are appropriate  for LTFB. It seems jag_store provides exact (two capabilities above), which of the two is ready (bug free) and which one should I use?  I also believe/assume that distributed data file ingestion should provide solution to the I/O problem we are currently having with loading 1M JAG except if I am mistaken.

davidHysom:
Capability 1. depends on PR 690, which had not been merged.

I can discuss more tomorrow, but am taking a vacation today.

JaeseungYeom:
I think this is ready to merge, and going to do that soon as Sam and Brian suggested. Merge first and repent later.

PR #607, opened 2018-09-05
add capability to pull num_neruons from data reader for data_reader_jag_conduit_hdf5

davidHysom:
Changes you will need to make:

  in data_reader_jag_conduit_hdf5.prototext add the following:
    use_images: true
    use_scalars: false
    use_inputs: true
  #NOTE: if more than one of these is true, the linearized_sizes
  #      will be summed
  #      If all are 'false' (or not specified, which is the same
  #      thing) the functionality is ignored

  in cycgan*.prototext:

  layer {
    freeze: true
    fully_connected {
      #num_neurons: 49152
      get_image_dimension_from_reader: true        #this is the magic!
      has_bias: true
    }
    name: "gen1fc4"
    data_layout: "data_parallel"
    weights: "gen1fc4linearity"
    parents: "gen1relu3"
  }

  layer {
    name: "slice_data"
    data_layout: "data_parallel"
    parents: "data"
    children: "image_data_dummy param_data_id"
    slice {
      #slice_points: "0 49152 49157"
      get_slice_points_from_reader: "images inputs"    #this is the magic!
    }

### DIFF
--- a/src/proto/factories/layer_factory.cpp
+++ b/src/proto/factories/layer_factory.cpp
@@ -74,13 +74,36 @@ Layer* construct_layer(lbann_comm* comm,
   // Fully connected layer
   if (proto_layer.has_fully_connected()) {
     const auto& params = proto_layer.fully_connected();
-    int num_neurons = params.num_neurons();
-    if (proto_layer.num_neurons_from_data_reader()) {
-      const auto dr  = lbann::peek_map(data_readers, execution_mode::training);
-      if (!dr) {
-        LBANN_ERROR("training data reader does not exist!");
+    int num_neurons = 0;
+    if (params.get_input_dimension_from_reader() 
+        || params.get_image_dimension_from_reader()
+        || params.get_scalar_dimension_from_reader())
+       {
+       const auto dr1  = lbann::peek_map(data_readers, execution_mode::training);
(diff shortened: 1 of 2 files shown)


JaeseungYeom:
I am not sure why Sam has not approved it yet. I think it is fine but the method is specific to a particular reader that required conduit. So, it should probably be within the macro guard for conduit.

src/proto/factories/layer_factory.cpp
```diff
@@ -74,13 +74,36 @@ Layer* construct_layer(lbann_comm* comm,
...
-      const auto dr  = lbann::peek_map(data_readers, execution_mode::training);
-      if (!dr) {
-        LBANN_ERROR("training data reader does not exist!");
+    int num_neurons = 0;
+    if (params.get_input_dimension_from_reader() 
+        || params.get_image_dimension_from_reader()
+        || params.get_scalar_dimension_from_reader())
+       {
+       const auto dr1  = lbann::peek_map(data_readers, execution_mode::training);
+       lbann::data_reader_jag_conduit_hdf5 *dr = dynamic_cast<lbann::data_reader_jag_conduit_hdf5*>(dr1);
```
JaeseungYeom: lbann::data_reader_jag_conduit_hdf5 is only defined when LBANN_HAS_CONDUIT is set

PR #608, opened 2018-09-06
Nicer pattern for GPU activation layers

timmoon10:
### Description
This PR introduces a nicer pattern to implement entry-wise layers on the GPU (mainly the activation layers). Instead of having to implement and launch a custom CUDA kernel for each layer, we just have to implement structs with operator() as a device function (one for forward prop and one for backprop). We then call a templated function that launches CUDA kernels internally, similar to Thrust's transform function (https://thrust.github.io/doc/group__transformations.html). I've implemented ReLU with this new interface as a test case.

### Question
I propose that we change the CPU activation layers to match this implementation. The CPU version is a bit nicer since we just have to implement two virtual functions, but CUDA kernels don't support polymorphism. My attempts to kludge it by passing around device function pointers have not had any success.

### Validation
- Reasonable results with AlexNet on ImageNet-10 (14.3%/58.8% top-1/top-5 training accuracy, 7.88 training objective function, 27.7%/75.4% top-1/top-5 validation accuracy, 7.22 validation objective function, 40.8s training time for epoch 0 on 1 Pascal node). (630d4ff)
- Got ~1140 sec/epoch with Resnet-50 on ImageNet-1K (16 Pascal nodes). I'll need to run more experiments to see if this is a performance regression or just Pascal being slow. (630d4ff)

### DIFF
--- a/include/lbann/layers/activations/relu.hpp
+++ b/include/lbann/layers/activations/relu.hpp
@@ -27,7 +27,7 @@
 #ifndef LBANN_LAYER_ACTIVATION_RELU_HPP_INCLUDED
 #define LBANN_LAYER_ACTIVATION_RELU_HPP_INCLUDED
 
-#include "lbann/layers/activations/activation.hpp"
+#include "lbann/layers/layer.hpp"
 
 namespace lbann {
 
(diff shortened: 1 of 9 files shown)


timmoon10:
Looking at the Bamboo logs on Pascal, this PR got 12.9 sec/epoch (AlexNet; 1% of ImageNet-1K; 16 Pascal nodes) while last night's develop build got 14.9 sec/epoch. Any performance changes are much smaller than Pascal's noise. In any case, I think maintainability and extensibility are more important for these layers, as long as we don't get a major performance regression. Our bottleneck is not GPU performance, but the fact that we have no GPU implementation for many layers.

Doing gradient checking with sigmoid is a good idea. I'll implement the sigmoid layer with this framework.

timmoon10:
Gradient checks with the sigmoid layer come back clean. I've also taken the liberty of adding some infrastructure so that activation layers have a CPU implementation similar to the GPU implementation.

timmoon10:
I'd like to merge this first and do the rest of the operations later.

bvanessen:
That sounds good to me.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #622, opened 2018-09-16
GPU memory usage callback prints aggregated statistics.

naoyam:


src/callbacks/callback_gpu_memory_usage.cpp
```diff
@@ -27,6 +27,30 @@
...
+namespace {
+template <typename T>
+T get_mean(const std::vector<T> &v) {
+  return std::accumulate(v.begin(), v.end(), 0.0) /
+      v.size();
+}
+template <typename T>
+T get_median(const std::vector<T> &v) {
+  std::vector<T> tmp = v;
+  std::sort(tmp.begin(), tmp.end());
```
ndryden: You might consider using `std::nth_element` to do only a partial sort instead of completely sorting.

PR #623, opened 2018-09-17
Unary math layers

timmoon10:
### Description
This PR uses the framework from #608 to implement a number of basic unary math functions, both on CPU and GPU:
- Logical: not
- Sign: abs, negative, sign
- Rounding: round, ceil, floor
- Power: reciprocal, square, sqrt, rsqrt
- Exponential and logarithmic: exp, expm1, log, log1p
- Trigonometric: cos, sin, tan, acos, asin, atan
- Hyperbolic: cosh, sinh, tanh, acosh, asinh, atanh

This list contains all of the floating-point unary operations amongst [TensorFlow basic math functions](https://www.tensorflow.org/api_guides/python/math_ops#Basic_Math_Functions), except those involving the error function, gamma function, and zeta function. It also contains most of the unary operations in the [C math header](http://www.cplusplus.com/reference/cmath/). Several layers that were perviously considered "activation layers" are now "math layers", i.e. atan, tanh, exp, log.

### Validation
- Gradient checks on Catalyst with the tanh and sqrt layers come back clean.
- A gradient check on Catalyst with the exp layer fails, but manual inspection finds that the result is reasonable (our numerical differentiation assumes the fifth derivative is small, which doesn't hold for the exponential function).
- Gradient checks on Pascal get the same results as on Catalyst.


samadejacobs:
Seems Brian and I were reviewing/writing at the same time.

Issue #624, opened 2018-09-17
Framework for layer unit tests


timmoon10:
### Problem
As we move into more interesting models, the number of supported layers has been growing steadily. I've also added 27 new layers in #623. We need a more systematic way to test all of our layer implementations.

### Proposed pathway
- [x] Implement a weights initializer that allows the user to directly specify weight values.
- [x] Implement a "weights layer" that simply outputs the values of a set of weights.
- [x] Implement a callback that throws an error if the objective function value is not within a specified range.
- [x] Implement an L2 norm layer. This is helpful for gradient checking since its error signal varies with its input values.
- [ ] Implement simple unit tests for as many layers as possible. For most layers, it should be sufficient to have weights layers of size O(5), the tested layer, and an L2 norm layer. The objective function callback should check whether the objective function value matches an externally computed value. If the layer is differentiable, gradient checking should be performed.
- [x] Once this framework is established, a unit test will be required for acceptance of new layers. In most cases this will just be a matter of copy-pasting another unit test, swapping out the tested layer, and updating the expected objective function value.

### Questions
- Each test should be small and fast, but will we run into any issues if we suddenly plonk down ~50 of them? Will we want to tinker with our Bamboo interface? Pinging @forsyth2.

ndryden:
I just wanted to say that I endorse having more unit tests.

forsyth2:
I don't think adding more fast unit tests will place much strain on Bamboo. Just make sure you add them in `bamboo/unit_tests` so Bamboo can see them. Tests currently in that directory should serve as good examples.

Issue #53, opened 2017-03-14
Create a self contained "smoke" test example


bvanessen:
We need a compiled binary that we can run to check that an installation of LBANN is correct.  The current idea is to create an autoencoder with synthetic data that can run for a fixed number of iterations and achieve a fixed reconstruction accuracy.  This test can then be setup for both model and data parallelism and can be used as a benchmark for installation on new systems and for validating the installation.

timmoon10:
I think this is satisfied with the unit tests in model_zoo/tests/layer_tests.

PR #626, opened 2018-09-18
Ingredients for layer unit tests

timmoon10:
This PR makes progress toward a simple framework for layer unit testing (#624):
- Implemented a "value initializer", a weights initializer that populates entries based on a user-provided list of values.
- Implemented a "weights layer" that outputs its weights tensor.
- Implemented an L2 norm squared layer on CPU and GPU.
- Removed the "L2 loss" layer, which had a somewhat misleading name (I blame TensorFlow) and was implemented as an activation layer. I've gone ahead and made changes to the affected models (mostly CycleGAN models).
- Added a unit test for MNIST softmax classification. The prototext had already been written, but it wasn't registered with Bamboo.
- Added a unit test for the L2 norm squared layer. This is a template for future layer unit tests.
- The LBANN executables now exit with a failure status if an exception is caught. PyTest was previously reporting false successes.

The unit tests pass when I run manually on Pascal and Catalyst, but I'd like to see how they perform under Bamboo.


timmoon10:
Note that a combination of the weights layer, value initializer, and synthetic data reader may be helpful for creating a deterministic, self-contained smoke test (#53). The limitation is that each mini-batch sample will be identical.

PR #627, opened 2018-09-19
fix the compilation error by adding the macro guard

JaeseungYeom:
The data_reader_jag_conduit_hdf5 is only defined when the macro definition LBANN_HAS_CONDUIT is set.
When it is not set, this creates a compilation error as the definition of the reader class is undefined.

### DIFF
--- a/src/proto/factories/layer_factory.cpp
+++ b/src/proto/factories/layer_factory.cpp
@@ -96,6 +96,10 @@ Layer* construct_layer(lbann_comm* comm,
        if (params.get_scalar_dimension_from_reader()) {
          num_neurons += scalar_dim;
        }
+    #else
+      err << "get_*_dimension_from_reader() not suppoeted";
+      LBANN_ERROR(err.str());
+      return nullptr;
     #endif // defined(LBANN_HAS_CONDUIT)
     } else {
       num_neurons = params.num_neurons();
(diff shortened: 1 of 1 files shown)


davidHysom:
Hm, I was working on the same thing this morning. I think there should be an #else <throw exception> because if one of the booleans is true and lbann was not compiled --with-conduit, then it's most certainly an error. Do you agree?

src/proto/factories/layer_factory.cpp
```diff
@@ -239,6 +242,7 @@ Layer* construct_layer(lbann_comm* comm,
       return new slice_layer<layout, Dev>(comm,
                                           params.slice_axis(),
                                           slice_points);
```
bvanessen: The scoping on this endif statement seems wrong. But I am look at it on an iPad at the moment.

JaeseungYeom:
There is return nullptr at the end of the function. Also, there is a exception generation just before that.
I compiled this with and without the --with-conduit option but I will double check.
The matching brace at line 214 closes at line 246.

src/proto/factories/layer_factory.cpp
```diff
@@ -95,6 +96,11 @@ Layer* construct_layer(lbann_comm* comm,
        if (params.get_scalar_dimension_from_reader()) {
          num_neurons += scalar_dim;
        }
+    #else
+      err << "get_*_dimension_from_reader() not suppoeted";
```
davidHysom: suppoeted -> supported

PR #629, opened 2018-09-20
modifications to load jag samples from file when fetch_data is called…

davidHysom:
…, instead of loading all samples during setup (which blows out memory for 1M samples)


davidHysom:
overtaken by #647

Issue #632, opened 2018-09-25
Remove device synchronization in computing objective functions


naoyam:
This is required to overlap the input layer with forward propagation too. It is currently only overlapped with back propagation.

naoyam:
@timmoon10 said he'll work on this.

timmoon10:
The fundamental problem is that the evaluation layer has GPU data that needs to be allreduced and eventually end up on CPU so that it can be reported to the user. We currently implement this by performing a blocking GPU-CPU memory copy and then launching an asynchronous allreduce. An alternative approach is to perform a blocking allreduce on GPU and then copy asynchronously to CPU. This is probably less efficient, but it does allow better overlapping between CPU and GPU. A third option is to keep objective function and metric statistics on GPU, but this will require very deep surgery.

timmoon10:
After a failed attempt to shim in a solution, I think it would be better to close #193 before making another try.

ndryden:
@timmoon10 When you have a chance after the current round of cleanups, can you take a look at this again?

PR #633, opened 2018-09-25
GPU evaluation layer no longer blocks host.

timmoon10:
The trade-off is that it now performs a blocking GPU allreduce. Gradient checking for model_zoo/tests/model_mnist_conv_graph.prototext with 2 Pascal nodes comes back clean. This is an experimental solution to #632, pending performance benchmark results.

### DIFF
--- a/src/layers/transform/evaluation.cpp
+++ b/src/layers/transform/evaluation.cpp
@@ -37,7 +37,8 @@ namespace {
 /** CPU implementation of evaluation layer forward prop. */
 void fp_cpu(lbann_comm& comm,
             const AbsDistMat& input,
-            DataType& value) {
+            DataType& value,
+            Al::request& req) {
   const auto& local_input = input.LockedMatrix();
   const auto& local_height = local_input.Height();
   const auto& local_width = local_input.Width();
@@ -50,7 +51,7 @@ void fp_cpu(lbann_comm& comm,
     }
   }
   value = value / mini_batch_size;
-  comm.allreduce(&value, 1, input.DistComm());
+  comm.nb_allreduce(&value, 1, input.DistComm(), req);
 }
 
 #ifdef LBANN_HAS_GPU
@@ -135,11 +136,13 @@ void fp_gpu(lbann_comm& comm,
 } // namespace
 
 EvalType abstract_evaluation_layer::get_value(bool scaled) {
+  switch (get_device_allocation()) {
+  case El::Device::CPU: get_comm()->wait(m_allreduce_req); break;
 #ifdef LBANN_HAS_GPU
-  if (get_device_allocation() == El::Device::GPU) {
-    m_copy_event.synchronize();
-  }
+  case El::Device::GPU: m_copy_event.synchronize(); break;
 #endif // LBANN_HAS_GPU
+  default: LBANN_ERROR("invalid device");
+  }
   if (scaled) { return m_scale * m_value(0, 0); }
   else        { return m_value(0, 0); }
 }
@@ -160,7 +163,8 @@ void abstract_evaluation_layer::setup_data() {
 void abstract_evaluation_layer::fp_compute() {
   switch (get_device_allocation()) {
   case El::Device::CPU:
-    fp_cpu(*get_comm(), get_prev_activations(), m_value(0, 0));
+    fp_cpu(*get_comm(), get_prev_activations(), m_value(0, 0),
+           m_allreduce_req);
     break;
 #ifdef LBANN_HAS_GPU
   case El::Device::GPU:
--- a/src/layers/transform/evaluation.cpp
+++ b/src/layers/transform/evaluation.cpp
@@ -37,13 +37,12 @@ namespace {
 /** CPU implementation of evaluation layer forward prop. */
 void fp_cpu(lbann_comm& comm,
             const AbsDistMat& input,
-            DataType& value,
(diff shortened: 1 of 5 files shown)


src/layers/transform/evaluation.cpp
```diff
@@ -37,30 +37,31 @@ namespace {
...
+  value = 0;
 #pragma omp parallel for reduction(+:value) collapse(2)
   for (El::Int col = 0; col < local_width; ++col) {
     for (El::Int row = 0; row < local_height; ++row) {
       value += local_input(row, col);
     }
   }
   value = value / mini_batch_size;
-  comm.nb_allreduce(&value, 1, input.DistComm(), req);
+  comm.allreduce(&value, 1, input.DistComm());
```
bvanessen: Why does this become blocking for the CPU path?  Shouldn’t it remain independent of the GPU path?

timmoon10: This was mostly for simplicity. I can restore the non-blocking CPU implementation if need be.

src/layers/transform/evaluation.cpp
```diff
@@ -70,85 +72,104 @@ void fp_gpu(lbann_comm& comm,
...
-                  ones_d.LockedBuffer(), 1,
-                  sum_d.Buffer(col+1, 0));
+    GPUMat col_sums_d;
+#ifdef HYDROGEN_HAVE_CUB
+    col_sums_d.SetMemoryMode(1);  // Use CUB GPU memory pool
+#endif // HYDROGEN_HAVE_CUB
+    col_sums_d.Resize(local_width, 1);
+    ones_d.Resize(local_height, 1);
+    El::Fill(ones_d, one);
+    El::Gemv(El::TRANSPOSE, one, local_input, ones_d, zero, col_sums_d);
```
ndryden: Just to clarify, you converted the loop of dot products to a GEMV? This should perform better, but would be good to confirm.

src/utils/cuda.cu
```diff
@@ -0,0 +1,82 @@
...
+
+event_wrapper::event_wrapper(const event_wrapper& other)
+  : m_event(nullptr), m_stream(other.m_stream) {
+  CHECK_CUDA(cudaEventCreate(&m_event));
+  if (!other.query()) { record(m_stream); }
+}
+
+event_wrapper& event_wrapper::operator=(const event_wrapper& other) {
+  m_stream = other.m_stream;
+  if (!other.query()) { record(m_stream); }
```
ndryden: This query may make copying an unexpectedly expensive operation.

src/utils/cuda.cu
```diff
@@ -0,0 +1,82 @@
...
+////////////////////////////////////////////////////////////
+
+event_wrapper::event_wrapper() : m_event(nullptr), m_stream(0) {
+  CHECK_CUDA(cudaEventCreate(&m_event));
+}
+
+event_wrapper::event_wrapper(const event_wrapper& other)
+  : m_event(nullptr), m_stream(other.m_stream) {
+  CHECK_CUDA(cudaEventCreate(&m_event));
+  if (!other.query()) { record(m_stream); }
```
ndryden: This query may make copying an unexpectedly expensive operation.

src/utils/cuda.cu
```diff
@@ -0,0 +1,82 @@
...
+
+namespace lbann {
+namespace cuda {
+
+////////////////////////////////////////////////////////////
+// CUDA event wrapper
+////////////////////////////////////////////////////////////
+
+event_wrapper::event_wrapper() : m_event(nullptr), m_stream(0) {
+  CHECK_CUDA(cudaEventCreate(&m_event));
```
ndryden: For better performance, this should be created without timing, e.g. `cudaEventCreateWithFlags(&m_event, cudaEventDisableTiming)`.

src/utils/cuda.cu
```diff
@@ -0,0 +1,82 @@
...
+}
+
+event_wrapper::event_wrapper(const event_wrapper& other)
+  : m_event(nullptr), m_stream(other.m_stream) {
+  CHECK_CUDA(cudaEventCreate(&m_event));
```
ndryden: Ditto.

src/utils/cuda.cu
```diff
@@ -0,0 +1,82 @@
...
+  m_stream = other.m_stream;
+  if (!other.query()) { record(m_stream); }
+  return *this;
+}
+  
+event_wrapper::~event_wrapper() {
+  CHECK_CUDA(cudaEventDestroy(m_event));
+}
+
+void event_wrapper::record(cudaStream_t stream = 0) {
```
ndryden: Are we sure that we want to risk something going on the default stream?

src/layers/transform/evaluation.cpp
timmoon10: That's right. From a latency perspective, it's much better to launch one GEMV kernel than many DOT kernels. I suspect the compute time will be better two since it only needs to read in `ones_d` once.

src/utils/cuda.cu
timmoon10: The semantics of copying `event_wrappers` might still need work. My reasoning is that the copied version should at least be subsequent to the original.

src/utils/cuda.cu
timmoon10: Good point.

PR #634, opened 2018-09-26
GPU L2 weight regularization no longer blocks host

timmoon10:
The trade-off is that it now performs a blocking GPU allreduce. Gradient checking for model_zoo/tests/model_mnist_conv_graph.prototext (with L2 regularization added) with 2 Pascal nodes comes back clean. This depends on #633 and is an experimental solution to #632, pending performance benchmark results.


naoyam:
When you say the blocking GPU allreduce is trade-off, does that mean there is a performance penalty for GPU?

bvanessen:
I think that the tradeoff is that it used to be a blocking memcopy and now it is an allreduce.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

src/utils/cuda.cu
```diff
@@ -0,0 +1,83 @@
...
+
+namespace lbann {
+namespace cuda {
+
+////////////////////////////////////////////////////////////
+// CUDA event wrapper
+////////////////////////////////////////////////////////////
+
+event_wrapper::event_wrapper() : m_event(nullptr), m_stream(0) {
+  CHECK_CUDA(cudaEventCreate(&m_event));
```
naoyam: If this event is not used for timing measurement, using option cudaEventDisableTiming would be better as it is supposed to incur less overhead. See https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html#group__CUDART__EVENT_1g7b317e07ff385d85aa656204b971a042.

naoyam:
OK, that makes sense. Thanks.

> I think that the tradeoff is that it used to be a blocking memcopy and now it is an allreduce. Brian C. Van Essen vanessen1@llnl.gov (w) 925-422-9300 (c) 925-290-5470
> […](#)
> On Sep 25, 2018, at 9:36 PM, Naoya Maruyama ***@***.***> wrote: When you say the blocking GPU allreduce is trade-off, does that mean there is a performance penalty for GPU? — You are receiving this because your review was requested. Reply to this email directly, view it on GitHub, or mute the thread.

timmoon10:
I've run some experiments on 4 Pascal nodes (Resnet-50; ILSCVR2012) and I got 2418.43 sec for an epoch on develop (0d946b3) and 2418.42 sec on this PR (df0213a). We know that backprop communication completely covers up any CPU work, so this is not a surprising result.

PR #635, opened 2018-09-27
modify build script to optionally use ninja

benson31:
Allows the use of the ninja generator in LBANN through the build_lbann_lc.sh script. The primary advantage for me is that everything is contained in a single build.ninja file, so debugging build/compilation/linkage issues is easier. Also, it's faster than make on a per-project basis.


benson31:
@forsyth2 This could be helpful for compilation tests because of the way Ninja serializes output. It's still a little weird for the superbuild, but it's far less jumbled than the make build.

PR #636, opened 2018-09-27
Allow disabling OpenMP parallelism for data fetch loop

JaeseungYeom:
Allow to disable OpenMP parallelism for data fetching loop.
In some cases, the underlying data I/O library may not be thread safe, and this will allow users to avoid using OpenMP.


bvanessen:
The jag store does not have a problem with a multi-threaded call into conduit or hdf5.  We don’t want to globally set the number of threads to 1 because that can affect other parts of the code.  To make hdf5 thread safe you have to enable certain compile flags.

I will review this later.

Brian C. Van Essen
(w) 925-422-9300
(c) 925-290-5470

On Sep 27, 2018, at 4:24 PM, davidHysom <notifications@github.com<mailto:notifications@github.com>> wrote:


@davidHysom approved this pull request.

Seems OK to me, but not sure if it solve all problems with conduit/hdf5 interface. If running with a single processor. Concerns I have:

  1.  with multiple CPUs, different CPUs may open the same conduit file via the hdf5 interface, and read different samples. Is this "safe?" (this may be a question Cyrus can best answer)
  2.  how can we tell, in practice, if conduit/hdf5 interface is thread-safe and/or multiple-open safe? How can we write a unit-test (or something similar) to test?
  3.  we could also simply set the environment variable: export "OMP_NUM_THREADS=1" and this would accomplish the same behavior in fectch_data. Of course, it would also serialize openMP in other portions of the code.

PR #646, opened 2018-10-06
Save image callback can accommodate any layer.

### LINKED ISSUE #285: Save image at any given layer
Generalize save image callback to save image at any given (arbitrary) layer, this extension will not only be useful for auto encoder but also GAN and other future needs.

timmoon10:
The save image callback can now interface with any layer, not just input layers. It no longer interacts with a data reader, but calls OpenCV directly. I've (somewhat gratuitously) taken the liberty of reworking the prototext interface and updating all models in the model zoo, but this can be reverted if need be.

This PR closes #285.

Sample outputs:

| MNIST | ImageNet |
|--------|------------|
| test-data | imagenet-epoch1-images |

### DIFF
--- a/include/lbann/callbacks/callback_save_images.hpp
+++ b/include/lbann/callbacks/callback_save_images.hpp
@@ -22,36 +22,32 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
 // implied. See the License for the specific language governing
 // permissions and limitations under the license.
-//
-// lbann_callback_save_images .hpp .cpp - Callbacks to save images, currently used in autoencoder
 ////////////////////////////////////////////////////////////////////////////////
 
 #ifndef LBANN_CALLBACKS_CALLBACK_SAVE_IMAGES_HPP_INCLUDED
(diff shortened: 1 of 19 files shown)


timmoon10:
We do throw an error if the number of channels is invalid: https://github.com/LLNL/lbann/blob/9b943538663fe6b828983ad0c35bbcfd31358f1c/src/callbacks/callback_save_images.cpp#L62

JaeseungYeom:
I think your original "255 / (upper - lower)" was a better choice.

timmoon10:
The problem is that only the top entry will have an output of 255. If there are entries that are 1e-7 away from the top entry, scaling by "256 / (upper - lower)" will give all of them a value of 255.

JaeseungYeom:
I said better only because I believe it is typically how it is done, not because it is numerically superior. There will be no perfect solution for discretization.  For example, now a pixel value that is less by (1 - 1e-7) will be represented as the max. As a result, an algorithm to locate the brightest pixel may produce very different results after transformation. For images temporarily represented in floating points, the issue you are concerned about would be rare. However, for data that are presented as images, I see your point.

PR #648, opened 2018-10-08
add the DAG version of the siamese model prototext and

JaeseungYeom:
that of the prototext for fine-tuning

The differences from the originial prototext are
- batch normalization weights are frozen
- batch normalization weights are transferred for fine-tuning

### DIFF
--- a/model_zoo/models/siamese/finetune-cub/model_cub_batchnorm_transferred_and_frozen.prototext
+++ b/model_zoo/models/siamese/finetune-cub/model_cub_batchnorm_transferred_and_frozen.prototext
@@ -1,0 +1,944 @@
+model {
+  name: "directed_acyclic_graph_model"
+  data_layout: "data_parallel"
+  mini_batch_size: 64
+  block_size: 256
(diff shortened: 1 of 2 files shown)


model_zoo/models/siamese/finetune-cub/model_cub_batchnorm_transferred_and_frozen.prototext
```diff
@@ -0,0 +1,944 @@
...
+
+  weights {
+    name: "bn_conv1_bias"
+    constant_initializer {
+      value: 0.0
+    }
+  }
+
+
+  weights {
```
timmoon10: I think it would be safer to initialize running statistics weights with no optimizer:
```
weights {
  name: "something"
  some_initializer { something }
  optimizer {}
}
```
It shouldn't have an effect since we don't provide gradient updates to the running statistics, but it makes our usage clearer.

JaeseungYeom:
addressed Tim's suggestion.

PR #652, opened 2018-10-10
Hiding model script outputs.

timmoon10:
This will reduce repo bloat and make it easier to update these models with new features.


bvanessen:
I am not sure what this is trying to do.  I see that it deletes a bunch of prototext files and adds them to the .gitignore file. Is this to avoid having the versions of the files that are written out by LBANN added to the repo?

timmoon10:
That's right. This is mostly for convenience so that I can change the Python scripts without having to make corresponding updates to the prototext files.

timmoon10:
It turns out that @samadejacobs manually edits the model prototexts after they are generated by a Python script. Best not to mess with this for now.

PR #653, opened 2018-10-11
Always check CUDNN return values

naoyam:
Checking return values is not a performance concern, so it should always
be done. Previously, CHECK_CUDNN only checks errors with Debug build,
but that also introduces fairly heavy-weight device synchronizations
before and after error checking. FORCE_CHECK_CUDNN always checks return
values irrespective of the build mode, but due to the synchronizations
it is not only too costly but also can cause deadlocks.

This PR changes the behavior such that CUDNN return values are always
checked in any build mode. It still keeps the synchronizations for the
Debug build mode, but in my opinion we should also remove them as they
make the macro unusable.

FORCE_CHECK_CUDNN is not necessary anymore as CHECK_CUDNN always checks
errors. It is still there in order to make this PR minimal.

If this looks good, CUDA_CHECK and FORCE_CHECK_CUDA should be similarly
changed too.

### DIFF
--- a/include/lbann/utils/cuda.hpp
+++ b/include/lbann/utils/cuda.hpp
@@ -57,6 +57,19 @@
       LBANN_ERROR(err_CUDA_SYNC.str());                         \
     }                                                           \
   } while (0)
+#define LBANN_CUDA_CHECK_LAST_ERROR(async)                              \
+  do {                                                                  \
+    cudaError_t status = cudaGetLastError();                            \
+    if (status != cudaSuccess) {                                        \
+      cudaDeviceReset();                                                \
+      std::stringstream err_CUDA_CHECK_LAST_ERROR;                      \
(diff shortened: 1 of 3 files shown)


ndryden:
It can cause deadlocks because it can prevent the host from making progress when the GPU is waiting on the host.

I also think the synchronization is too heavy-weight for typical debugging. Supporting it as a separate option might be useful, but it perturbs the application enough (i.e. it could hide some bugs) and adds enough overhead that I don't think it should be standard.

naoyam:
I removed `FORCE_CHECK_CUDNN` in c60bbf6.

I also removed device sync in 3598bb2. Instead, in debug mode, it checks any error floating around before calling a CUDNN function. This should mostly addresses Tim's concern without having deadlocks.

PR #654, opened 2018-10-11
Fix dropout when input local matrix is empty

naoyam:
An empty matrix was not correctly detected. This caused CUDNN returned a random value at `cudnnDropoutGetReserveSpaceSize`, which was then used as the parameter for `Resize` of Hydrogen, sometimes resizing a matrix to an extremely large size. This is likely what happened with Sam's experiments.


samadejacobs:
Now we know where the memories have gone!

PR #655, opened 2018-10-11
Entry-wise loss layers

timmoon10:
This PR implements several layers for entry-wise loss functions:
- Binary cross entropy
- Sigmoid binary cross entropy (equivalent to TensorFlow's sigmoid_cross_entropy_with_logits and PyTorch's BCEWithLogitsLoss).
- Boolean accuracy
- Boolean false negative rate
- Boolean false positive rate

Much of the infrastructure for these layers is copy-pasted from the binary math layers.

The Boolean loss functions are thresholded so that values below 0.5 are considered false and values greater than or equal to 0.5 are considered true. For consistency, I've modified our logical math layers (not, and, or, xor) to use a similar thresholding scheme.

Finally, I've removed the bce_with_logits_layer, which required the label to be specified in the prototext. I think I've updated the model prototext files correctly, but @samadejacobs should double check.


src/layers/math/binary.cpp
```diff
@@ -356,8 +356,8 @@ struct greater_equal_op {
 struct and_op {
   inline DataType operator()(const DataType& x1,
                              const DataType& x2) const {
-    const bool b1 = x1 != zero && !std::isnan(x1);
-    const bool b2 = x2 != zero && !std::isnan(x2);
+    const auto& b1 = x1 >= DataType(0.5);
+    const auto& b2 = x2 >= DataType(0.5);
```
bvanessen: I think that it makes more sense to keep these with the standard definition of non-zero is true and zero is false.

PR #656, opened 2018-10-12
Use H Al communicators

ndryden:
Use Hydrogen's Aluminum communicators instead of LBANN creating them separately.

(I have not tested this. All the machines are either sad or busy.)


ndryden:
Eventually yes, but for now it's easier to do algorithm/backend selection in LBANN. Eventually that should get handled by Aluminum, and then we can just call Hydrogen's allreduce.

PR #657, opened 2018-10-12
Use MPI-CUDA backend when NCCL unavailable

naoyam:


naoyam:
A bit convoluted, but should not affect anyone unless disabling NCCL.

PR #662, opened 2018-10-18
fixes to match GCC-8 updates in Hydrogen

benson31:
This depends on https://github.com/LLNL/Elemental/pull/51


benson31:
Here we go. :crossed_fingers:

PR #666, opened 2018-10-22
Categorical accuracy layer

timmoon10:
This PR implements an optimized layer for top-1 categorical accuracy:
- We can now handle the case where the labels are not represented as one-hot vectors. The top value is interpreted as the label.
- We have somewhat higher communication requirements (4 gathers, for the local maximum entry {value, index} of the {prediction, label} data). We can easily overlap communication and computation once we implement suitable wrappers around Aluminum's non-blocking gather.
- The GPU implementation no longer calls Thrust's sort or reduce routines, avoiding the problems caused when they use the default CUDA stream. We still use Thrust's device vector class for convenience.

I get similar accuracies with AlexNet as before (2 nodes of Pascal and Catalyst; ImageNet-10; data-parallel and model-parallel layouts). The next step is to adapt this implementation for the top-k categorical accuracy layer.


ndryden:
I think this is doing gathers instead of a gather/allreduce. Is this computing something slightly different than before?

ndryden:
Okay. I'm unclear on what the advantages of a non-blocking gather would be in this situation, as there's no computation to overlap. Do you mean reformulating so that it's split into two phases?

Related, we can reduce the number of gathers by half by packing the indices and labels into one buffer and doing some type-casting (although we may have to worry about aliasing different types). Not sure whether this is currently worth-while.

timmoon10:
The implementation is as follows:
1. Find local max prediction.
2. Gather max predictions.
3. Find local max label.
4. Gather max labels.
5. Find global max prediction.
6. Find global max label.
7. Compute accuracy.

Steps 2 and 3 can be overlapped and steps 4 and 5 can be overlapped. If non-blocking gathers are already implemented in Aluminum, this approach will be easier and more effective.

ndryden:
The MPI-CUDA backend supports gathers, and Hydrogen should handle them (or can be easily updated to). However, considering that for many datasets the local lengths will be quite small, the overhead of non-blocking operations will probably outweigh the benefits. It might also be worthwhile to special-case the one-hot vectors, to reduce communication. That said, I don't see any obvious issues.

timmoon10:
It turns out that Thrust device vectors use the default stream. I'll need to do a little more work...

PR #667, opened 2018-10-25
Remove deprecated functionality in model zoo prototexts

timmoon10:
This PR removes deprecated functionality from some of the simpler model prototexts in the model zoo. In particular, it removes the target layers and reconstruction layers from test models and MNIST/CIFAR-10/ImageNet models. It also removes some ancient models, e.g. models using motifs, the recurrent model class, or greedy layer-wise autoencoders.

This is progress toward #193 and #632. We still need to update models involving CANDLE, JAG, Siamese stuff, and GANs. We'll also need to figure out what to do with the Keras interface.


samadejacobs:
I will wait to make sure that JAG and CANDLE pilot1 are properly supported before we consider this done. I will look at upgrading them.

timmoon10:
This can be merged without losing any functionality.

PR #668, opened 2018-10-25
Removing deprecated greedy layer-wise autoencoder model

timmoon10:
The greedy layer-wise autoencoder has not been supported for quite some while. It uses the old target layer paradigm (deprecated by #193), is explicitly disabled in our Bamboo tests, and is not supported by our prototext parser (since #242). Instead of pretending to support it, I think it would be best to properly remove it.


samadejacobs:
Yes,  in my opinion, we can remove it and rely on git history to resurrect if need be. I will approve but let Brian made final decision.

PR #671, opened 2018-10-26
Remove deprecated functionality in CANDLE Pilot1 models

timmoon10:
This is a followup to #667. It primarily removes target layers and reconstruction layers from the CANDLE Pilot1 models.


model_zoo/models/candle/pilot1/combo.prototext
```diff
@@ -482,10 +499,39 @@ model {
 
   #TARGET
   layer {
-    parents: "fc data"
-    name: "target"
-    target {}
+    parents: "fc response"
```
samadejacobs: Please add one line R^2 formula being used here for reference/documentation.

PR #675, opened 2018-10-26
Remove deprecated functionality in JAG models

timmoon10:
This is a followup to #667. It primarily removes target layers and reconstruction layers from the JAG models.

### DIFF
--- a/model_zoo/models/gan/jags/cycle_gan/cycgan_m1.prototext
+++ b/model_zoo/models/gan/jags/cycle_gan/cycgan_m1.prototext
@@ -1,5 +1,4 @@
 model {
-  name: "directed_acyclic_graph_model"
   objective_function {
     l2_weight_regularization {
       scale_factor: 0.0001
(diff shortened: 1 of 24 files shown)


model_zoo/models/alexnet/model_alexnet.prototext
```diff
@@ -1,5 +1,4 @@
 model {
-  name: "directed_acyclic_graph_model"
```
samadejacobs: Why deleting this? Is "directed_acyclic_graph_model" default model name in LBANN?

model_zoo/models/autoencoder_cifar10/model_autoencoder_cifar10.prototext
```diff
@@ -47,18 +43,30 @@ model {
   #######
   layer {
     name: "data"
-    children: "encode1 reconstruction"
+    children: "image label"
```
samadejacobs: Calling the second child of data layer "label" seems like a misnomer to me. It is not "label", rather a replica of input data, can we find a clearer/meaningful name?

samadejacobs:
Also, the title of this PR is misleading, more files were touched than JAG models. This is a minor thing, so I approve.

model_zoo/models/alexnet/model_alexnet.prototext
timmoon10: I've changed DAG models to be the default.

timmoon10:
This PR includes PR #667. The diff would be smaller if Github had a nice interface for staged pull requests.

PR #672, opened 2018-10-26
Mean absolute error layer and L1 norm layer

timmoon10:
The mean absolute error layer is necessary to replace some deprecated functionality used in the CANDLE Pilot2 models. It also makes progress toward #346. I've taken the liberty of making some cosmetic changes to the tensor dimension error checking in the loss layers.

PR #676, opened 2018-10-26
Remove deprecated functionality in CANDLE Pilot2 models

timmoon10:
This is a followup to #667 and #672. It primarily removes target layers and reconstruction layers from the CANDLE Pilot1 models.


bvanessen:
Can you please update this PR with the latest version of develop so that it is easier to see what changes are specific to this actual PR.

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext
```diff
@@ -146,16 +164,17 @@ model {
   # FULLY_CONNECTED decode1
   #################
   layer {
+    parents: "relu_fc1"
     name: "d_fc1"
     data_layout: "data_parallel"
     fully_connected {
       num_neurons: 11264
-      weight_initialization: "glorot_uniform"
```
bvanessen: Is this being removed because it is the default?

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_bead_autoencoder_pilot2.prototext
```diff
@@ -117,13 +126,13 @@ model {
       conv_dims: "12"
       conv_pads: "0"
       conv_strides: "12"
-      weight_initialization: "he_normal"
```
bvanessen: Given my previous question, why is the initialization removed.

model_zoo/models/molecular_autoencoder_candle_pilot2/model_molecular_autoencoder_pilot2.prototext
```diff
@@ -44,24 +38,36 @@ model {
...
   #############
   # FULLY CONNECTED ENCODE 1
   #############
   layer {
+    parents: "data"
     name: "encode1"
     data_layout: "data_parallel"
     fully_connected {
       num_neurons: 12
-      weight_initialization: "he_normal"
```
bvanessen: And here

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext
timmoon10: This field is deprecated and doesn't have any effect. It's from before we had independent weight objects.

bvanessen:
Okay, then should the individual weight matrices have the correct initialization.

Brian C. Van Essen
(w) 925-422-9300
(c) 925-290-5470

On Nov 29, 2018, at 4:01 PM, Tim Moon <notifications@github.com<mailto:notifications@github.com>> wrote:


@timmoon10 commented on this pull request.

In model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext<https://github.com/LLNL/lbann/pull/676#discussion_r237705295>:

>      name: "d_fc1"
     data_layout: "data_parallel"
     fully_connected {
       num_neurons: 11264
-      weight_initialization: "glorot_uniform"


This field is deprecated and doesn't have any effect. It's from before we had independent weight objects.

bvanessen:
If those fields are deprecated, how do we set the initalization for the weight matrices.

timmoon10:
Something like the following:
```
weights {
  name: "matrix_weights"
  glorot_uniform_initializer {}
}
weights {
  name: "bias_weights"
  constant_initializer {
    value: 0.0
  }
}
layer {
  parents: "parent"
  name: "fc"
  weights: "matrix_weights bias_weights"
  fully_connected {
    num_neurons: 1000
    has_bias: true
  }
}
```

It shouldn't matter too much in this case though, since the FC layers are followed by batch norm.

timmoon10:
I've updated the models to use new LBANN functionality and for clarity. It's very clear now that they are feedforward autoencoders with a form like: `encode1 -> encode2 -> decode2 -> decode1`. These models run fine with synthetic data on CPU. It looks like cuDNN doesn't support 1D convolutions, so I doubt they have ever worked on GPU.

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext
```diff
@@ -1,92 +1,85 @@
...
     data_layout: "data_parallel"
     convolution {
       num_dims: 1
       num_output_channels: 1024
       conv_dims: "240"
       conv_pads: "0"
       conv_strides: "240"
-      weight_initialization: "he_normal"
-      has_bias: true
+      has_bias: false
```
bvanessen: Is there a reason that you disabled the bias for several of the layers?

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext
```diff
@@ -165,46 +153,31 @@ model {
...
-  }
-
   layer {
-    name: "d_reshape1"
+    parents: "decode2_bn"
+    name: "decode2"
     data_layout: "data_parallel"
-    reshape {
-      num_dims: 2
-      dims: "1024 11"
```
bvanessen: What was the reason for removing the reshape layer?

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext
timmoon10: They're followed by batch norm, so the bias will just be removed.

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext
timmoon10: We set "decode2_fc" to have the same output dimensions as "encode1", which is already a 2D tensor.

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext
bvanessen: I see what you are saying and so from a technical point of view I agree.  For this refactoring, I would like to keep things as a literal translation without changing the behavior.

model_zoo/models/molecular_autoencoder_candle_pilot2/model_conv_molecular_autoencoder_pilot2.prototext
bvanessen: Okay.

Issue #678, opened 2018-10-29
Cull deprecated features


timmoon10:
### Deprecated features
- [x] Target layers
- [x] Reconstruction layers
- [x] All metrics except layer metric
- [x] All objective function terms except layer term and L2 regularization term
- [x] All models except DAG model

### Motivation
The target-layer-based objective function paradigm has been deprecated for months by the evaluation layer paradigm (#193). The target layer paradigm is somewhat simpler for models that fit into a rigid classification/regression/autoencoder framework, but is impractical if the objective function has any complexity. The evaluation layer paradigm is also more convenient from a code maintenance standpoint and can take advantage of GPU support in the layer class. We've supported both paradigms for backward compatibility, but this also makes it difficult to make substantial changes to our core objective function machinery. This difficulty is blocking progress on some performance optimizations (#632). I would say that this is a good point to bite the bullet and just remove our support for the deprecated features. Since most of our specialized model classes rely on the target layer paradigm, we should remove them as well.

### Progress
I've made a series of PRs to remove deprecated functionality from the model zoo (#667, #668, #669, #671, #675, #676). Once these PRs are merged, I believe the only offending models will be the Siamese models (pinging @JaeseungYeom). If anyone has models that aren't in the model zoo, they should check for deprecated functionality before we officially stop supporting them. For an example, I would recommend looking at how we updated LeNet (2376c0027bf75c36db553006edc0b1aec35f3fd2).

### Recommendations
- Change model type to `directed_acyclic_graph_model` (or remove the `name` field since DAG models are now default). If the model was previously a `sequential_model`, this will require specifying the parent/child relationships between layers.
- Specify two identity layers (or split layers or dummy layers) as children of the input layer. Input layers expect two child layers (one for sample data and one for "ground truth" data), so being explicit helps make sure you're getting the data you expect from the input layer. An example:
```
layer {
  name: "input"
  children: "image label"
  input {
    io_buffer: "partitioned"
  }
}
layer {
  parents: "input"
  name: "image"
  identity {}
}
layer {
  parents: "input"
  name: "label"
  identity {}
}
```
- Remove target and reconstruction layers and replace them with the appropriate loss layers. In most cases, they will require two parents like the target layer.
- Replace any terms in the objective function (except L2 weight regularization) with:
```
layer_term { layer: "name_of_corresponding_loss_layer" }
```
- Replace any metrics with:
```
metric {
  metric_term {
    name: "human-readable metric name"
[...]

Issue #479, opened 2018-06-25
Support intersection-over-union metric


ndryden:
Intersection-over-union (IOU) and mean IOU are standard metrics for semantic segmentation and similar tasks.

timmoon10:
This is tricky to implement as a metric since it breaks our assumption that metrics can be computed independently for each mini-batch sample. We should eventually generalize our metrics, but for now we can compute IOU from the confusion matrix (see #712):
```python
#!/bin/python                                                                   
import sys
import numpy as np

# Print usage message if no arguments are provided                              
if len(sys.argv) < 2:
    print('Usage: mean_iou.py [confusion matrix CSV]')
    sys.exit()

# Compute mean IOU                                                              
confusion = np.loadtxt(open(sys.argv[1], "r"), delimiter=",")
iou_list = []
for i in range(confusion.shape[0]):
    intersection = confusion[i,i]
    union = np.sum(confusion[i,:]) + np.sum(confusion[:,i]) - intersection
    iou_list.append(intersection / union)
print(np.mean(iou_list))
```

ndryden:
Having a solution for this is good. I don't think we should close this, but we can de-prioritize it for now since we have a reasonable workaround. Eventually, I think it would be good to have a proper, native solution for this, since IOU is common.

PR #679, opened 2018-10-29
Remove support for deprecated metrics

timmoon10:
This PR removes all metrics that use the deprecated target-layer-based paradigm. ~~In particular, the metric class is now a non-polymorphic class that reproduces the behavior of the "layer metric" class.~~

This PR is ready to go, but it will break any models that are using deprecated functionality (#678). If we merge this after the next tagged release, I think that's fair warning for everyone to update their prototext files.


timmoon10:
I think it would be better if we kept the notion of an abstract metric class. This could be helpful for metrics with tensor values, e.g. confusion matrices, or metrics that can't be computed independently over mini-batch samples, e.g. mean IOU (see #712 and #479). I've deleted the commit that removed the "layer metric" class.

bvanessen:
If we have removed all dependencies on the older metrics approach, please merge this.

PR #680, opened 2018-10-29
Removing deprecated Siamese model

timmoon10:
The Siamese model class uses the old target layer paradigm and only works if the user sets up their model according to a very rigid pattern (layers between a slice layer and concatenation layer are duplicated as "heads"). We have already transitioned to specifying models as DAGs, so I think we are ready to remove the deprecated functionality. This is progress toward #678.


bvanessen:
Tim, based on the conversation yesterday, please feel free to fix the conflicts and then merge this PR.

PR #686, opened 2018-11-01
Gather can accommodate arbitrary data types

timmoon10:
The LBANN communicator's gather function now reinterprets its data as `El::byte` (i.e.. `unsigned char`) before calling Hydrogen's MPI wrapper function. This allows it to easily handle custom structs, like those used in the top-k categorical accuracy layer. This approach can also be used for allgather, bcast, and scatter.

I've also made some changes for style conformity, mostly replacing camel case with underscores.

### DIFF
--- a/include/lbann/comm.hpp
+++ b/include/lbann/comm.hpp
@@ -277,7 +277,7 @@ class lbann_comm {
 
   template < typename T, El::Device D, bool S = is_instantiated_El_mpi_type<T>::value >
   void broadcast(const int root, T* data, const int count, El::mpi::Comm c,
-                 El::SyncInfo<D> const& syncInfo);
+                 El::SyncInfo<D> const& sync_info);
 
   /// World broadcast of a buffer.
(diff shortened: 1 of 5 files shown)


ndryden:
I'm not sure how I feel about this versus casting before calling gather in the first place. Do we want all instances of gather to be byte-based?

benson31:
I guess I'm not clear on why this is the right thing to do. I understand that you're just shuffling bytes around so as long as you don't have endian issues to deal with, this is fine correctness-wise, but it does limit you to gathers of 2GB and also assumes that `T` is packed (I know it is now, but will it be always?). If you're hitting instantiation issues in Hydrogen, just tell me what types you want enabled and I can do that. I can also instantiate different (storage-only) matrix types if that's what you need.

timmoon10:
Hm, those are good points. I'll change it so I only cast my weird anonymous-namespace-struct to `El::byte`.

PR #688, opened 2018-11-01
Superbuild output update

benson31:
This PR adds some features that are mostly for me to help people debug their build issues. The change you need to care about is that superbuild output is no longer dumped to stdout/stderr. Instead, it is piped to output.log and error.log in some directory and tagged with the package name and build step. The "make" (or "ninja") output now just tells you whether a given step succeeded or failed and where those logs are located.

Also, I have added a target called "gather-logs" (`ninja gather-logs` or `make gather-logs`) that will tar up the relevant output files. There is also a target called "gather-build" that will tar up all of the build files (e.g., `build.ninja` or the `*.make,link.txt` files). Those are helpful if you have some obscure link error or whatnot. Finally, there's a target called "gather-all" that does both of these things (outputs 2 tarballs; just depends on the other targest). You can then `give` these tarballs or email them to me.


CMakeLists.txt
```diff
@@ -401,33 +401,42 @@ get_directory_property( DirDefs COMPILE_DEFINITIONS )
 # Configuration summary
 ################################################################
 
-message("== Configuration Summary ==")
-message("  PROJECT_SOURCE_DIR:   ${PROJECT_SOURCE_DIR}")
-message("  PROJECT_BINARY_DIR:   ${PROJECT_BINARY_DIR}")
-message("  CMAKE_INSTALL_PREFIX: ${CMAKE_INSTALL_PREFIX}")
-message("  CMAKE_BUILD_TYPE:     ${CMAKE_BUILD_TYPE}")
+set(_str "== Configuration Summary ==\n")
```
bvanessen: Please put in a note in the CMake file so that people know why you are using strings, not the message function.  Also, could we use the messsage(STATUS instead of the string to keep it closer to a standard approach?

benson31:
Yes there's a similar config output for hydrogen in my latest PR there

PR #693, opened 2018-11-05
Removing broken option for FC layers

timmoon10:
PR #685 adds a check to the prototext parser so that a fully-connected layer's output size must be known at instantiation time. This breaks the prototext option `num_neurons_is_num_labels`, which sets the output size based on the data reader's label size. We have two options:
- Remove the check (https://github.com/LLNL/lbann/blob/cabacb1a81b18795df617cac9b513ab01d3e9d66/src/proto/factories/layer_factory.cpp#L154).
- Remove the `num_neurons_is_num_labels` option.

This PR does the second, mostly because it was the fastest fix. There are still a few Siamese models that keep the `num_neurons_is_num_labels` option, but those are slated to be removed in PR #680.

In the future, I think it would be worthwhile implementing functionality so that we can provide any layer's output dimensions as a hint for any subsequent layer's output dimensions (see issue #320). This is more general and would be helpful for other layers that require user-specified output dimensions (like constant layers). It would also allow us to avoid messing around with data readers.

### DIFF
--- a/model_zoo/models/alexnet/model_alexnet.prototext
+++ b/model_zoo/models/alexnet/model_alexnet.prototext
@@ -301,7 +301,7 @@ model {
     parents: "drop7"
     data_layout: "model_parallel"
     fully_connected {
-      num_neurons_is_num_labels: true
+      num_neurons: 1000
       has_bias: false
     }
   }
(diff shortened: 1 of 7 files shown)


timmoon10:
Ok, it sounds like option 1 is a better approach. This bug is non-trivial since it breaks our AlexNet model.

timmoon10:
As a followup, your concern will be addressed by the solution in Issue #320, once we get around to implementing it. It will be much more general, robust, and simple than our current ad-hoc solution (which only supports fully-connected layers and only can accommodate a specified set of data reader modes).

PR #695, opened 2018-11-05
FC layer doesn't need output dimensions at instantiation.

timmoon10:
This fixes a bug introduced in PR #685. See PR #693 for a discussion and an alternative solution.

### DIFF
--- a/src/proto/factories/layer_factory.cpp
+++ b/src/proto/factories/layer_factory.cpp
@@ -48,7 +48,7 @@ Layer* construct_layer(lbann_comm* comm,
       return new name##_layer<layout, Dev>(comm);       \
     }                                                   \
   } while (false)
-  
+
   // Input layers
   if (proto_layer.has_input()) {
     const auto& params = proto_layer.input();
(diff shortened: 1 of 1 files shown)


JaeseungYeom:
or add `if (params.num_neurons_is_num_labels())`
and set num_neurons there by calling `dr->get_num_labels()`

timmoon10:
I think this would be spaghetti code. I don't see why we should apply this check since we check again during layer setup.

JaeseungYeom:
What I am suggesting is to gather the num neuron setup capabilities to one place rather than making it duplicate or eliminating sanity check.

timmoon10:
We check the neuron tensor dimensions after setting up each layer: https://github.com/LLNL/lbann/blob/a1b23401e3c288360c0310a127b8f5dbea94b6e8/src/layers/layer.cpp#L742

JaeseungYeom:
I don't think simply checking tensor dimensions is sufficient here since there are multiple ways to set up num neurons. When a check fails, it is not clear whether the user has initiated a capability unsupported by the reader, misused the capability, or simply put a wrong dimension.

bvanessen:
I understand the discussion going on here.  Once this has been tested, I am going to merge it in for a needed code release so that we have a working AlexNet.  Once some of the work in the I/O subsystem is put to bed we are going to address the challenges of both the data reader and the prototext.

PR #697, opened 2018-11-08
Refactoring and optimizing softmax layer

timmoon10:
The primary goal of this PR is to improve the performance of the GPU model-parallel softmax layer (and log softmax layer), mostly in order to improve performance with VRAM. In particular, the old implementation launched separate CUDA kernels for each mini-batch sample and was dominated by kernel launch latency. The GPU data-parallel implementation still uses cuDNN and the CPU implementations are mostly unchanged, so we shouldn't expect changes in our model zoo models. I did some refactoring along the way:
- The softmax and log softmax layers now conform to our templated layer paradigm. Specialized implementations have been moved from the header files to the appropriate source files.
- Bamboo unit tests have been added to perform gradient checking on softmax and log softmax layers. These gradient checks come back clean and the output values match the values I computed manually in Excel.
- `logsoftmax_layer` has been renamed to `log_softmax_layer` to make the name consistent with TensorFlow and PyTorch (heads up to @LukeJaffe).


timmoon10:
The only way I've seen is to pass `-ftz=true` as an argument to nvcc (https://github.com/pytorch/pytorch/issues/10140, https://docs.nvidia.com/cuda/floating-point/index.html#compiler-flags). Wiping out denormals completely seems like overkill, especially since Nvidia GPUs support adds and multiplies with denormals in hardware and thus shouldn't pose too much of a performance hit (https://devblogs.nvidia.com/cuda-pro-tip-flush-denormals-confidence/).

PR #703, opened 2018-11-14
Hint layers for layer output dimensions

timmoon10:
This PR adds functionality so that layers can take their output dimensions from other layers (closing #320). Some comments:
- Some layers have inflexible output dimensions (e.g. unary layers must have the same output and input dimensions). I've gone ahead and modified all such layers to explicitly set their dimensions during setup. I've done some code cleanup in the process.
- This deprecates many of our prototext options that get output dimensions from data readers, e.g. the `num_neurons_is_num_labels` option for the fully-connected layer.
- There are some unhandled edge cases, like if you give a layer a hint layer that comes "afterwards". We should consider this as expert functionality.
- This could be generalized to get other properties from the hint layer. For example, it could be useful to setup a deconvolution layer by just providing the corresponding convolution layer as a hint (this doesn't fit exactly though, since we'll want the deconvolution layer to match the convolution layer's input dimensions).


include/lbann/layers/layer.hpp
```diff
@@ -503,6 +502,14 @@ class Layer {
    */
   std::vector<std::unique_ptr<AbsDistMat>> m_gradient_wrt_inputs;
 
+  /** Hint layer.
+
+
+   *  During setup, the output tensor dimensions are set to match the
+   *  first output tensor of the hint layer. Derived classes may
```
samadejacobs: Derived classes may........
Incomplete sentence

Issue #474, opened 2018-06-25
Support global average/max pooling


ndryden:
This is essentially applying either a max or a average to each feature map in a tensor, mapping input that is `H x W x C` to output that is `1 x 1 x C`.

If you have a known fixed input size to the layer, this can be done with just a pooling layer with an appropriately-sized filter. Global average pooling should work for arbitrarily-sized inputs to the layer (so that it works in FCNs).

timmoon10:
We have a reduction layer that computes a mean or max over the entire tensor. We could add functionality to choose which dimensions to reduce along.

PR #709, opened 2018-11-20
Channel-wise mean layer

timmoon10:
This PR implements channel-wise mean reduction for data-parallel data, on both CPU and GPU. It is a temporary solution toward #474. At instantiation, the layer flashes a warning that its functionality may disappear at any time. Gradient checks come back clean.

The long-term goal is to replace it with a generic reduce_mean layer that can reduce tensors along arbitrary axes. A reduce_sum layer is the transpose of a broadcast layer, so this is on our pathway.

### DIFF
--- a/src/layers/misc/channelwise_mean.cpp
+++ b/src/layers/misc/channelwise_mean.cpp
@@ -37,6 +37,8 @@ void channelwise_mean_layer<data_layout::DATA_PARALLEL, El::Device::CPU>
   auto& local_output = get_local_activations();
 
   // Dimensions
+  // Note: channel_size is the number of input entries per channel and
+  // local_width is the number of local mini-batch samples.
   const auto& input_dims = get_input_dims();
   const El::Int num_channels = input_dims[0];
   const El::Int channel_size = std::accumulate(input_dims.begin() + 1,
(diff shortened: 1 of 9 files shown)


src/layers/misc/channelwise_mean.cpp
```diff
@@ -0,0 +1,91 @@
...
+     ::fp_compute() {
+
+  // Local matrices
+  const auto& local_input = get_local_prev_activations();
+  auto& local_output = get_local_activations();
+
+  // Dimensions
+  const auto& input_dims = get_input_dims();
+  const El::Int num_channels = input_dims[0];
+  const El::Int channel_size = std::accumulate(input_dims.begin() + 1,
```
ndryden: Calling this "channel size" is a bit confusing, and could maybe do with a comment for clarification. What this really is is the size of a single channel of a single sample. (My first intuition was that it was the size of the channel dimension.)

src/layers/misc/channelwise_mean.cu
```diff
@@ -0,0 +1,179 @@
...
+      DataType private_sum = 0;
+      for (El::Int i = gidx; i < channel_size; i += nthreadsx) {
+        private_sum += input[i + channel*channel_size + col*input_ldim];
+      }
+
+      // Shared memory reduction to get sum for each block
+      /// @todo unroll loops
+      __shared__ DataType shared_sums[block_size];
+      shared_sums[tid] = private_sum;
+      for (El::Int stride = block_size / 2; stride > 0; stride /= 2) {
```
ndryden: This traversal pattern could do with a quick comment, it's not especially clear what's going on.

ndryden:
I agree a generic `reduce_mean` layer would be a nice thing to have. In some sense, what we want is the equivalent of `np.mean` (or `torch.mean`) where you can specify a axis/dimension to reduce over.

src/layers/misc/channelwise_mean.cu
naoyam: Cub has some primitives that hide dirty details of reductions.

Issue #49, opened 2017-03-14
confusion matrix


bvanessen:

bvanessen:
Add support for computing confusion matrices

timmoon10:
Closed with #712. In the future, we may want to generalize our metrics to handle tensor-valued metrics.

PR #712, opened 2018-11-21
Confusion matrix callback

timmoon10:
### Description
This PR adds a callback to compute confusion matrices and save them as CSV files (closing #49). It would be better to implement this as a metric, but that would require surgery in the base metric class to handle tensor-valued metrics. The CSV files can be parsed to compute the IOU metric (closing #479).

### Example usage
```
layer {
  parents: "input"
  name: "labels"
  identity {}
}
layer {
  parents: "softmax"
  name: "predictions"
  in_top_k { k: 1 } # Get one-hot prediction vectors
}
callback {
  confusion_matrix {
    prediction: "predictions"
    label: "labels"
    prefix: "experiment_dir/confusion-" # Files will look like "experiment_dir/confusion-train-epoch4.csv"
  }
}
```

### Example output
The following confusion matrix was obtained with LeNet on a validation set after one MNIST training epoch. Rows correspond to predictions and columns to labels.

screenshot from 2018-11-20 17-01-00

### DIFF
--- a/include/lbann/callbacks/CMakeLists.txt
+++ b/include/lbann/callbacks/CMakeLists.txt
@@ -6,6 +6,7 @@ set_full_path(THIS_DIR_HEADERS
   callback_check_reconstruction_error.hpp
   callback_checknan.hpp
   callback_checksmall.hpp
+  callback_confusion_matrix.hpp
   callback_debug.hpp
   callback_debug_io.hpp
   callback_dump_activations.hpp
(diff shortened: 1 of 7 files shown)


ndryden:
The [Wikipedia article on confusion matrices](https://en.wikipedia.org/wiki/Confusion_matrix) has a very nice set of useful things that can be derived from them. I think finally having them is very useful.

Unrelated: The color choice on that confusion matrix picture is terrible. The diagonal values should be green, since LeNet is not confusing things. The non-zero off-diagonals should be red. This is probably just Excel being silly though.

PR #713, opened 2018-11-21
Removing references to lscratchf partition

timmoon10:
The lscratchf partition is gone (at least on Pascal), so this PR replaces most references to the corresponding directories in lscratchh. LeNet with MNIST now works out-of-the-box.

There are some datasets that I couldn't find in lscratchh:
- NCI cancer (@samadejacobs)
- CANDLE Pilot2 (@samadejacobs)
- JAG (@davidHysom)


timmoon10:
Good catch. That one mostly duplicates model_zoo/data_readers/data_reader_imagenet.prototext, so I figure we don't lose much by removing it. Going further, I would lobby for removing model_zoo/models/imagenet/model_imagenet.prototext entirely since it's from a time before we had good support for convolutional networks.

bvanessen:
Good idea, please remove model_zoo/models/imagenet/model_imagenet.prototext as well.

PR #719, opened 2018-11-27
CosmoFlow Prototype

ndryden:
Prototype CosmoFlow model for 128x128x128 data cubes. The main difference is that this uses ReLUs instead of leaky ReLUs as activations, since we do not have GPU support for the latter yet.

This is based on the [CosmoFlow paper](https://arxiv.org/pdf/1808.04728.pdf) (and [this prior work](https://arxiv.org/pdf/1711.02033.pdf)), as well as [this](https://github.com/NERSC/CosmoFlow) Github repo.


bvanessen:
Can you add a run script so that we can easily hand it off to Brian G.

PR #722, opened 2018-11-28
flexible number of patches in Siamese model

JaeseungYeom:
Allow different number of patches in triplet_data_readers via specifying the prototext variable 'num_image_srcs' in the data reader prototext.


JaeseungYeom:
That is totally fine by me. At least you can see what is needed to be done from prototext, and facilities already exist.

PR #726, opened 2018-11-29
Jag utils

davidHysom:
Initial commit: directory for executables specific to the JAG data set. 

build_index.cpp constructs an index file with the following format:
1st line: num_valid_samples num_files
2nd line: base file directory
remaining lines: conduit_filename num_valid_samples num_failed_samples <indices of failed samples, if any>

Others to be added include:
  - driver for properly extracting random samples from 100M data set
  - drivers for re-packaging conduit files to (1) remove images not used by lbann (should reduce file size by 2/3); (2) put same number of images in each file (currently some files contain 10 samples, others 1000; this makes load balancing difficult); (3) remove invalid samples


davidHysom:
yes; anything in that directory will require building --with_conduit

JaeseungYeom:
Will the building work without the flag?

davidHysom:
oops; thanks. I it will break the build. I'll add guards and test build before merging.

PR #737, opened 2018-12-06
Refactoring callback to dump layer outputs

timmoon10:
### Summary
This PR refactors the output dumping callback to have more flexible and consistent behavior. This is primarily for debugging, but it can be used for a first pass of inference.

### The status quo
The "dump_activations" callback has two modes:
- If no layers are provided, it dumps outputs for all layers at every training mini-batch.
- If layers are provided, it dumps outputs for the selected layers at only the last testing mini-batch.

Output files are obtained by throwing data matrices into Elemental's `Write` function.

### Changes
The "dump_outputs" callback allows the user to choose layers and execution modes for dumping outputs. If none are provided, the default behavior is to dump outputs for all layers and/or all execution modes.  Output files are in CSV format, where each line corresponds to a mini-batch sample.

Some design decisions, all of which are up for debate:
- Renamed "dump_activations" to "dump_outputs", which is somewhat clearer and more consistent with the terminology we use elsewhere.
- Output files are in CSV format. This is more standard and easier to parse. If need be, it would be easy to add support for other delimited formats like TSV.
- Output files are formatted so that each line corresponds to a mini-batch sample. This makes it easier to read (trying to pick out a column in a text file is awkward). In addition, it's a more natural format for post-processing with Unix utilities like AWK. However, there may be some confusion since this format doesn't match the data matrices, but is the transpose.

### Inference
This functionality can be used for inference, although it'll be slow. For an example, suppose you have trained up the following model:
```
input -> image -> fc -> softmax -> cross entropy
  |                                     ↑
   ----> label --------------------------
```
Now load up the trained weights into the following model:
```
input -> image -> fc -> in_top_k (with k=1)
  |
   ----> label (unused)
```
Adding the following to the prototext will cause the model to emit predictions as one-hot vectors:
```
callback {
  dump_outputs {
    layers: "in_top_k"
    execution_modes: "test"
  }
}
```
If prediction indices are needed, the output files can be post-processed with AWK:
```awk
#!/bin/awk -f
BEGIN { FS = "," }
{
    idx = 0
    val = $1
    for (i = 1; i < NF; ++i) {
        if ($(i+1) > val) {
            idx = i
            val = $(i+1)
        }
    }
    print(idx)
}
```


include/lbann/callbacks/callback_dump_outputs.hpp
```diff
@@ -0,0 +1,100 @@
...
+ */
+class lbann_callback_dump_outputs : public lbann_callback {
+public:
+
+  /** Constructor.
+   *  @param layer_names    Names of layers with output dumps
+   *                        (default: dump outputs for all layers).
+   *  @param modes          Execution modes with output dumps
+   *                        (default: dump outputs for all modes).
+   *  @param batch_interval Frequency of output dumps (default: dump
```
samadejacobs: What is the choice for batch interval? I used the about to be deprecated dump activation a lot but mostly at every epoch end. Also, if this will support inference, one would expect the dump to be at forward_prop_end. I have no strong opinion but just want to know. One case I see against batch interval is that the user would have to compute number of iterations for an epoch if dumping at epoch boundary is desired (which mostly has been the way dump activations callback).

model_zoo/models/jag/gan/cyclic/model_template.prototext
```diff
@@ -73,34 +73,35 @@ model {
...
-  
+
   #Optional callbacks
   callback {
-    dump_activations {
-      basename: "/usr/workspace/wsa/jacobs32/github.saj.lbann/jag_imgs/cyclic_gan/"
-      interval: 100
-      layer_names: "image_data_dummy gen1fc4 param_data_id gen2fc4"
+    dump_outputs {
+      prefix: "/usr/workspace/wsa/jacobs32/github.saj.lbann/jag_imgs/cyclic_gan/"
```
samadejacobs: This is bad, would have to clean them up soon!

include/lbann/callbacks/callback_dump_outputs.hpp
timmoon10: The batch interval is used by the base callback class, so I don't think it's a good idea to hijack it to do something non-standard.

It's a good call to move the functionality to `on_forward_prop_end`/`on_evaluate_forward_prop_end`. I'll do that.

PR #742, opened 2018-12-07
Clamp layer

timmoon10:
This PR implements clamp layers, which constrain values between user-provided minimum and maximum values. This is equivalent to Keras' clip (https://keras.io/backend/#clip), TensorFlow's clip_by_value (https://www.tensorflow.org/api_docs/python/tf/clip_by_value), and PyTorch's clamp (https://pytorch.org/docs/stable/torch.html#torch.clamp). The Bamboo unit test passes with gradient checking and metric checking.

This is motivated by infinities and NaNs that sometimes pop up in VAEs. In Doersch's MNIST VAE model, the output from fully-connected "logsd" layer is fed into an exponential layer. Clamping "logsd" to a maximum value of 25 prevents the exponentiation from getting larger than O(1e10). Unfortunately, my experiments with ImageNet-10 data find that this doesn't affect the learning behavior very much. If you're producing these large values, it's already too late and your model has probably diverged. This seems like a problem of model design rather than numerics.


ndryden:
Related comment: People quite often clip gradients in a similar way to avoid exploding gradients.

PR #743, opened 2018-12-07
Nheads

andy-yoo:
Changes are made to replace current triplet Siamese model with more general m-head Siamese model.


JaeseungYeom:
Can you summarized what have been changed?

andy-yoo:
Oh, I do have prototext file for three- and four-head Siamese models.
I can add them to model_zoo directory.
I intend to keep the old triplet prototext file for reference (well, until we are sure the m-head Siamese works).

Yes, I have. The results from old triplet and new three-head Siamese produced results that are close enough.
I have not tried the save_image callback.

andy-yoo:
Basically, the changes are the same as you did, replaced hard-coded 3 and name ‘triplet’.
The placement of these changes are little bit different.
Also, changed some hard-coded test conditions to more general case.

The changes are pretty simple, and reviewing via git should highlight the changes clearly.
Let me know when you are done with the review.

Thanks for your help in advance.

JaeseungYeom:
Currently, in git, there are two files that shows up as if the entire files have been changed. Which test condition are you referring to? Can you put up a line number?

andy-yoo:
Are they data_reader_multihead_siamese.cpp|.hpp?

PR #745, opened 2018-12-07
Bugfix shared data readers

bvanessen:
Fixing how data readers are shared between models.


bvanessen:
My intent was to make the ae model not shareable.  Thus it would have its own independent data reader.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #747, opened 2018-12-08
Add conduit superbuild

benson31:
This PR adds HDF5 and conduit to our superbuild packages. The build appeared to work with build_lbann_lc.sh on pascal.


scripts/build_lbann_lc.sh
```diff
@@ -807,6 +787,9 @@ CONFIGURE_COMMAND=$(cat << EOF
 -D LBANN_SB_BUILD_ALUMINUM=${WITH_ALUMINUM} \
 -D ALUMINUM_ENABLE_MPI_CUDA=${ALUMINUM_WITH_MPI_CUDA} \
 -D ALUMINUM_ENABLE_NCCL=${ALUMINUM_WITH_NCCL} \
+-D LBANN_SB_BUILD_CONDUIT=${WITH_CONDUIT} \
+-D LBANN_SB_BUILD_HDF5=${WITH_CONDUIT} \
```
JaeseungYeom: We might use HDF5 in some other context at the same time. e.g., check-pointing #462 , which is unrelated to the conduit use.

benson31: If none of this functionality exists at the head of develop when this gets approved/merged, I would argue that such a need does not exist at this time and the appropriate changes can be incorporated into the build system as part of those pending or future PRs.

benson31:
@davidHysom no problem. I'm in no particular hurry to merge this -- it was to support y'all's debugging efforts. I tested that it builds, but I did not test that the resulting build actually works on anything that actually calls into conduit (I always test fake alexnet to just check that the binary is sensible, and that was fine, but also doesn't call into conduit). When I asked for a reasonable example to run, I was basically told that nothing existed that would work "out-of-the-box". If that's changed, I'm more than happy to test stuff out if you send me a command line that will work.

JaeseungYeom:
I am getting error as below with `build_lbann_lc.sh --with-conduit` on pascal and quartz.

```
make[5]: *** No rule to make target `gen_SRCS.stamp1', needed by `src/CMakeFiles/gen_hdf5-static'.  Stop.
make[4]: *** [src/CMakeFiles/gen_hdf5-static.dir/all] Error 2
```

benson31:
This is the result of an obvious and glaring mistake in the HDF5 CMake. They have a target depend on a file they create. This is only a problem if you don't tell the generator that the build system actually creates that file because then it wouldn't be able to construct a proper DAG. And guess what. Yep, they don't tell the generator that they created the file. So, kaboom. I will patch around this until it gets fixed upstream.

benson31:
HDF5 has resolved their build issue at the head of `develop` so things should work now.

davidHysom:
> @benson31  >I was basically told that nothing existed that would work "out-of-the-box". If that's changed, I'm more than happy to test stuff out if you send me a command line that will work.
Piffle. This should work:
$ find /p/lscratchh/brainusr/datasets/10MJAG/1M_A -name '*bundle' > filelist
$ srun --ntasks ... <path_to_build>/model_zoo/jag_utils/build_index --filelist=filelist --base_dir=/p/lscratchh/brainusr/datasets/10MJAG/1M_A --output_fn=index.txt

On lassen: $ /p/gpfs1/brainusr/datasets/10MJAG/1M_A -name '*bundle' > filelist

I'm hoping this can be merged soon, since I'm getting back to working on data_store_jag, and this requires conduit with MPI.

davidHysom:
> I'm more than happy to test stuff out if you send me a command line that will work
@benson31 I've added a driver: model_zoo/jag_utils/test_mpi
This should be run with two processors and no command line arguments.
However, main() is #if zeroed out -- since, otherwise, it will currently break
a build that's compiled --with-conduit but where the conduit build lacks,
or has improper, mpi support.

benson31:
@davidHysom Thanks! Is there a way to statically determine whether CONDUIT was built with the required support? If so, I can add something to the CMake to detect incompatibilities at configure time.

davidHysom:
perhaps: check to see if $CONDUIT_DIR/libconduit_relay_mpi.so exists

benson31:
@davidHysom I tried rebuilding this today and it took some tweaks to get it up and running, but I believe it's there now. If you just give `build_lbann_lc.sh --with-conduit` a whirl, it should successfully build the whole stack including HDF5 and CONDUIT (still with gcc/4.9.3, but blame the maintainer of `build_lbann_lc.sh` for that). If you are successful, I'd like to get this merged by this week's meeting. @bvanessen

davidHysom:
OK. But I always forget: how do I clone your branch?

benson31:
`git clone https://github.com/benson31/lbann -b add-counduit-superbuild` is the fastest way.

davidHysom:
fatal: Remote branch add-counduit-superbuild not found in upstream origin

davidHysom:
thanks; worked for me on pascal and lassen (ok, pascal gave me a cuda error at runtime, so I build without cuda/aluminum/nccl)


My test was:


  1. did the build go through?

  2. did model_zoo/jag_utils/test_mpi work?


1st time through, found an error in test_mpi, so pushed a quick PR fix.


Note: test_mpi *does not* use conduit's hdf5 interface. So, I need to modify test_mpi to also test that. Will do so later tonight/early tomorrow

PR #748, opened 2018-12-08
Updated the callback_save_model to save the model's weights at the end

bvanessen:
of training and testing (useful for CycleGAN example).  Exporting the
model's prototext is temporarily disabled in this callback because the
function does not work.


include/lbann/models/model.hpp
```diff
@@ -241,6 +241,10 @@ class model {
   /** Save the model's weight to file */
   virtual bool save_weights(persist& p);
 
+  /** Reload the model's weights from a file */
+  virtual bool reload_weights(const std::string latest,
```
timmoon10: - We should pick a more descriptive name than `latest` and document the functionality.
- The implication of `latest` is too restrictive. There are plenty of times when we might want to reload weights from many epochs ago, e.g. if training has diverged.

src/callbacks/callback_save_model.cpp
```diff
@@ -133,4 +135,42 @@ bool lbann_callback_save_model::save_model_weights(model *m) {
...
+  lbann_comm *comm = m->get_comm();
+  if(comm->am_model_master()) {
+    latest = get_shared_checkpoint_dirname(m, ckpt_dir, epochLast, stepLast);
+    std::cout << "Loading model weights from " << latest << std::endl;
+  }
+  comm->model_broadcast(0, &(latest[0]), sizeof(latest), El::SyncInfo<El::Device::CPU>{});
+
+  DIR *weight_dir;
+  struct dirent *weight_file;
+  if((weight_dir = opendir(latest.c_str())) == NULL)
```
timmoon10: This is too clever. It's more readable if you break it up into two lines. We should also prefer `nullptr` to `NULL`.

src/callbacks/callback_save_model.cpp
```diff
@@ -133,4 +135,42 @@ bool lbann_callback_save_model::save_model_weights(model *m) {
...
+  {
+    std::cout << "error opening " << latest << "\n";
+    return false;
+  }
+  // Populate weight list
+  while ((weight_file = readdir(weight_dir)) != NULL){
```
timmoon10: Too clever. Prefer `nullptr`.

include/lbann/models/model.hpp
bvanessen: Reasonable point.  Latest refers to the last checkpoint and I kept the nomenclature here so that you could easily reload from the last checkpoint of saved model state.  But we may want to make it more flexible.

src/callbacks/callback_save_model.cpp
bvanessen: Fixed the nullptr.  The rest will have to wait.

src/callbacks/callback_save_model.cpp
bvanessen: Fixed.

Issue #710, opened 2018-11-20
Reduction and broadcast layers


timmoon10:
It would be helpful to have layers that can reduce tensors along arbitrary dimensions:
- [ ] reduce_sum
- [ ] reduce_mean
- [ ] reduce_max
- [ ] reduce_min
- [ ] broadcast (transpose of reduce_sum)

Assuming these layers are performant, this would deprecate the functionality in #709.

PR #755, opened 2018-12-11
Tessellate layer

timmoon10:
This PR implements the tessellate layer, which is a generalization of the "tile" operation in Keras, Caffe, TensorFlow, and PyTorch. In particular, the tessellate layer's output dimensions don't have to be integer multiples of the input dimensions. This can perform tensor broadcasting (see Issue #710). A Bamboo unit test passes with gradient checking and metric checking.

This is currently only supported for tensors with 3 or fewer dimensions, e.g. image data. I tried to keep it simple, so I haven't put much effort into optimizing performance. If it becomes a bottleneck, we can improve performance by writing custom kernels for scalar broadcasting and for data-parallel data layouts.


timmoon10:
I think "tessellating a tensor" is as simple as can be accomplished in English. I've added a mathematical definition of the operation.

PR #759, opened 2018-12-11
Strict dimension checks for layer input tensors

timmoon10:
Currently, many of our layers require input tensors to have identical dimensions, e.g. binary math layers or some loss layers. This can get annoying since, for example, you can't directly compute the cross entropy between a 1000x1x1-tensor and a 1000-tensor. This PR relaxes the requirement to only require that input tensors have the same number of entries. I'm personally ambivalent about this change since you can now add a 2x3x4-tensor and a 6x2x2-tensor without any warning.

For reviewing, it will be helpful to know that the dimension checks for the categorical accuracy, cross entropy, mean absolute error, mean squared error, top-k categorical accuracy, binary math, and covariance layers are identical. The dimension checks for the Hadamard, sum, and weighted sum layers are also identical.

### DIFF
--- a/include/lbann/layers/loss/categorical_accuracy.hpp
+++ b/include/lbann/layers/loss/categorical_accuracy.hpp
@@ -55,25 +55,25 @@ public:
   void setup_dims() override {
     Layer::setup_dims();
     set_output_dims({1});
+
+    // Check that input dimensions match
     if (get_input_size(0) != get_input_size(1)) {
       const auto& parents = get_parent_layers();
-      const auto& dims0 = get_input_dims(0);
-      const auto& dims1 = get_input_dims(1);
       std::stringstream err;
       err << get_type() << " layer \"" << get_name() << "\" "
-          << "expects inputs with identical dimensions, but "
-          << "layer \"" << parents[0]->get_name() << "\" outputs a ";
-      for (size_t i = 0; i < dims0.size(); ++i) {
-        err << (i > 0 ? "x" : "") << dims0[i];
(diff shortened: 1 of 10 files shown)


timmoon10:
I'm wary of duplicating intricate dimension-checking over 10 layers (or more). Perhaps we should start thinking of proper tensor broadcasting semantics.

ndryden:
I the mean time, could we just write a function (or two) that does the appropriate checks? We already have the check/error code duplicated.

timmoon10:
Thinking more about it, do we even want this behavior? Following Numpy's broadcasting rules (https://docs.scipy.org/doc/numpy-1.15.0/user/basics.broadcasting.html), broadcasting a 1000x1x1-tensor with a 1000-tensor results in 1000x1x1000-tensors. We can add a special case for this particular situation, but what happens when we want to provide general tensor broadcasting?

ndryden:
So, that's a good point. I fixed the issue I ran into via a reshape layer. PyTorch, for example, is actually even more strict than we are, in that it requires explicitly reshaping the output of a convolutional layer before passing it to a fully-connected layer. On the other hand, its loss functions automatically handle multi-dimensional inputs appropriately. Perhaps making it clear that everything is really tensors, and you have to handle that is worth-while. Certainly, it should be reasonable to practitioners (and if any practitioner _isn't_ aware that it's all a tensor, ...)

timmoon10:
I've modified this PR to perform strict dimension checks. Merging it won't have any effect on behavior, but it does standardize the dimension checking code.

PR #764, opened 2018-12-12
Sample list for jag_conduit data reader

JaeseungYeom:
Use the list file that contains the valid samples identified in advance.
- Currently, positive list only
  - Negative list requires the scanning of data set, and will need the number of valid samples per file for easier partitioning.
  - Negative list is also not generally applicable while reduces the list file size significantly when applicable
- have not tested yet.
  - Only tested separately from lbann using regular MPI
  - Currently relies on some MPI operations that are not wrapped in lbann_comm
  - Need to use model specific ranks
- Subset selection to be implemented:
- Shuffling to be implemented:
  - currently relying on std::list, which needs to be changed to vector to allow shuffling, or at least copied to. This requires (for efficiency) reserving the size of the vector, and the number of total lines in the file at the top of the file.
- The expected sample list file format is that each line begins with a file name, followed by a keys of valid samples in that file. However, it is not required to have all the samples in a file in a single line. Thus, it is (theoretically) possible to list each sample individually per line.


davidHysom:
My understanding, after what Brian had to say -- and looking at all the templates -- is that the sample_list class is intended to be much more general than just for the jag_reader conduit class. If that's true I suggest breaking this into two PRs: 1) for the sample_list class; 2) for the incorporation of sample_list class into data_reader_jag_conduit.

I'd also like to see better documentation ... what, exactly, are the problems/solutions that sample_list class addresses?

That said, I'm approving your PR, because my above comments are only suggestions.

PR #765, opened 2018-12-12
Feature jag misc

bvanessen:
Update the model zoo files for the 5 model version and add a script for looking for duplicate outputs.


model_zoo/lbann_cycgan.cpp
```diff
@@ -136,14 +136,14 @@ int main(int argc, char *argv[]) {
       if (master) std::cerr << "\n STARTING train - G1 solver model at step " << super_step << " \n\n";
       model_2->train( super_step*pb_model_2.num_epochs(),pb_model_2.num_batches());
       // Evaluate model on test set
-      model_2->evaluate(execution_mode::testing,pb_model_2.num_batches());
+      //      model_2->evaluate(execution_mode::testing,pb_model_2.num_batches());
```
samadejacobs: This make assumption that we will not need to evaluate in between training which may be true for current large-scale inference that uses inference engine but does not consider "small scale/normal" training/inference where training and inference are done in the same run.

PR #770, opened 2018-12-13
Add factory template

benson31:
This is a general factory template that we should start using when we need factory patterns (creating layers, callbacks, etc.).

To limit its size, the scope of this PR will be simply to add this template and show that it works "in general". Incorporating its usage into "real code" will be a larger endeavour and should be deferred to a later PR.

A test case should be included once the team has established the use of a real unit test framework.


timmoon10:
Minor quibble. Can we change the documentation to use "@" for tags instead of "\"? I've revamped the documentation for layers with that choice (#772).

benson31:
Adding a test would involve integrating a unit test framework. I don't want to wait forever, so I'm going to defer that task to such a time that we have a unit test framework. I have a test for this class that can be incorporated as soon as that framework exists.

Issue #767, opened 2018-12-13
Background I/O for partitioned data - needs phase specific buffers


bvanessen:
Adding double buffering to the partitioned I/O buffer does not maintain a per execution phase buffer.  This works for normal training approaches but will have a poor interaction with LTFB.  The distributed I/O buffer has support for this and it needs to be moved over to the partitioned buffer prior to deleting the distributed I/O buffer.

PR #773, opened 2018-12-17
LTFB can handle multiple processes per model

### LINKED ISSUE #663: LTFB only works with one process per model
We are a bit sloppy and use the model rank and process rank interchangeably. The callback will probably hang in a SendRecv if the "model instances" have multiple MPI ranks.

When we get around to debugging LTFB, perhaps it would be worthwhile reimplementing it to handle heterogeneous network architectures. One way to do this would be use the checkpoint-restart functionality: each model would save a checkpoint (either to the file system or into Protobuf's wire format), send it to its partner, restart the received model, and then perform the tournament. This will depend on a good solution to #639, especially since we'll need to break out of the assumption that each "model instance" has an identical architecture.

timmoon10:
This is a bugfix to close #663. It also does some refactoring, cleaning, and documentation, in preparation of future development of LTFB. Example usage:
```
callback {
  ltfb {
    batch_interval: 100
    metric: "categorical accuracy"
    weights: "conv1_kernel conv1_bias"
    low_score_wins: false
  }
}
```


timmoon10:
While this is ready for review, I haven't been able to test it yet. I run into runtime errors, probably due to Issue #767.

<details> <summary> Error message </summary>

```
[3] Epoch : stats formated [tr/v/te] iter/epoch = [11/1/2]
            global MB = [1280/ 130/ 256] global last MB = [  70  / 130  / 244  ]
             local MB = [ 256/ 130/ 256]  local last MB = [  14+0/ 130+0/ 244+0]
LTFB (model "Model0", step 40): starting tournament...
LTFB (model "Model0", step 40): tournament partners - {2,0}, {3,4}, {1}
LTFB (model "Model0", step 40): evaluating local model...
***************************************************************
Number of samples does not match the number of responses
Stack trace:
   0: lbann::stack_trace::get()
   2: lbann::partitioned_io_buffer::fetch_to_local_matrix(lbann::generic_data_reader*, execution_mode)
   3: lbann::generic_input_layer::fetch_data_in_background(int, std::string)
   4: std::_Function_handler<void (), std::reference_wrapper<std::_Bind_simple<std::reference_wrapper<std::_Bind<std::_Mem_fn<void (lbann::generic_input_layer::*)(int, std::string)> (lbann::generic_input_layer*, int, char const*)> > ()> > >::_M_invoke(std::_Any_data const&)
   5: std::_Function_handler<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> (), std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<void>, std::__future_base::_Result_base::_Deleter>, void> >::_M_invoke(std::_Any_data const&)
   6: std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>&, bool&)
   7: pthread_once (demangling failed)
   8: lbann::type_erased_function::Function<std::packaged_task<void ()> >::call_held()
   9: lbann::thread_pool::do_thread_work_pinned_thread_(int, cpu_set_t)
  10: std::thread::_Impl<std::_Bind_simple<std::_Mem_fn<void (lbann::thread_pool::*)(int, cpu_set_t)> (lbann::thread_pool*, unsigned long, cpu_set_t)> >::_M_run()
  11: /usr/tce/packages/gcc/gcc-4.9.3/lib64/libstdc++.so.6(+0xb5dc0) [0x2aaacf54adc0] (could not find stack frame symbol)
  12: /lib64/libpthread.so.0(+0x7dd5) [0x2aaab022cdd5] (could not find stack frame symbol)
  13: clone (demangling failed)
```

</details>

src/proto/factories/callback_factory.cpp
```diff
@@ -97,14 +97,11 @@ lbann_callback* construct_callback(lbann_comm* comm,
...
-    auto&& w = parse_list<>(proto_cb.ltfb().weights_tosend());
-    std::unordered_set<std::string> metric_names(m.begin(), m.end());
-    std::unordered_set<std::string> weight_names(w.begin(), w.end());
-    return new lbann_callback_ltfb(proto_cb.ltfb().round_size(),
-                                   metric_names,
-                                   proto_cb.ltfb().increasing_metric_mode(),
-                                   weight_names,
+    const auto& params = proto_cb.ltfb();
+    return new lbann_callback_ltfb(params.batch_interval(),
+                                   params.metric(),
```
samadejacobs: We are not using multiple metrics now but how would this support space-delimited metric (e.g., categorical accuracy) and multiple metrics with the same name as metric specified one of the metrics for training/inference ?

timmoon10: - I've removed support for multiple metrics. At the end of the day, LTFB will choose winners based on a single number. If it's the average of multiple metrics, the user can compute the mean manually with a weighted sum layer. Similar tricks can be used for the min/max of multiple layers, although there will be some mathematical differences (average of maxes rather than max of averages).
- Since we only expect one metric, there's no problem if there are spaces in the metric name. Come to think of it, though, we may want to force metric names to have no spaces (like in layer names).
- I think we should require the metric name to be a unique identifier, like the layer name. If we allow metric names to be duplicated, we'll need to define a "metric ID" or something similar to be a unique identifier, so the practitioner won't save any effort.

bvanessen:
I am working on a fix.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

timmoon10:
When I build with #777 (and some extra hacks provided by @bvanessen), this runs perfectly fine with 5 trainers, 4 procs/trainer. I'll merge this now so it'll be ready to use once #767 is fixed.

PR #780, opened 2018-12-19
Updated the background I/O fetch function to take the execution mode

bvanessen:
at the time of dispatch rather than at the time of execution.  This
ensures that the correct data is placed in the buffer.


bvanessen:
This will work in conjunction with #777 and should have been in that PR.

PR #784, opened 2018-12-20
Callback to perturb Adam hyperparameters

timmoon10:
This reproduces the Adam hyperparameter exploration functionality in #615, but in a much less hacky way. I think a good approach for hyperparameter exploration is to create a bunch of callbacks that each handle an individual set of hyperparameters, e.g. one for dropout, another for SGD, and so on.

I'm under the impression that Adam hyperparameters can range over orders of magnitude, so the perturbations take place in log space. I am more precise in the documentation and I would appreciate comments if it's at all unclear. The user has the option to perturb Adam just during setup or periodically during training. For instance, doing it after each LTFB round is similar the exploration stage in population-based training.

### DIFF
--- a/include/lbann/callbacks/CMakeLists.txt
+++ b/include/lbann/callbacks/CMakeLists.txt
@@ -21,6 +21,7 @@ set_full_path(THIS_DIR_HEADERS
   callback_io.hpp
   callback_learning_rate.hpp
   callback_ltfb.hpp
+  callback_perturb_adam.hpp
   callback_print.hpp
   callback_save_images.hpp
   callback_save_model.hpp
--- a/include/lbann/callbacks/callback_perturb_adam.hpp
+++ b/include/lbann/callbacks/callback_perturb_adam.hpp
@@ -1,0 +1,126 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2016, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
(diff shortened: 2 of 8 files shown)


ndryden:
I'm curious why the perturbations are drawn from a normal distribution. This would seem to add extra mass near a fixed value (hardcoded to zero) in parameter space. I'd be interested in a justification for this (mostly for my own curiosity).

timmoon10:
If we were doing this once during setup, a uniform distribution (in log space) would probably be the way to go. I'm thinking about doing this after each LTFB round (like in population-based training) and Gaussian noise seems like a more natural perturbation than uniform noise (cue hand waving).

ndryden:
Well, random search for hyperparameter exploration uses a uniform distribution (see [here](http://www.jmlr.org/papers/volume13/bergstra12a/bergstra12a.pdf)). Though if the argument is that this is meant to be a perturbation from the prior round, perhaps a normal distribution could be more easily justified.

The other thing I noticed is that this is perturbing the epsilon, which is not really a hyperparameter, it's just a constant meant to avoid dividing by zero. I don't think it really needs to be tuned, and reducing the dimensionality through which you search should help get better results faster.

timmoon10:
Yea, the epsilon perturbation is there just for completeness. The default is to not perturb it.

samadejacobs:
I agree on discussion so far. We could just add a clear statement like "epsilon is a very small number to prevent any division by zero better left as default (not to perturb)". Though anyone using this should know what they are doing!

ndryden:
Never assume that future you knows what they’re doing. :)

PR #792, opened 2018-12-21
Update build system docs

benson31:
Please view the build instructions in Github and in the sphinx-generated documentation.

We need to get rid of `build_lbann_lc.sh`. It is causing too many problems and I wouldn't wish its maintenance on my worst enemies (and, indeed, I am continuing to actively unvolunteer myself from that responsibility).


timmoon10:
Maybe we should rename the script to something like `developer_build.sh`. That would scare off users and allow us to put horrifying hacks for whatever random features or experiments we're tinkering with.

davidHysom:
+10

bvanessen:
Okay, at this point we are not going to get rid of the build script.  I will discuss how to clean it up going forward.

docs/BuildingLBANN.md
```diff
@@ -0,0 +1,382 @@
...
++ `LBANN_WITH_TBINF` (Default: `ON`): Enable the Tensorboard interace.
+
++ `LBANN_WITH_VTUNE` (Default: `OFF`): Build with extra annotations for VTune.
+
++ `LBANN_DETERMINISTIC` (Default: `OFF`): Force as much of the code as possible
+  to be deterministic. This is not a guarantee as certain operations
+  in third-party libraries cannot be forced into a deterministic mode,
+  especially for CUDA-enabled builds.
+
++ `LBANN_SEQUENTIAL_INITIALIZATION` (Default: `OFF`): Force sequentially
```
ndryden: We should document this better, because we're not really referring to [sequential consistency](https://en.wikipedia.org/wiki/Consistency_model#Sequential_Consistency), but that initialization is the same regardless of how many processors participate. (I think I'm the one who actually called it "sequential consistency", but that doesn't mean we should not embellish upon my poor choices.)

docs/BuildingLBANN.md
```diff
@@ -0,0 +1,382 @@
...
+  consistent initialization of data structures.
+
++ `LBANN_WARNINGS_AS_ERRORS` (Default: `OFF`): Promote compiler
```
ndryden: If all developers build with this on, there should be no issues with users building with this on. Therefore it should always be on. 😉

benson31: My concern is overly enthusiastic users (ok, probably just me) using weird compilers at newer or older versions than we test on. E.g., clang 6+ complains about the way El/Hydrogen abuse templates in the MPI wrapper and GCC does not. Then there's all the weird stuff Intel emits. I don't want to be on the hook for appeasing all user-end compilers that dislike something in our style. But if you're volunteering... :)

docs/BuildingLBANN.md
```diff
@@ -0,0 +1,382 @@
...
+  communication library optimized for machine learning and interaction
+  with GPUs. We cannot recommend its use strongly enough. It can be
+  built using [spack](https://github.com/llnl/spack).
++ [CONDUIT](https://github.com/llnl/conduit) is used to ingest
+  structured data produced by scientific simulations.
+
+
+## Building with [Spack](https://github.com/llnl/spack)
+
+Some variation on the theme of `spack install lbann`. bvanessen should
```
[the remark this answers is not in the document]
benson31: #796 Done.

bvanessen:
I apologize for merging that while it was still under review.  I wanted to get it out so that we had some updated documentation.  I also wanted to test that the new doc system worked.  We (I mean Tom) will make sure that these comments get addressed.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

docs/BuildingLBANN.md
timmoon10: I'm not sure if we actually use `LBANN_SEQUENTIAL_INITIALIZATION` anywhere. I think that functionality was folded into `LBANN_DETERMINISTIC` in PR #363.

PR #798, opened 2018-12-27
simplified sample list framework

JaeseungYeom:
simplified sample list
- no template type for easier and faster prototyping
- exclusive/inclusive list handling
- tested offline (separately from LBANN as in tools/sample_list_test_jag
- utilize the m_shuffled_list in the base generic_data_reader class, and retire the local indexing mechanism as well as the local data file ingestion that were specifically developed for data_reader_jag_conduit.
- provide [] operator to sample_list, oblivious to whether the sample list is distributed (as we do currently) or centralized (trainer master partitions and distribute the list)
- Override shuffle_indices(gen) to clear/close open file descriptors, which are cached on-demand.
- model master reader reads a sample list and broadcasts within the model.

Future work to be done in follow up PRs
- make sample_list template of sample name type and the type for the data file description to which that sample belongs to. The former is string by default, but can be size_t for instance. The latter can be a pair of file name string (or index to a loop up table) and open file descriptor.
- utilize a look up table instead of using filename string in internal sample list.
- Centralized scheme: This is not a priority but a future improvement. Currently, [] operator is future-proof. However, the distribution method relies on native MPI calls, which needs to be changed to the wrappers in lbann::comm. Then, data_reader_jag_conduit::load() needs to implement the option similar to the way done in the tools/sample_list_jag/main_centralized.cpp


tools/sample_list_test_jag/conduit_load.cpp
```diff
@@ -0,0 +1,688 @@
...
+  dreader.set_independent_variable_type(independent);
+
+  const std::vector< std::vector<variable_t> > dependent = {{response_t}};
+  dreader.set_dependent_variable_type(dependent);
+
+  const std::vector<std::string> image_keys_manual = {
+    "(0.0, 0.0)/0.0",
+    "(90.0, 0.0)/-0.01",
+    "(90.0, 78.0)/-0.03"
+  };
```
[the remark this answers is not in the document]
JaeseungYeom: So, we can use any time index as needed. The above lines you referenced are just initialization example to show that it is not hard-coded. However, currently ML folks have only explored the data at bang time, and therefore we only use such
https://github.com/LLNL/lbann/blob/492eaf33be9c09ecc6963f1595e1b0ed54eb8650/model_zoo/models/jag/data_reader_jag_conduit.prototext#L22

bvanessen:
With the current set of commits, I now have this running on the 3model version of the CycleGAN.

samadejacobs:
Ok, I will take a look. How was your larger testing? I realized that is where things break most of the time.

davidHysom:
I didn't have any problems. I re-ran detect_corruption on the 100M set, and all seems OK.

bvanessen:
I was working on the 1M data set, but something broke in Dave’s update of the sample list this morning.  I am working on it.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

include/lbann/data_readers/sample_list_jag_impl.hpp
```diff
@@ -329,6 +329,9 @@ inline void sample_list_jag::clear() {
   m_sample_list.clear();
 }
 
+template <class Archive> void sample_list_jag::serialize( Archive & ar ) {
```
JaeseungYeom: better make this inline

benson31: Just FYI, function templates are automatically inline.

JaeseungYeom: I see. Good to know.

benson31: Technically, I mean function templates are allowed to violate the ODR by default. _Specializations_ are not, so if you were to specialize this function, `inline` could (should) decorate the specializations. (Don't specialize this function -- overload it instead, but that's another issue.) If you want to actually force the compiler to inline the definition, from an ISO C++ level that's not possible, but compilers might have custom annotations to give programmers more influence.

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -943,98 +762,126 @@ void data_reader_jag_conduit::load() {
...
+    std::cout << "Time to load sample list: " << tm2 - tm1 << std::endl;
   }
+}
+
+void data_reader_jag_conduit::load_list_of_samples_from_archive(const std::string& sample_list_archive) {
+  // load the sample list
+  double tm1 = get_time();
+  std::stringstream ss(sample_list_archive); // any stream can be used
+
+  cereal::BinaryInputArchive iarchive(ss); // Create an input archive
```
JaeseungYeom: I have never heard this serialization library before, but it seems very useful and easy to use. I had to quickly glance at their web documentation to get a idea. One question I had was that is it ok to use binary archive with stringstream? They mention to set std::ios::binary when using with stream for binary archive.

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -943,98 +762,126 @@ void data_reader_jag_conduit::load() {
...
+    // Make sure to have all the reader instances be aware of the number of
+    // samples in use regardless of whether it participates data ingestion
+    // or not.
+    sample_list_header hdr = m_sample_list.load_header(sample_list_file);
+    // Avoid reading the entire list file but only reads the header.
+    m_shuffled_indices.resize(hdr.get_sample_count());
+  } else {
+    /// The use of these flags need to be updated to properly separate
+    /// how index lists are used between trainers, models, and ranks
+    if (m_list_per_trainer || m_list_per_model || m_list_per_rank) {
```
JaeseungYeom: I know this is a place holder for future extension. However, how is m_list_per_trainer going to be different from the case handled in else block.

JaeseungYeom:
Building fails with the following error:

```
CMake Error at CMakeLists.txt:192 (find_package):

  Could not find a package configuration file provided by "CEREAL" with any
  of the following names:

    CEREALConfig.cmake
    cereal-config.cmake
```

bvanessen:
Make clean


Brian C. Van Essen

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Jan 7, 2019, at 4:42 PM, Jae-Seung Yeom <notifications@github.com<mailto:notifications@github.com>> wrote:


Building fails with the following error:

CMake Error at CMakeLists.txt:192 (find_package):

  Could not find a package configuration file provided by "CEREAL" with any
  of the following names:

    CEREALConfig.cmake
    cereal-config.cmake

JaeseungYeom:
Also, I can remove the temporary testing harness code I put under tools/sample-list-jag if no one objects. Otherwise, I can update the Makefile to find cereal.

bvanessen:
I think that we should remove the test harness because it will quickly get out of date (if it isn’t already).

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

src/data_readers/data_reader_jag_conduit.cpp
bvanessen: Each trainer will read the file with one rank and then distribute internally.  So very similar, just need to figure out how to make it work and also back out the is_partitioned code.

src/data_readers/data_reader_jag_conduit.cpp
bvanessen: I don't see how to do that with a stringstream.  Also pulling from their example I didn't see that trait set on a stringstream.

src/data_readers/data_reader_jag_conduit.cpp
JaeseungYeom: For is_partitioned. I can think of two approaches. One is to partition the sample list into the number of trainers (which is easily done by calling `set_num_partitions()`) and have each trainer master construct a new sample list using the partitioned sample list. We already have interfaces to support it. Another approach is just to give a different sample list file to each trainer. We can use glob to get the list of sample list files, then figure out who gets what based on the file name or a header line in each list file.

bvanessen:
We are going to avoid having lbann partition the data if I remember our last conversation about this.  It will be on the user to provide separate index lists for each trainer.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

bvanessen:
@samadejacobs please test this, I now have the per-rank index list working and the performance and training quality matches the existing head of develop.

samadejacobs:
I have tested (on 8 ranks of Pascal) the index_list _per_reader version of this, it works.

bvanessen:
I am going to merge this PR because the index lists and data store with bundled super_node data exchange are working.  There is still some code cleanup to be done and fixing support for data store exchange with a per sample approach.  However, this PR has gotten too big and is encompassing other improvements that need to be made available to develop.

ndryden:
Just a note, this PR introduced the compile error reported in #871-- specifically commit `[fee533b](/LLNL/lbann/pull/798/commits/fee533b4b300062ccbe1c53c94e813f88c852799)` added the offending line. We should probably make sure PRs are re-reviewed after extensive changes before merging them.

PR #799, opened 2018-12-30
adding option to read filenames from a filelist, instead of using glob.

davidHysom:
usage:
  // one-time pre-processing:
  $ find /p/lscratchh/brainusr/datasets/100M -name '*bundle' > 100M_filelist

  // cmd line
  // add to cmd line: --filelist=100M_filelist

P_0 reads the filelist, and bcasts to other procs


davidHysom:
This PR is no longer relevant, due to the way our thinking about data_reader_jag_conduit has evolved.

PR #805, opened 2019-01-03
Added a command line flag that can disable the background I/O fetch.

bvanessen:
This option will set a flag in each model that will then disable the
input layers from asynchronously fetching data.  When disabled, all
data will be fetched (using the thread pool) at the start of the
fp_compute call.

### DIFF
--- a/src/utils/lbann_library.cpp
+++ b/src/utils/lbann_library.cpp
@@ -179,7 +179,7 @@ model *build_model_from_prototext(int argc, char **argv,
                                    pb.model());
     model->setup(io_thread_pool);
 
-    if(opts->has_bool("disable_background_io_activity") && opts->get_bool("disable_background_io_activity")) {
+    if(opts->get_bool("disable_background_io_activity")) {
       model->allow_background_io_activity(false);
     }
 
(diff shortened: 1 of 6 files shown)


src/utils/lbann_library.cpp
```diff
@@ -179,6 +179,10 @@ model *build_model_from_prototext(int argc, char **argv,
                                    pb.model());
     model->setup(io_thread_pool);
 
+    if(opts->has_bool("disable_background_io_activity") && opts->get_bool("disable_background_io_activity")) {
+      model->allow_background_io_activity(false);
```
davidHysom: You don't need has_bool(); The code was changed; instead of throwing an exception if the option doesn't exist, 'false' is returned. (other get_XX() methods still throw exceptions, since there's no reasonable default value)

bvanessen: Thanks.

bvanessen:
@timmoon10 No this behavior does not belong in the data reader because it is a property of the input layer and when it calls the data reader fetch function.  I placed this at the model level specifically, so that for any input layers, they would have consistent functionality for the model.

PR #806, opened 2019-01-03
Fix cmake 3.12 support

benson31:
NVCC throws a tantrum if you pass the `libcublas_device.a` library twice on the device-link line. This avoids that.


CMakeLists.txt
```diff
@@ -209,6 +181,20 @@ if (NOT OpenCV_FOUND)
 endif ()
 set(LBANN_HAS_OPENCV ${OpenCV_FOUND})
 
+# CUDA-ness of LBANN is 1:1 with Hydrogen. Iff Hydrogen has CUDA, LBANN gets CUDA.
```
bvanessen: Do you really mean if-and-only-if hydrogen has CUDA, or is the iff just a typo.

CMakeLists.txt
```diff
@@ -416,21 +401,6 @@ if (LBANN_HAS_ALUMINUM)
...
-  endif ()
-  target_link_libraries(lbann PUBLIC ${cuBLAS_LIBRARIES})
-
-  if(LBANN_HAS_NCCL2)
-    target_link_libraries(lbann PUBLIC cuda::nccl)
-  endif()
-endif()
-if(LBANN_HAS_CUDNN)
-  target_link_libraries(lbann PUBLIC ${cuDNN_LIBRARIES})
-endif ()
```
bvanessen: Are these getting included elsewhere and that is why they are removed from here?

CMakeLists.txt
benson31: I mean if and only if. #mathematicianed.

CMakeLists.txt
benson31: They all get added to the cuda::toolkit IMPORTED target.

bvanessen:
Looks good, but I haven't tested it yet.

benson31:
Builds on pascal with cuda 9.2 and cuda 10.0 with appropriate Hydrogen branch. Builds on Lassen with gcc/7.3.1 and cuda 9.2.

benson31:
IMO, this is good to go now.

benson31:
Just merged LLNL/elemental#64 so this should build with CUDA 9 or CUDA 10 out-of-the-box.

PR #807, opened 2019-01-03
LTFB communication with checkpoint files

timmoon10:
### The status quo
During an LTFB tournament, every trainer goes through its weights and calls sendrecv on the values. The data is exchanged directly between corresponding ranks in partner trainers.

Pros:
- Fast and compatible with GPU data.

Cons:
- Currently only exchanges weight values. Things get cumbersome if we also want to exchange optimizer state or hyperparameters.
- It doesn't work for heterogeneous model architectures.

### The new approach
During an LTFB tournament, every trainer saves a checkpoint for its model and loads a model from another trainer. We currently use "shared" checkpoints, although "distributed" checkpoints are a next step.

Pros:
- Exchanges optimizer state and hyperparameters, so we can now use LTFB for hyperparameter exploration, e.g. with PR #784.
- Reuses code for checkpointing.

Cons:
- This PR doesn't work by itself since loading checkpoint data for data readers causes weird behavior. You must disable checkpoint functionality in the input layer, or else the program will hang (see PR #808 for this hack).
- Communication goes through the file system, which is slow and swells up our disk space usage. If LTFB tournaments are infrequent and make up a small portion of runtime, this is not _that_ bad. If we can develop a wire format for model checkpoints, we should use it in conjunction with MPI.
- Our checkpoints do not store model architecture information, so we don't support LTFB with heterogeneous models. If our checkpoints were self-sufficient and could instantiate models from scratch, it would be easy to add support for heterogeneous LTFB.
- Our checkpoint functionality is old and badly in need of refactoring.

### Usage
I've been able to get LTFB working on AlexNet with 5 trainers (5 Pascal nodes, 2 procs/node, 2 procs/trainer) with something like the following:
```
# Make sure to remove the imcomm callback
callback {
  ltfb {
    batch_interval: 40
    metric: "categorical accuracy"
    weights: "conv1_kernel conv1_bias" # Leave black to exchange all weights
    low_score_wins: false
    communication_algorithm: "checkpoint_file" # Use "sendrecv_weights" for old algorithm
  }
}
```

### DIFF
--- a/include/lbann/callbacks/callback_ltfb.hpp
+++ b/include/lbann/callbacks/callback_ltfb.hpp
@@ -61,6 +61,46 @@ namespace lbann {
 class lbann_callback_ltfb : public lbann_callback {
 public:
 
+  /** LTFB communication scheme.
+   *
+   *  The specifics of these algorithms are experimental and will be
+   *  in flux.
+   */
+  enum class communication_algorithm {
+    /** Directly exchange weights values with sendrecv.
+     *
+     *  Corresponding ranks in partner models will iterate through
+     *  their weights and exchange values with sendrecvs.
+     *
+     *  Notes:
+     *    - Requires all models to be identical aside from their
+     *      weights values, so this is not suitable for hyperparameter
+     *      or model architecture exploration.
+     *    - Optimal if communication performance between ranks is
+     *      uniform and independent. If intra-model communication is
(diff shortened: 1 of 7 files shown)


ndryden:
> If we can develop a wire format for model checkpoints, we should use it in conjunction with MPI.

Sounds like what protocol buffers are for. ;)

Issue #787, opened 2018-12-20
CNPY is not actually optional


benson31:
Attempting to build LBANN without CNPY fails almost immediately. Proposed solutions are to either fix the compilation errors or remove `LBANN_WITH_CNPY` from the CMake and make CNPY a required dependency. The former seems like the right move.

ndryden:
Well, without CNPY we can't load NumPy data. That seems like a widely-used enough thing that perhaps having it be required is worth-while. This impression might be wrong though (considering we don't have any _current_ applications using NumPy data).

benson31:
IMO, a minimal build of LBANN could be useful for testing. No CNPY, no CUDA, no OpenCV -- basically just synthetic data. @timmoon10 tells me that things like gradient checking should still work in that regime, and unit tests don't use external data at all. As LBANN acquires users, more of them will complain "why do I need to install this thing if I don't have numpy data??" Basically I'm just suggesting that we make the thing that currently looks optional actually optional.

timmoon10:
Looks like the dump outputs callback just assumes you have CNPY. Mea culpa.

Did you see any errors coming out of the Numpy data reader?

benson31:
Just `#include <cnpy.h>` header not found.

ndryden:
> As LBANN acquires users, more of them will complain "why do I need to install this thing if I don't have numpy data??"

Or, they'll complain "I need NumPy data, why do I have to reinstall LBANN?!?!?"

Really, they should just be satisfied with something that's easier to build from source than TensorFlow.

benson31:
"I want to train at float at 10am, double at 11am, and fp16 at noon. Why do I need to build three LBANNs?"

ndryden:
[Less farfetched than you might think](https://arxiv.org/abs/1812.08011). 😉

benson31:
[...]

PR #813, opened 2019-01-08
Minor improvements to dump outputs callback

timmoon10:
Changes:
- Check the `prefix` parameter to see if a directory needs to be created. I previously got a bunch of errors when I left `prefix` as an empty string.
- The `prefix` parameter is now a file prefix. For instance, setting the prefix to "something" will create files that look like "somethingmodel0-training-epoch0-step0-conv1-output0.csv". Setting the prefix to "something/" will create a directory called "something" and put the files inside, e.g. "something/model0-training-epoch0-step0-conv1-output0.csv". If we prefer the old behavior, we should change the parameter name from `prefix` to `dir`.
- Hide CNPY functionality inside `LBANN_HAS_CNPY` macros. I haven't checked yet if we can compile LBANN without CNPY. See Issue #787.

### DIFF
--- a/include/lbann/callbacks/callback_dump_outputs.hpp
+++ b/include/lbann/callbacks/callback_dump_outputs.hpp
@@ -31,10 +31,10 @@
 
 namespace lbann {
 
-/** Dump layer output tensors to files.
- *  This callback saves a file for each output tensor of each selected
- *  layer, computed at each mini-batch step. Output files have the
- *  form
+/** @brief Dump layer output tensors to files.
+ *
+ *  Saves a file for each output tensor of each selected layer,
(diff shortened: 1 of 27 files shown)


bvanessen:
Prefix really should be a directory.  We actually don't want to append anything to the beginning of the model files since I don't know that the restart code will properly pick it up.  I like your suggestion to rename the prototext and internal fields to reflect the proper behavior.

timmoon10:
I've made the change so that the `prefix` option has been renamed to `directory`. All of the prototexts in the model zoo have been updated. Setting `directory` to "something" will create outputs that look like "something/model0-training-epoch0-step0-conv1-output0.csv".

PR #814, opened 2019-01-08
Bump OpenCV version.

ndryden:
Bump OpenCV version from 3.3.0 to 3.4.5 (the most recent release on the 3.x branch). OpenCV has moved on to 4.x now, so we should look at that, but I'm not sure whether they make major API changes.

Built/seems to run fine on Sierra. You may want to test other systems.

### DIFF
--- a/superbuild/opencv/CMakeLists.txt
+++ b/superbuild/opencv/CMakeLists.txt
@@ -41,7 +41,7 @@ else ()
     CACHE STRING "The URL from which to clone OpenCV")
 endif ()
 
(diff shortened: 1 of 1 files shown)


benson31:
This is fine. For posterity: one can experiment with other versions by changing the `OPENCV_TAG` in the Superbuild invocation by adding `-DOPENCV_TAG=<tag>`, where `<tag>` is a commit-ish from OpenCV's repo (whichever is pointed to by `OPENCV_URL`).

Issue #774, opened 2018-12-18
Invalid data reader position on validation dataset


ndryden:
When running in some configurations on both Catalyst and Lassen, using both AlexNet and ResNet-50, with both real and synthetic data, I get the following error after training has completed (thus, I believe it's occurring during validation):
`/usr/WS1/dryden1/lbann-ndryden/src/data_readers/data_reader.cpp 107 :: generic data reader load error: !position_valid -- current pos = 12819 and there are 12811 indices`
(There are slightly different numbers depending on the rank.)

Here's a sample backtrace from a debug build running on one node (two ranks) on Catalyst:
```
--------------------------------------------------------------------------------
[0] Epoch : stats formated [tr/v/te] iter/epoch = [496/6/20]
            global MB = [ 256/ 256/ 256] global last MB = [ 115  /   1  / 136  ]
             local MB = [ 256/ 256/ 256]  local last MB = [ 115+0/   1+0/ 136+0]
--------------------------------------------------------------------------------
Model0 (instance 0) training epoch 0 objective function : 12.0982
Model0 (instance 0) training epoch 0 categorical accuracy : 0.0883037%
Model0 (instance 0) training epoch 0 top-5 categorical accuracy : 0.472267%
Model0 (instance 0) training epoch 0 run time : 18874.3s
Model0 (instance 0) training epoch 0 mini-batch time statistics : 38.053s mean, 39.4247s max, 17.1164s min, 1.00503s stdev
****************************************************************
/usr/WS1/dryden1/lbann-ndryden/src/data_readers/data_reader.cpp 107 :: generic data reader load error: !position_valid -- current pos = 1281 and there are 1281 indices
Stack trace:
   0: lbann::stack_trace::get[abi:cxx11]()
   1: lbann::exception::exception(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool)
   2: lbann::generic_data_reader::fetch_data(El::Matrix<float, (El::Device)0>&, El::Matrix<long long, (El::Device)0>&)
   3: lbann::fetch_data_functor::operator()(El::Matrix<float, (El::Device)0>&, El::Matrix<float, (El::Device)0>&, El::Matrix<long long, (El::Device)0>&, lbann::generic_data_reader*) const
   4: lbann::partitioned_io_buffer::fetch_to_local_matrix(lbann::generic_data_reader*, execution_mode)
[...]

PR #816, opened 2019-01-10
Bug fix for invalid data reader position

bvanessen:
Fixed a bug where the data reader's current position is not valid but within # ranks per model of the end of the data set (e.g. it is a rank with no valid data on the last iteration).


bvanessen:
Fixes #774

ndryden:
Doesn't crash, but for some reason I'm observing an objective function cost that's literally 1000x what I'd expect.

ndryden:
I'm no longer seeing the issue with the objective function (I think that was just due to me having some unrelated code lying around in my repo that interacted badly). I am still seeing a performance regression on Sierra, not sure why.

ndryden:
Performance issue seems to have just been some significant variance on Sierra. I'm now seeing pretty consistent performance.

PR #822, opened 2019-01-11
LTFB sendrecv_weights algorithm will communicate optimizer state

timmoon10:
LTFB with the "sendrecv_weights" algorithm will communicate the optimizer state for SGD (velocity) and Adam (moment1, moment2). Note that this doesn't transfer hyperparameters, so you still need to use the "checkpoint_file" algorithm for hyperparameter exploration. I've also done some refactoring of SGD and Adam, mostly adding `unique_ptr`s and cleaning up documentation.


samadejacobs:
Tim and I talked about this offline. Currently this supports Adam and SGD. Future work will be to support any kind of optimizer state.

PR #825, opened 2019-01-16
Correct batchnorm defaults

ndryden:
We claim that batchnorm's decay defaults to 0.9 and its epsilon to 1e-5. But we just pass the protobuf parameters directly to the layer constructor, and protobuf will default them to 0.0 if not present. This fixes that.


ndryden:
> Is there any case where someone might want to set epsilon to zero?
The like floating point exceptions?

If this use case shows up, we can address it, but I don't think it will be common.

PR #828, opened 2019-01-18
Remove deprecated objective functions and target layer

### LINKED ISSUE #193: A proposal for a new objective function paradigm

### LINKED ISSUE #678: Cull deprecated features

timmoon10:
This eliminates the last vestiges of the old target-layer-based objective function paradigm. This should be fairly painless since we already switched to the layer-based objective function paradigm several months ago (see Issue #678). All the gradient checking unit tests run and pass. AlexNet also runs (1 Pascal node, 1% of ILSVRC-2012) with no substantial change to learning behavior or performance.

This closes #193 and closes #678.


timmoon10:
Replacements for deprecated objective functions, using the Python interface in https://github.com/ndryden/lbann/pull/1:
- `binary_cross_entropy`: `BinaryCrossEntropy([pred, label])`
- `cross_entropy`: `CrossEntropy([pred, label])`
- `cross_entropy_with_uncertainty`:
```python
def cross_entropy_with_uncertainty(pred, label):
    masked_pred = Multiply([pred, label]) # Assuming `label` is Boolean
    pred_sum = Reduction(masked_pred)
    return Negative(Log(pred_sum))
```
- `geom_negloglike`:
```python
def geom_negloglike(pred, label):
    ones = Constant(hint_layer=pred, value=1.0)
    term1 = Multiply([label, Log(Subtract([ones, pred]))])
    term2 = Log(pred)
    full = WeightedSum([term1, term2], scaling_factors='-1.0 -1.0')
    return Reduction(full)
```
- `mean_absolute_deviation`: `MeanAbsoluteError([pred, label])`
- `mean_absolute_error`: `MeanAbsoluteError([pred, label])`
- `mean_squared_error`: `MeanSquaredError([pred, label])`
- `poisson_negloglike`:
```python
def poisson_negloglike(pred, label): # Requires log-gamma, which is not yet supported
    ones = Constant(hint_layer=pred, value=1.0)
    term1 = pred
    term2 = Multiply([label, Log(pred)])
    term3 = LogGamma(Add([label, ones]))
    full = WeightedSum([term1, term2, term3], scaling_factors='1.0 -1.0 1.0')
    return Reduction(full)
```
- `polya_negloglike`:
```python
def polya_negloglike(pred, label): # Requires log-gamma, which is not yet supported
    count = Reduction(label)
    alpha_sum = Reduction(pred)
    lgamma_alpha_sum = Reduction(LogGamma(pred)) 
    lgamma_alpha_level_count_sum = Reduction(LogGamma(Add([pred, label])))
    return WeightedSum([LogGamma(alpha_sum),
                        LogGamma(Sum([count, alpha_sum])),
                        lgamma_alpha_level_count,
                        lgamma_alpha_sum],
                       scaling_factors='-1.0 1.0 -1.0 1.0'])

```
- `group_lasso_weight_regularization`:
```python
def group_lasso_weight_regularization(weights, height, width):
    w = WeightsLayer(weights=weights, dims='%d %d'.format(width, height))
    slice = Slice(w, slice_axis=0, slice_points=' '.join(range(width+1)))
    cols = []
    for _ in range(width):
        cols.append(Sqrt(L2Norm2(slice)))
    return Sum(cols)
```
- `l1_weight_regularization`:
```python
def l1_weight_regularization(weights, dims):
    w = WeightsLayer(weights=weights, dims=' '.join(dims))
    return L1Norm(w)
```

bvanessen:
@timmoon10 Can you please put these functions into the new python module library.

timmoon10:
I'm a bit wary about including the negative log-likelihoods. As far as I can tell, they're not natively supported in PyTorch, Keras, or Caffe (you might be able to compute log-likelihoods for geometric and Poisson distributions in TensorFlow if you delve into their support for statistical distributions, but I'm not sure). These are expert functionalities for very statistically sophisticated models and I don't think it will be worthwhile to maintain such specialized operations.

I think group LASSO has a similar problem in that it's still a rather exotic technique that isn't supported by anyone else. The L1 regularization seems common enough, though, to warrant including in the Python interface.

PR #830, opened 2019-01-23
Introducing trainer construct and naming convention in comm class

bvanessen:
Updating internal naming convention to properly reflect the differences 
between trainers and models. A trainer is a set of
parallel resources that is used to train one or more deep neural
network models (i.e. a set of MPI ranks on multiple nodes and GPUs).
A model refers to a specific neural network architecture and its
associated set of weights.  To clarify this relationship, the entire
comm class has updated to reflect that communication is scoped within
or between trainers (rather than models).


bvanessen:
I don't agree that the inter-trainer communication doesn't make sense.  They just don't make sense if the models being trained are different.  I think that for a multi-trainer data parallel execution they continue to make sense.  Quantization and compression between trainers is still an interesting option and I believe that it is something that we will support with the new Al interface, correct?

bvanessen:
I merged in #811 into develop and here.

ndryden:
Is this ready for a review or is it still in-progress?

Issue #155, opened 2017-12-22
Better pointer handling


timmoon10:
As our code becomes increasingly complicated, I think better pointer management would be good for robustness and simplification. I propose the following changes:
- [x] Layers maintain input and output matrices in vectors of `unique_ptr`s (see #153).
- [ ] Elemental matrices are stored in `unique_ptr`s whenever possible.
- [x] Models maintain layers in a vector of `unique_ptr`s.
- [x] Models maintain weights in a vector of `unique_ptr`s
- [ ] Models maintain metrics in a vector of `unique_ptr`s.
- [x] Models maintain callbacks in a vector of `unique_ptr`s.
- [x] Models maintain objective function in a `unique_ptr`.
- [x] Models maintain default optimizer in a `shared_ptr`.
- [x] Objective functions maintain terms in a vector of `unique_ptr`s.
- [x] Weights maintain optimizer in a `unique_ptr`.
- [x] Refactor GPU support so that layers no longer need to maintain a reference to a cudnn_manager object. Instead, we should create a `cuda` namespace with globally accessible functions. This may become moot with Hydrogen (@benson31).
- [ ] We should prefer function arguments that are references rather than pointers.

ndryden:
I think this would be quite helpful for thinking about what objects own what data, who is responsible for freeing it, etc.

If we do this, we should probably also switch to using `std::make_shared` and `std::make_unique`. (The latter might be slightly troublesome with some pedantic compilers, because the committee accidentally left it out of C++11, so it's only officially present in C++14.)

benson31:
```cpp
#include <memory>
namespace lbann
{

template <typename T>
using MakeShared = std::make_shared<T>;

#if __cplusplus < 201402L

template <typename T, typename... Ts>
std::unique_ptr<T> MakeUnique(Ts&&... params)
{
    return std::unique_ptr<T>(new T(std::forward<Ts>(params)...));
}

#else

template <typename T>
using MakeUnique = std::make_unique<T>;

#endif
```
You could also keep things in the `make_shared` or `make_unique` format if you prefer the STL look and feel (I usually do, but I've been reading a lot of Elemental lately).

Real `std::make_unique` handles a few other cases that we could add if needed (e.g., `T[]`), but for me, this has covered the 99% use case.

benson31:
```cpp
}// namespace lbann
```
Oops.

timmoon10:
[...]

PR #836, opened 2019-01-25
Models store layers in smart pointers

timmoon10:
The lifetime of a layer is now explicitly managed with `unique_ptr`s. This is consistent with [best practices](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c149-use-unique_ptr-or-shared_ptr-to-avoid-forgetting-to-delete-objects-created-using-new) and is more compatible with @benson31's proposed refactor of the factory methods (see PR #770). I think this was the hardest part of smart-pointer-izing LBANN (see Issue #155).

The layer factory method now calls `make_unique()` and returns a `unique_ptr`. The model's `add_layer()` function takes a `unique_ptr` (per [best practices](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r32-take-a-unique_ptrwidget-parameter-to-express-that-a-function-assumes-ownership-of-a-widget)) and moves it into a `vector` of `unique_ptr`s (`m_layers`). Since smart pointers can sometimes be a pain to deal with, I've refactored the model class to only touch `m_layers` where needed. It's easier most of the time to call `get_layer()`, which returns a reference.

Running with AlexNet on Pascal hasn't shown any unusual behavior. We'll see if Bamboo picks up anything strange when it runs tonight.


include/lbann/models/model.hpp
```diff
@@ -329,39 +327,58 @@ class model {
...
+   *  In case that a layer is frozen, also freeze layers that precede
+   *  it if that makes senses for the particular model, such as
+   *  sequential or siamese.  For othe models, users can manually
+   *  control the behaivor by indicating whether to freeze each layer
+   *  in the model description prototext.
+   *
+   *  For general DAG models, users need to manually specify each
+   *  layer to freeze in the model description prototext.
    */
-  virtual void freeze_layers_under_frozen_surface();
```
timmoon10: This function has an implementation in models/model.cpp, but it's overridden by an empty function in models/directed_acyclic_graph.hpp. I've gone ahead and removed the unused implementation. In the future, we may want to think about how to properly freeze layers in DAGs.

include/lbann/models/model.hpp
```diff
@@ -125,11 +117,20 @@ class model {
...
+  /** Size of model's list of layers. */
+  El::Int get_num_layers() const noexcept;
+  /** @param pos Position in model's list of layers. */
+  Layer& get_layer(El::Int pos);
+  /** @param pos Position in model's list of layers. */
+  const Layer& get_layer(El::Int pos) const;
+  /** @brief Return list of layers in model.
+   *  @details The list is in execution order for forward propagation.
+   */
+  std::vector<Layer*> get_layers();
```
bvanessen: I have some questions about the paradigm here.  If we are using unique pointers do we want to be returning raw pointers for this function.

benson31: Ideally what you want is `std::vector<Layer&>`, but that's not an easy (or sane) thing to create. The semantics is `std::vector<SomeObserver<Layer>>` --> `std::vector<observing_ptr<layer>>` --> `std::vector<layer*>`. I think this is the best we can do. Even better, though, would be to support iterator access only, so this function can go away completely.

PR #837, opened 2019-01-25
Add setup.py for the Python frontend

oyamay:
This commit adds `setup.py` to the Python frontend (`scripts/proto`) so that users can
1. install the frontend by running `cd scripts/proto; pip intall -e .`, and
2. use it by using `import lbann` in Python in anywhere.


timmoon10:
This looks nice! We don't have to do it now, but we should rename lbann_proto.py to proto.py (and similar for lbann_modules.py) so that importing will look like `import lbann.proto`.

ndryden:
Agree on the renaming, we can do that in our next round of cleanups/updates.

PR #839, opened 2019-01-28
Feature data set metadata

bvanessen:
Creates a new metadata file that is used to store both the schema and normalization values for a given data reader.  The eliminates a lot of copy and past from the JAG Conduit data readers.

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -22,6 +22,9 @@ Internal features:
 I/O & data readers:
  - Overhauled the I/O system to use an independent background thread
    pool for fetching data
+ - Added support for data set metadata file that provides both schema
+   and normalization values unique to a given data set.  Demonstrated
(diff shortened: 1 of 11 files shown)


bvanessen:
I understand that this is very specialized to the JAG data set, but when we go to Hydra data I am told that the schema and initialization values will change, but the data reader / preprocessing will not change much.  Additionally, if we start to pull in other data sets using HDF5 / conduit, it would be nice to have a more generic data reader.  

Furthermore, in general the data readers are often different between the train and test with respect to the preprocessing steps.  I think that these data readers are so similar is an uncommon case.

bvanessen:
I don’t want it in the data reader because then it gets duplicated between systems like lassen and lustre.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #841, opened 2019-01-29
Feature simplified conduit paths

bvanessen:
Updated the JAG Conduit data reader so that the prefix paths to the
output scalar, output images, and input fields are contained in the
data_set_metadata schema prototext block.  This makes the data reader
more general purpose to changes in the layout of the data set.


bvanessen:
Note that this PR depends on #839

src/proto/proto_common.cpp
```diff
@@ -792,10 +796,10 @@ void print_help(lbann_comm *comm)
 
   std::cerr <<
        "General usage: you need to specify three prototext files, e.g:\n"
-       "  srun -n# proto --model=<string> --optimizer=<string> --reader=<string>\n"
+       "  srun -n# proto --model=<string> --optimizer=<string> --reader=<string> --metadata=<string>\n"
        "\n"
        "  However, if you are re-running an experiment from a previously saved\n"
-       "  file, you only need to specify --model=<string>\n"
+       "  file, you only need to specify --prototext=<string>\n"
```
JaeseungYeom: I think "prototext" is kind of too general and does not give much hint of what it is for.

PR #842, opened 2019-01-29
Improving Python implementation of ResNets

timmoon10:
Changes:
- Implemented ResNets as layer modules. It is now possible to import resnet50.py and use a ResNet as a component within a more complicated model, e.g. a Siamese network or a GAN.
- Implemented ResNet-{18,34,50,101,152}.
- Added command-line options to choose exported prototext file, ResNet variant, etc.
- Expanded documentation.

This is much more usable, although it's somewhat less friendly as a tutorial. I think it would be best to rename "resnet50.py" to "resnet.py" and to move it to the model zoo. We should probably make a directory called "image_classification" or "vision" and put similar Python files for LeNet and AlexNet (and maybe VGG or Inception).


scripts/proto/lbann/models/resnet.py
```diff
@@ -0,0 +1,536 @@
...
+                              name=self.name + '_bn_scale')
+        bn_bias = lp.Weights(initializer=lp.ConstantInitializer(value=0.0),
+                             name=self.name + '_bn_bias')
+        self.bn_weights = [bn_scale, bn_bias]
+        self.bn_stats_aggregation = bn_stats_aggregation
+
+        # Initialize ReLU
+        self.relu = relu
+
+    def __call__(self, x):
```
ndryden: We should be consistent about whether we overload `forward` or `__call__`. Both are equivalent, but it's probably prettier for users to overload `forward`.

timmoon10: Good catch. Fixed.

PR #854, opened 2019-02-04
NumPy .npz data reader

oyamay:
This PR adds `numpy_npz_reader` which reads data stored in a NumPy (.npz) file.
`numpy_npz_reader` supports:
* multi-dimensional input data
* labels and multi-dimensional responses (optional)
* float, double, and int16 data types
   * An user-given scaling factor `scaling_factor_int16` can be applied to input data

The `data_reader_merge_samples` data reader supports `numpy_npz_reader` with an optional field `merge_skip_overlapped`, which skips some .npz files so that each training process has disjoint dataset.

This PR also adds `data_reader_mnist_numpy_npz_int16.prototext` and `mnist_to_npy_and_npz.py` to test the data reader by using the MNIST dataset.


oyamay:
I'm current working on the `merge_skip_overlapped` option so I'll push some commits to this soon.

oyamay:
I reverted the changes around `merge_skip_overlapped` since this function will be replaced with the sample list function (#898).

PR #856, opened 2019-02-05
Merge lbplot to the Python interface

oyamay:
This PR merges the `lbplot` package to the `lbann` package so that users do not have to install two different Python packages related to LBANN.

Some changes:
* Output files with `.txt` is treated as LBANN output files.
* The `-p` option of `lbplot` is removed since it is a required argument.

A typical line to use `lbplot` looks like:
```shell
lbplot outout1.txt outout2.txt -n experiment1 experiment2
```


oyamay:
Added two options:
* `--no-accuracy`: Do not plot accuracy curves. This option is used for output files which do not contain categorical accuracy.
* `--merge-train-val`: Plot training/validation curves in the same figure. Validation curves are plotted as dashed lines with the same color to the training curves.

PR #865, opened 2019-02-07
Model class for the Python frontend

oyamay:
This PR defines the `lbann.proto.Model` class, so that additional Python functions (such as graph visualization) can be implemented to the model class.
This also implements `lbann.proto.save_prototext`, which saves a prototext which contains a model, a data reader, and/or an optimizer (all objects are optional).
`lbann.proto.save_model` internally uses the class/function without change of the semantics.

### DIFF
--- a/scripts/proto/lbann/proto.py
+++ b/scripts/proto/lbann/proto.py
@@ -496,53 +496,70 @@ class Callback:
 _generate_classes_from_message(Callback, lbann_pb2.Callback)
 
 # ==============================================
+# Model
+# ==============================================
+
+class Model:
+    """Base class for models."""
+
+    def __init__(self, mini_batch_size, epochs,
+                 layers, weights=[], objective_function=None,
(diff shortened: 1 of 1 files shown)


davidHysom:
> This also implements `lbann.proto.save_prototext`, which saves a prototext which contains a model, a data reader, and/or an optimizer (all objects are optional).
When you run lbann is already does this: saves the model, reader, and optimizer to a file, with additional information that lets you re-run the exact experiment.
How does what you're suggesting differ from, or add to that? Is it meant as a replacement?

oyamay:
It is just a generic function to save any combinations of LBANN PB objects into a prototext file.
Since all of the arguments are optional, this function can save a prototext which contains a single PB object (like `save_model` does), or a set of objects (like LBANN does).

And after everything (including data readers and optimizers) is written in Python, we don't need to split them into different prototexts, so this function will be needed in near future.

PR #869, opened 2019-02-08
Optimizing the implementation of the new Conduit-based generic data store

bvanessen:
Working on optimizing the Conduit-based data store:

1) Add support for per-sample exchange
2) Minimize data movement
3) Potentially remove sample index from conduit node hierarchy in data store
4) avoiding MPI sends to self


bvanessen:
@davidHysom Here is a PR that we can share for immediate improvements.

PR #875, opened 2019-02-12
Adding directions for how an LBANN user can install with Spack.

bvanessen:


docs/BuildingLBANN.md
```diff
@@ -96,8 +96,40 @@ The following third-party packages are **optional**.
...
+  ```bash
+  cd <path to spack repo>/spack_environments/users/llnl_lc/x86_64_gpu/
+  spack install
+  spack env loads
+  source loads
+  ```
+
+Note that there are a number of options for all of these packages and
+can be viewed via commands such as `spack info lbann`.  To specify the
+compiler you can add options such as `%gcc@7.3.0`.  For further
```
benson31: `^gcc@7.3.0`?

bvanessen:
No compilers are specified with a %gcc@7.3.0

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #880, opened 2019-02-13
fix issue where comms are moved erroneously

benson31:
Missed this when I made the changes for the lazy init stuff (then again, so did all the reviewers...)


benson31:
I don't disagree. When `Comm`s were passed by value, this was the right thing to do; however, with the update to the `Comm` semantics, that was incorrect. It slipped my mind, so I only did one find-and-replace (`Comm c`->`const Comm& c`) instead of two... Glad this merged and all is totally and 100% right with the world now... ;)

PR #887, opened 2019-02-15
Make Cub always log certain performance-critical events.

naoyam:
Once stabilized, it is generally expected that:
- Memory request does not actually acquire a new memory chunk.
- Memory release just returns the pointer to the memory pool and never
calls cudaFree.

These events can be a significant performance penalty. Cub will log
messages when they happen only if its debug option is enabled, but the
options also enables other logging, yielding extremely verbose output.
This patch only enables logging at the above events no matter the debug
option is enabled.


naoyam:
@benson31 I added a patch command to the superbuild file for Cub. I'm not sure the change is what superbuild should look like, but at least it works on Lassen. Let me know if there is a better way.

PR #888, opened 2019-02-15
Some more smart pointers and enhanced exception safety

benson31:
This just cleans up some smart pointer usage. The significant change to the front end is having to pass `comm.get()` into functions. The correct fix for this is to change the API to accommodate `*comm`, but that seems like a bigger endeavor than I was looking for today. In general, references should be used instead of pointers. Also, in C++11+, functions should prefer to return constructs by value and it should be safe to do so. There is guaranteed moves and likely copy elision. Maybe someday we'll even get C++17's guaranteed copy elision... But that's probably still 7-8 years away; it's only 2019... ;)


src/utils/lbann_library.cpp
```diff
@@ -48,193 +48,189 @@ std::shared_ptr<thread_pool> construct_io_thread_pool(lbann_comm *comm) {
...
-      }
+  //under development; experimental
+  if (opts->has_bool("use_data_store") && opts->get_bool("use_data_store")) {
+    if (master) {
+      std::cout << "\nUSING DATA STORE!\n\n";
     }
-
-    if (opts->has_string("create_tarball")) {
-      finalize(comm);
-      return 0;
```
benson31: This code block was just wrong and if it were ever used, it would, with high probability, result in a segfault. In the single-model case, one would probably call `model_ptr->anything()` on a null `model_ptr` and in the multimodel case, one's next call to this function would attempt to operate on a finalized comm. So, question: What is the intent here? What behavior should happen in the multi-model case? I'm guessing the right behavior is something like:
```cpp
int main(int argc, char** argv)
{
    create_array_of_models();
    if (opts->has_string("create_tarball")) {
        return EXIT_SUCCESS;
    }
    train_models();
}```
But I wasn't 100% sure so I only added code like this to `lbann.cpp`. Please advise on what to do for other front-ends.

**Comment by benson31** [comment]
This should be good to review now.

File: src/base.cpp [hunk 2]
```diff
@@ -47,12 +47,13 @@
...
 namespace lbann {
 
-lbann_comm* initialize(int& argc, char**& argv, int seed) {
+lbann_comm_ptr initialize(int& argc, char**& argv, int seed) {
   // Initialize Elemental.
   El::Initialize(argc, argv);
   // Create a new comm object.
   // Initial creation with every process in one model.
-  auto* comm = new lbann_comm(0);
+  auto comm = lbann_comm_ptr{new lbann_comm(0), &lbann::finalize };
```
timmoon10: It bothers me that the user calls `initialize` without also calling `finalize`. It seems to me that we're essentially making `lbann_comm` a singleton object. Going further down this path, we would put `initialize` inside `lbann_comm`'s constructor and `finalize` in the destructor. This has it's own weirdness - the user must make sure it's lifespan matches the beginning and end of the program. We may also run into the case where someone might want multiple instances of `lbann_comm`, but this is not too important.

model_zoo/lbann.cpp
```diff
@@ -49,18 +49,13 @@ int main(int argc, char *argv[]) {
...
-  El::GemmUseGPU(32,32,32);
-#endif
-
   try {
     // Initialize options db (this parses the command line)
     options *opts = options::get();
     opts->init(argc, argv);
     if (opts->has_string("h") or opts->has_string("help") or argc == 1) {
-      print_help(comm);
-      finalize(comm);
```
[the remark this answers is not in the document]
benson31: `finalize` is not skipped. (That applies pre-edit)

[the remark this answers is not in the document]
benson31: Agree re another PR. That would have been a change of semantics, which I was largely trying to avoid.

Punchline: `finalize()` is still called whenn `comm` is destroyed at the end of scope.

include/lbann/proto/proto_common.hpp
```diff
@@ -30,29 +30,29 @@ void init_data_readers(
...
-void save_session(lbann_comm *comm, int argc, char **argv, lbann_data::LbannPB& p);
+void save_session(const lbann_comm& comm, const int argc, char * const* argv, lbann_data::LbannPB& p);
 
 ///
 void read_prototext_file(
-  std::string fn,
+  const std::string& fn,
   lbann_data::LbannPB& pb,
-  bool master);
+  const bool master);
```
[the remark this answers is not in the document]
benson31: "Use `const` whenever possible." Whether you are the master does not change through the function, so the value is `const`. Can't hurt.

src/base.cpp
benson31: I very deliberately did not put the `initialize` in the Ctor nor `finalize` in the dtor. That is *not* what we want to do. This makes `'lbann_comm_ptr` an _ad hoc_ RAII wrapper for MPI since `initialize` is only called in scopes in which the caller is responsible for freeing the MPI resources. The `lbann_comm` that gets returned from `initialize` is really "COMM_WORLD"; when the world blows up, nothing lives on anyway.

The "proper" fix for this is to actually RAII the interface. We should create a class `lbann::session` or something that calls `initialize()` in its ctor (i.e., when it acquires MPI/CUDA/whatever resources) and 'finalize()` in its dtor (i.e., when it returns those resources). This should be completely distinct from the `lbann_comm` object. Indeed, it will be an error to try to create an `lbann_comm` if you're not nested inside a scope in which an `lbann::session()` does not exist.

benson31: Just to clarify about the "singleton" comment: Note that `lbann_comm_ptr` is *not* wrapping every `lbann_comm` object -- only the one returned from `initialize' is managed this way. This does *not* make the `lbann_comm` object a singleton by any means. Logically, the thing that gets returned from `intialize` is "COMM_WORLD", which is "an object" that must "exist" (quotes cuz handle to opaque blahblahblah C mumbojumbo) for the entire valid MPI session. If this object gets deleted, logically the session has ended. It's now time to `finalize()`. Indeed, you can split that thing up all you want into sub-`lbann_comm` objects and this is semantically valid (and encouraged).

include/lbann/base.hpp
```diff
@@ -42,6 +42,11 @@ namespace lbann {
 // Forward-declaration.
 class lbann_comm;
 
+// Note that this should only be used to wrap the thing coming out of
+// initialize()! This will be removed when we have proper RAII around
+// these things.
+using world_comm_ptr = std::unique_ptr<lbann_comm, void(*)(lbann_comm*)>;
```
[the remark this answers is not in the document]
benson31: The first one is the thing being pointed to, an `lbann_comm` object. The second is the type of the Deleter to be used when the held object gets destroyed. The default is [std::default_delete](https://en.cppreference.com/w/cpp/memory/default_delete). Because this is only used to hold our logical "COMM_WORLD", we want `finalize()` to be called when this thing gets destroyed (and regardless of how this thing gets destroyed -- uncaught exception, unhandled signal, or normal exit). The type of `finalize()` (or more precisely of `&finalize()`) is `void(*)(lbann_comm*)`, so that's the second argument. It would also be correct to use `std::function<void(lbann_comm*)>`. I can change it to that if that would be easier to understand.

bvanessen:
Okay the that makes sense. I thought that it was supposed to be the deleter. I do think that either is okay but maybe the std::function may be more clear.

Thanks


Brian C. Van Essen

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Feb 20, 2019, at 10:10 AM, Tom Benson <notifications@github.com<mailto:notifications@github.com>> wrote:


@benson31 commented on this pull request.

In include/lbann/base.hpp<https://github.com/LLNL/lbann/pull/888#discussion_r258585587>:

> @@ -42,6 +42,11 @@ namespace lbann {
 // Forward-declaration.
 class lbann_comm;

+// Note that this should only be used to wrap the thing coming out of
+// initialize()! This will be removed when we have proper RAII around
+// these things.
+using world_comm_ptr = std::unique_ptr<lbann_comm, void(*)(lbann_comm*)>;


The first one is the thing being pointed to, an lbann_comm object. The second is the type of the Deleter to be used when the held object gets destroyed. The default is std::default_delete<https://en.cppreference.com/w/cpp/memory/default_delete>. Because this is only used to hold our logical "COMM_WORLD", we want finalize() to be called when this thing gets destroyed (and regardless of how this thing gets destroyed -- uncaught exception, unhandled signal, or normal exit). The type of finalize() (or more precisely of &finalize()) is void(*)(lbann_comm*), so that's the second argument. It would also be correct to use std::function<void(lbann_comm*)>. I can change it to that if that would be easier to understand.

include/lbann/base.hpp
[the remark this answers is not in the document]
benson31: I have updated to `std::function<void(lbann_comm*)>`.

model_zoo/lbann.cpp
```diff
@@ -74,16 +69,19 @@ int main(int argc, char *argv[]) {
     stack_profiler::get()->activate(comm->get_rank_in_world());
 
     // Initalize a global I/O thread pool
-    std::shared_ptr<thread_pool> io_thread_pool = construct_io_thread_pool(comm);
+    std::shared_ptr<thread_pool> io_thread_pool = construct_io_thread_pool(comm.get());
```
ndryden: Maybe `auto` would be nicer here?

src/base.cpp
```diff
@@ -47,12 +47,13 @@
...
 namespace lbann {
 
-lbann_comm* initialize(int& argc, char**& argv, int seed) {
+world_comm_ptr initialize(int& argc, char**& argv, int seed) {
   // Initialize Elemental.
   El::Initialize(argc, argv);
   // Create a new comm object.
   // Initial creation with every process in one model.
-  auto* comm = new lbann_comm(0);
+  auto comm = world_comm_ptr{new lbann_comm(0), &lbann::finalize };
```
ndryden: Should we be doing this with `make_unique` or something?

model_zoo/lbann.cpp
benson31: `auto` *would* be nicer here, but it would be incorrect. `construct_io_thread_pool` returns `unique_ptr` (it's a factory function that doesn't store a reference to the created object, so `unique_ptr` is correct), and explicitly declaring the type casts to `shared_ptr` since the application code will be sharing that reference. One could make a stylistic argument that it should be
```cpp
auto io_thread_pool = std::shared_ptr<thread_pool>{construct_io_thread_pool(comm.get())};
```
They are the same and I have no strong feelings about the style difference. If there are opinions one way or the other, I'm happy to change it.

ndryden: No, if we can't do `auto`, then let's keep it as is.

src/base.cpp
benson31: You cannot use `make_unique` with a custom deleter. The return type of `make_unique<T>` is explicitly `unique_ptr<T,std::default_delete<T>>`. In lieu of a custom `make_unique` implementation that accommodates deleters in some fashion, this is the best that can be done.

model_zoo/lbann.cpp
benson31: Ok, that's fine. For reference, the style shown in this comment is shown [here](https://herbsutter.com/2013/05/30/gotw-90-solution-factories), but there is no comment as to why that style might be preferred.

benson31:
@bvanessen @samadejacobs Can you guys take a minute or two to review this? It's been up for 10 days and it's not a huge diff. Thanks!

src/utils/lbann_library.cpp
```diff
@@ -48,193 +48,188 @@ std::shared_ptr<thread_pool> construct_io_thread_pool(lbann_comm *comm) {
       " (Limited to # Unused Compute Cores or 1)" << std::endl;
   }
 
-  std::shared_ptr<thread_pool> io_thread_pool = std::make_shared<thread_pool>();
+  auto io_thread_pool = make_unique<thread_pool>();
```
bvanessen: @benson31 shouldn't this be std::make_unique

bvanessen: Or is this our internal make_unique since we aren't c++14

benson31: It's `lbann::make_unique`, which is `std::make_unique` if that's available.

PR #892, opened 2019-02-20
GPU implementation of RMSprop

### LINKED ISSUE #341: GPU-implementation for RMSprop
Create GPU optimized implementation of RMSprop for GPU matrices.

timmoon10:
This closes #341. I see very close learning behavior between CPU and GPU when I run on LeNet with a small learning rate. RMSprop seems to diverge on both CPU and GPU on AlexNet, at least when I use the default settings in the model zoo.


bvanessen:
@JaeseungYeom Please try this with the siamese network.

PR #898, opened 2019-02-20
Sample list bug for too many open files

bvanessen:
Fixed a bug in in the sample list and JAG Conduit data reader so that
all files do not have to be opened at load time.  Instead, sample
files are dynamically opened and the file descriptors are cached.
Using a fixed sized unordered map to provide constant access time to
the file descriptor cache and random selection criteria for FD closure
on hash collision.

### DIFF
--- a/include/lbann/data_readers/data_reader_jag_conduit.hpp
+++ b/include/lbann/data_readers/data_reader_jag_conduit.hpp
@@ -59,7 +59,7 @@ class data_reader_jag_conduit : public generic_data_reader {
   using sample_locator_t = std::pair<std::string, hid_t>;
   using sample_map_t = std::vector<sample_locator_t>; ///< valid sample map type
   using sample_t = sample_list_jag::sample_t;
-  using sample_id_t = sample_list_jag::sample_id_t;
+  using sample_file_id_t = sample_list_jag::sample_file_id_t;
   /// linear transform on X defined as: first * X + second => X'
   using linear_transform_t = std::pair<double, double>;
 
--- a/include/lbann/data_readers/data_reader_jag_conduit.hpp
+++ b/include/lbann/data_readers/data_reader_jag_conduit.hpp
@@ -356,7 +356,7 @@ class data_reader_jag_conduit : public generic_data_reader {
   bool load_conduit_node(const size_t i, const std::string& key, conduit::Node& node) const;
   /// Check if a key exist for sample i
   bool has_conduit_path(const size_t i, const std::string& key) const;
-  void close_conduit_node(const size_t i);
+  //  void close_conduit_node(const size_t i);
 
   /// Obtain image data
   std::vector< std::vector<ch_t> > get_image_data(const size_t i, conduit::Node& sample) const;
--- a/include/lbann/data_readers/data_reader_jag_conduit.hpp
+++ b/include/lbann/data_readers/data_reader_jag_conduit.hpp
@@ -216,19 +216,19 @@ class data_reader_jag_conduit : public generic_data_reader {
   std::string get_description() const;
 
   /// Return the image simulation output of the i-th sample
-  std::vector<cv::Mat> get_cv_images(const size_t i, conduit::Node& sample);
+  std::vector<cv::Mat> get_cv_images(const size_t i, conduit::Node& sample) const;
 
   /**
    * Return the images of the i-th sample as an 1-D vector of lbann::DataType
    * There is one image per view, each of which is taken at closest to the bang time.
    */
-  std::vector<ch_t> get_images(const size_t i, conduit::Node& sample);
+  std::vector<ch_t> get_images(const size_t i, conduit::Node& sample) const;
 
   /// Return the scalar simulation output data of the i-th sample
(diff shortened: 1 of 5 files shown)


include/lbann/data_readers/data_reader_jag_conduit.hpp
```diff
@@ -216,19 +216,19 @@ class data_reader_jag_conduit : public generic_data_reader {
   std::string get_description() const;
 
   /// Return the image simulation output of the i-th sample
-  std::vector<cv::Mat> get_cv_images(const size_t i, conduit::Node& sample) const;
+  std::vector<cv::Mat> get_cv_images(const size_t i, conduit::Node& sample);
```
davidHysom: I know modern compilers *may* get around this, but returning a local object can result in a copy, which is not efficient. I'd like to see some discussion on this -- perhapsI'm behind the times c++ wise.

ndryden: A compiler _might_ be able to elide the copy here (and C++17 has specific requirements about copy elision being required, not as an optimization, but as part of the language semantics to avoid intermediate materialization of values). I'm not certain because the implementation of these methods is complicated enough, I'm not sure that the elision requirements would apply. @benson31 probably has more insight here.

bvanessen:
That is a good point.  I had to make the function non-const because we modify the state of the sample list object.  I will work to refactor that into something that preserves the const-ness of these functions.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

ndryden:
You can always make the function `const` and the internal state `mutable`. `const` is meant to be an indicator about changes to externally-visible state, not to say that the function cannot change internally. (Caches are a prototypical example.)

include/lbann/data_readers/data_reader_jag_conduit.hpp
benson31: It will never be a copy. In the best case, the copy is elided (I'd actually have to read the function to see if it fits the requirements for the C++17 guarantee). In the worst case, the vector is _moved_; in C++11+ it should never be copied. This "worst case" means that 3-4 pointer-sized things are copied (24-32 bytes), depending on the STL impl being used. See also my comments on #888.

benson31:
⬆️ This, do this.

bvanessen:
I have this working on both Pascal and Lassen with minimal performance impact.

include/lbann/data_readers/sample_list_jag.hpp
```diff
@@ -112,53 +98,144 @@ class sample_list_jag {
...
-    m_sample_id_map[id].first = filename;
+  void set_samples_filename(sample_file_id_t id, const std::string& filename) {
+    std::get<0>(m_file_id_stats_map[id]) = filename;
   }
 
-  void set_samples_hdf5_handle(sample_id_t id, hid_t h) {
-    m_sample_id_map[id].second = h;
+  void set_files_hdf5_handle(const std::string& filename, hid_t h) {
+    sample_file_id_t id = 0;
+    for (auto&& e : m_file_id_stats_map) {
```
JaeseungYeom: As a performance optimization, we can use another map from the filename to sample_file_id_t.
If maintaining two different data structures consistently is a burden, there is a container called boos::multi_index.

include/lbann/data_readers/sample_list_jag.hpp
```diff
@@ -179,30 +256,31 @@ class sample_list_jag {
...
- protected:
-
-  /// The number of partitions to divide samples into
-  size_t m_num_partitions;
+  static bool pq_cmp(fd_use_map_t left, fd_use_map_t right) {
+    return ((left.second).first < (right.second).first) ||
+           (((left.second).first == (right.second).first) &&
+            ((left.second).second < (right.second).second)); }
 
+ private:
```
JaeseungYeom: Why do we change from protected to private?

bvanessen:
@samadejacobs @JaeseungYeom @davidHysom Does this PR work for you?

include/lbann/data_readers/sample_list_jag.hpp
```diff
@@ -12,10 +12,18 @@
...
+#include <cereal/types/deque.hpp>
 #include <cereal/types/vector.hpp>
+#include <cereal/types/tuple.hpp>
 #include <cereal/types/string.hpp>
 #include <cereal/types/utility.hpp>
+#include "conduit/conduit_relay_io_hdf5.hpp"
+
+/// Number of system and other files that may be open during execution
+#define LBANN_MAX_OPEN_FILE_MARGIN 128
+#define LBANN_MAX_OPEN_FILE_RETRY 3
```
davidHysom: possibly have these as the defaults for two variables that could be overridden on the cmd line

bvanessen:
I agree that we should make these command line overridable.  Given the discussions about changing how we do options I was thinking that we should add that later.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

JaeseungYeom:
I ran this version successfully on pascal as I reported on slack channel.

PR #902, opened 2019-02-21
Correcting modulo arithmetic to correct for mini-batch size

bvanessen:
Fixed a bug where the mapping of samples to ranks in the data store
did not properly take into account the bracketing by the mini-batch
size.  When computing the original owners it is necessary to have the
original mini-batch size.  To compute the owner index first find its
position inside of the mini-batch (mod mini-batch size) and then find
how it is striped across the ranks in the trainer.  For the exchange
data routines, the sample target or owner is calculated modulo the
active mini-batch size.

### DIFF
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -691,7 +691,7 @@ class generic_data_reader : public lbann_image_preprocessor {
   }
 
   /// sets up a data_store.
-  virtual void setup_data_store(model *m);
+  virtual void setup_data_store(model *m, int mini_batch_size);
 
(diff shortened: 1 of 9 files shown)


bvanessen:
@davidHysom Note that this PR depends on #901

bvanessen:
We do support changing the mini-batch size during training, but I don't think that this will effect the data store adversely.  The original ownership is determined during the first epoch and once that is set even if you have a new mini-batch size that only effects the mapping for the exchange.  The original ownership should stay the same.  (Famous last words.)  One issue would be if the mini-batch size changed between setup and first epoch.

PR #903, opened 2019-02-22
Refactoring optimizers

timmoon10:
### Changes
- Gradient allreduces are performed in-place on the gradient itself. We no longer need to maintain a separate staging matrix.
- The status of the gradient allreduce is managed with an enum rather than a series of `bool`s. This makes reasoning much easier. We have four states: `ready`, `cleared` (gradient is considered zero, but hasn't been physically zeroed out), `allreduce_needed`, and `allreduce_started`.
- All member variables in the base optimizer class have been made private.
- Matrices are maintained in smart pointers (per Issue #155).
- Expanded documentation, especially in base optimizer class.
- Using uniform style across optimizers.

### Correctness tests

I compared learning behavior on AlexNet (1 Pascal node; 1% ImageNet-1K) against the head of `develop` (mostly 49bc1a9). I used the defaults in the model zoo, unless otherwise noted.

<details> <summary> SGD </summary>

optrefactor

</details>

<details> <summary> Adam </summary>

optrefactor

</details>

<details> <summary> AdaGrad </summary>

optrefactor

</details>

<details> <summary> RMSprop (learning rate 0.001) </summary>

optrefactor

</details>

<details> <summary> Hypergradient Adam (CPU) </summary>

optrefactor

</details>

It appears that this PR doesn't have a substantial impact on the learning behavior.

### Performance tests

Timing runs on ResNet-50 are in progress (16 Pascal nodes; ImageNet-1K; SGD). When I look at the timings from my correctness tests, I don't see a significant difference in run times.


samadejacobs:
Seems like objective function is going up with rmsprop at the end.....can we get two more data points (epochs 11 and 12) to be sure training is not diverging?

timmoon10:
<details> <summary> RMSprop </summary>

optrefactor

</details>

I think the code is working correctly. I suspect I'm using bad hyperparameters for RMSprop. I had to lower the learning rate to 0.001 to avoid divergence.

samadejacobs:
Sorry I wasnt clear, my interest was in whether the "develop" line in the previous plot would continue to go up or at what point does come down/starts converging?

timmoon10:
My timing tests on ResNet-50 (8 Pascal nodes; ImageNet-1K; SGD) are not encouraging (I got weird variance on 16 nodes).

<details> <summary> Epoch times </summary>

optrefactor

</details>

It looks a ~3% performance hit. I'll more into it.

timmoon10:
Rerunning the ResNet-50 experiments (8 Pascal nodes; ImageNet-1K; SGD) over the weekend got somewhat better results:
<details> <summary> Epoch times </summary>

optrefactor

</details>

The average performance degradation is only 1.1% and we have a 59% p-value that the epoch time distributions are identical (compare with the 11% p-value I got on Friday). I'm not sure what's up.

src/optimizers/optimizer.cpp
```diff
@@ -29,186 +29,177 @@
...
-  m_gradient_allreduce_finished = false;
-
-}
-
 void optimizer::add_to_gradient(const AbsDistMat& gradient,
-                                DataType scale) {
-  if (!is_initialized()) {
-    LBANN_ERROR("attempted to access gradients before they are set up");
+                                DataType scale,
+                                bool allreduce_needed) {
```
ndryden: I think your changes to this method result in us no longer doing non-blocking allreduces for gradient accumulation. All the layers with parameters call `add_to_gradient` with `allreduce_needed` as `true`, and so trigger an immediate blocking allreduce.

timmoon10: `allreduce_needed` doesn't trigger the allreduce immediately, but just let's the optimizer know that one is needed before returning the gradient in `get_gradient()`. Most of the time, the non-blocking allreduce still happens in the same place (immediately after gradient contributions have been received from the appropriate layers during backprop).

PR #908, opened 2019-02-27
Refactoring model class

timmoon10:
### Summary
This fixes some bugs in the constructor and destructor of the model class, so I can now copy AlexNet and immediately delete the copy. This expands on and supersedes #907.

In the process, I've done some major cleaning up in the model class. There's more to be done, but it seems like a reasonable point to consolidate.

### Major changes
- Bugfixes in copy constructors in conv and pooling layers (see PR #907).
- Better handling of optimizer in model copy constructor.
- Make all member variables in model class private.
- Consolidate multiple access functions for mini-batch step into `El::Int get_step(execution_mode)`.
- Remove redundant "current" from variable and function names.
- Remove "phase" logic from model class (this was used for greedy layer-wise autoencoders in ancient times).
- Expanded documentation.

### Warning
This will break backward compatibility with old checkpointed models.

### DIFF
--- a/include/lbann/models/model.hpp
+++ b/include/lbann/models/model.hpp
@@ -53,6 +53,10 @@ class lbann_callback;
 class model {
 public:
 
+  // ===========================================
+  // Life cycle functions
+  // ===========================================
+
   model(lbann_comm *comm,
         El::Int mini_batch_size,
         objective_function *obj_fn,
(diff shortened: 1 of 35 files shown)


benson31:
Superficially this looks like a great first step in this effort (especially the `private` data stuff! We should make it universal! ;) ). I'll test it out before the meeting.

include/lbann/models/model.hpp
```diff
@@ -49,71 +49,54 @@ namespace lbann {
...
 
-  model(lbann_comm *comm,
+  // ===========================================
+  // Life cycle functions
+  // ===========================================
+
+  model(lbann_comm* comm,
         El::Int mini_batch_size,
-        objective_function *obj_fn,
+        objective_function* obj_fn,
```
ndryden: If we're gonna move the `*` around here, we should be consistent (at least in this file).

PR #910, opened 2019-02-28
Refactoring convolution layer

timmoon10:
This PR makes some bugfixes so that a model can be setup multiple times. This is helpful for MTNN, where a model might be copied, modified, and re-setup. In conjunction with #908, I can now setup AlexNet multiple times, copy it, setup the copy multiple times, and train the original model without incident (I can't yet run the copied model).

Major changes:
- Bugfixes in conv and pooling layer copy constructors (PR #907).
- Convolution kernel dimensions are now constructed with a virtual function, not stored in a variable.
- Minor stylistic changes in convolution layer.
- Check convolution parameters during setup.

Pinging @naoyam to minimize merge conflicts with his distconv work.

### DIFF
--- a/include/lbann/layers/learning/base_convolution.hpp
+++ b/include/lbann/layers/learning/base_convolution.hpp
@@ -63,7 +63,7 @@ protected:
    *  convolution. The default convolution operation has one group, and a
    *  depthwise convolution has as many groups as there are input channels.
    */
-  int m_num_groups;
+  int m_groups;
 
   /** Scaling factor for bias term.
    *  If the scaling factor is zero, bias is not applied.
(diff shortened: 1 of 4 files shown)


include/lbann/layers/learning/base_convolution.hpp
```diff
@@ -74,101 +74,87 @@ class base_convolution_layer : public learning_layer {
...
-      m_kernel_size(0),
-      m_pads(pads),
-      m_strides(strides),
-      m_dilations(dilations),
-      m_num_groups(groups),
-      m_bias_scaling_factor(has_bias ? DataType(1) : DataType(0)),
-      m_kernel_gradient(this->m_comm->get_trainer_grid()),
-      m_bias_gradient(this->m_comm->get_trainer_grid())
+    : Layer(comm),
+      m_output_channels(std::max(output_channels, 1)),
```
naoyam: Why doing max with 1 here?

timmoon10: It's a way to handle bad input. Although now that you mention it, it's probably better to throw an exception.

naoyam: Erroneous input should be exposed rather than hidden.

PR #913, opened 2019-03-01
Add a driver that just prints help and exits

benson31:
This patches around the fact that the help logic requires MPI and CUDA
to be initialized.

The original signature is still there because I didn't want to chase around all the tests/frontends/etc. This guarantees no breakage of existing code.

Don't call `lbann-help` with an MPI launcher. If you do, you might get multiple output. `MPI_Init` is never called; there is absolutely no logic in this driver. It stupidly prints the help to `stderr` and exits.


bvanessen:
Was there something wrong with just calling lbann —help?

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

benson31:
Yes.

PR #918, opened 2019-03-06
Miscellaneous Python development for VRAM

timmoon10:
My apologies for such an unfocused PR. Changes:
- LSTM module.
- Prototext files must now be generated by constructing a `Model` class and calling `save_proto`.
- Renamed `concatenation_axis` and `slice_axis` fields to just `axis`.
- Refactored viz script.
  - Fixed a bug where layers with multiple children weren't properly displayed.
  - All layers are displayed the same. It was previously a problem because layers with no children only displayed the layer name. There is no more color scheme support.
  - Replaced custom parsers for color schemes, protobuf messages, and command-line arguments.
  - Expanded documentation. It's very convenient that argparse will autogenerate help messages for Python scripts.


oyamay:
Could you remove `"scripts/viz/lbviz"` from [`setup.py`](https://github.com/timmoon10/lbann/blob/vram-dev/scripts/proto/setup.py#L28)?
Perhaps adding `"scripts/viz.py"` is a good idea, but I think if we do so we should rename it since the name may conflict with other scripts.

PR #919, opened 2019-03-06
Templated sample list

JaeseungYeom:
templatize the `sample_list_jag` for `file_handle_t` and `sample_name_t`.
- name change from `sample_list_jag` to `sample_list`
- This is based off PR #898.
- It compiles with gcc 4.9.3 and clang 6.0.0.
- tested with ltfb, and confirmed that the behavior matches that of the head of devel.
- While most of the methods managing `file_handle_t`-type file handles remain in the base class, the actual file accesses using the file handles are wrapped into virtual functions. For instance, `sample_list_hdf5` inherits `sample_list_open_files` implementing such virtual functions that rely on the hdf5 and the conduit library. 
- sample_list_conduit_io_handle.hpp has been added


JaeseungYeom:
I pull from the latest develop to use the latest conduit. Since then, there have been compilation errors coming from callbacks.

davidHysom:
I had a build error after a pull this morning, in callback_dump_outputs.hpp
I commented out some defaults in the ctor so I could compile -- but this needs looking at. I didn't see anything in 'git log' that indicated it had been recently changed.

benson31:
@davidHysom @JaeseungYeom That's my mistake -- sorry! Overambitious find-and-replace. It's fixed in commit 66f29e25b.

davidHysom:
The following will not work with hydra data (wrt the current schema):
 conduit::relay::io::hdf5_group_list_child_names(hdf5_file_hnd, "/", sample_names);
 if(sample_names.size() != (included_samples + excluded_samples)) {
  //ERROR

The reason is that the hydra conduit files have a top-level child that is not a sample -- it's some sort of childless metadata

bvanessen:
> The following will not work with hydra data (wrt the current schema):
> conduit::relay::io::hdf5_group_list_child_names(hdf5_file_hnd, "/", sample_names);
> if(sample_names.size() != (included_samples + excluded_samples)) {
> //ERROR
> 
> The reason is that the hydra conduit files have a top-level child that is not a sample -- it's some sort of childless metadata

Can we get them to put in a sample name into the conduit file.

bvanessen:
@JaeseungYeom Where are we at with this PR.  It is still marked as in progress?  Is it ready for review?

davidHysom:
I added a comment to the PR yesterday concerning how hyda conduit data will break (cause an exception to be thrown). One workaround would be to get John to get rid of the top-level child that is not a sample_id. The other workaround would be to add a flag to the templated sample list, e.g, if (opts->get_bool("hydra")) <do this> else <do this>.

Below is John's mail concerning the metadata field; I asked what it was, I thought it was a unique file identifier:

"No, 'METAf65b2483bd13f9af2017359e1f9b9069' is a reserved key whose value contains metadata that is needed for decoding certain kinds of data. If you know what the data is that you are looking for, you can safely ignore those fields. It is necessary to convert the conduit file back to a python pickle if the need for that were ever to arise."

JaeseungYeom:
> @JaeseungYeom Where are we at with this PR. It is still marked as in progress? Is it ready for review?

ready to review

JaeseungYeom:
If I have a better understanding of how hydra data file is structured, I might be able to derive a class that override functions specific to that.

bvanessen:
Have you tested it with the JAG Conduit data reader and verified that it reproduces the same results?

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

davidHysom:
I gave you the schema for the hydra conduit data.
I also posted a simplified version (without the troublesome META field) on slack, May 24.

It seems the sample_list classes are geared towards conduit.
Is there a PR planned to extend them to, e.g, imagenet?

davidHysom:
There's a long email thread between myself and John concerning hydra.
I just realized you weren't CCd on it. I'll forward it to you when I get into the lab this morning.

bvanessen:
I still need to review this PR, but I am fine with the first cut being JAG specific.  Once we merge this, we can extend it for ImageNet.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

JaeseungYeom:
> Have you tested it with the JAG Conduit data reader and verified that it reproduces the same results? 
Yes.

JaeseungYeom:
I am not sure if you guys had a chance to review this. The sample_list extends to the multiple inheritance hierarchy to reuse the code as much as possible and support the data specific IO operations. The base class `sample_list` being the most generic can be applied to imagenet and has no component referencing conduit. In fact, it does not manage any open file handle. `sample_list_open_files` inherits the base class, which itself is an abstract class. It currently has two derived classes: `sample_list_hdf5` and `sample_list_conduit_io_handle`. The former is the one used by `data_reader_jag_conduit` as seen in the line "using sample_list_t = sample_list_hdf5<sample_name_t>." The latter could have been used. However, currently the data store plugged into the data reader prevents it as it hard-codes the specific conduit interface without template parameterization.

Issue #912, opened 2019-03-01
Segfault when training a copied model


timmoon10:
When PR #908 is merged, we'll be able to copy-construct models and call setup multiple time on both the originals and copies. However, I run into an error when I try training the copy.

<details> <summary> Modification to model_zoo/lbann.cpp </summary>

```c++
    auto model = build_model_from_prototext(argc, argv, pb,
                                            comm.get(), io_thread_pool, true);
    auto copy = std::unique_ptr<lbann::model>(model->copy());
    copy->setup(io_thread_pool);
    model.swap(copy);
```
</details>

<details> <summary> Error message when running AlexNet on a Pascal node </summary>

```
****************************************************************
Caught signal 11 (SIGSEGV - invalid memory reference) on rank 1
Stack trace:
   0: lbann::stack_trace::get()
   1: lbann::exception::exception(std::string, bool)
   2: /usr/WS1/moon13/src/lbann/build/gnu.Release.pascal.llnl.gov/lbann/build/liblbann.so(+0xdbdad4) [0x2aaaabc90ad4] (could not find stack frame symbol)
   3: /lib64/libpthread.so.0(+0xf5d0) [0x2aaac7a975d0] (could not find stack frame symbol)
   4: lbann::partitioned_io_buffer::num_samples_ready(lbann::execution_mode)
   5: lbann::generic_input_layer::fp_compute()
   6: lbann::Layer::forward_prop()
   7: lbann::model::forward_prop(lbann::execution_mode)
   8: lbann::model::train_mini_batch()
   9: lbann::model::train(int, int)
  10: /usr/WS1/moon13/src/lbann/build/gnu.Release.pascal.llnl.gov/lbann/build/model_zoo/lbann() [0x40aeab] (could not find stack frame symbol)
  11: __libc_start_main (demangling failed)
  12: /usr/WS1/moon13/src/lbann/build/gnu.Release.pascal.llnl.gov/lbann/build/model_zoo/lbann() [0x40b674] (could not find stack frame symbol)
****************************************************************
```

</details>

bvanessen:
Closed by #920

timmoon10:
I confirm that my modified version of model_zoo/lbann.cpp runs fine now.

PR #920, opened 2019-03-06
Fixed a bug in the copy function of the partitioned_io_buffer. 

bvanessen:
Also added a proper destructor.  This is working on addressing #912 .


benson31:
@timmoon10 Can you see if this works with your test that doesn't have all my code's baggage going along with it...

benson31:
In my tests, this has resolved all memory issues surrounding the io buffers.

PR #924, opened 2019-03-08
Callback augmentation

samadejacobs:
Add batch interval at test time to callback.
Add min and max global statistics to print (metric) callback


src/models/model.cpp
```diff
@@ -1256,7 +1256,9 @@ void model::do_batch_begin_cbs(execution_mode mode) {
       break;
     case execution_mode::validation:
     case execution_mode::testing:
-      cb->on_batch_evaluate_begin(this);
+      if (get_step() % cb->get_batch_interval() == 0) {
```
timmoon10: I agree with making this the default behavior, but I worry it might introduce bugs in our callbacks. We should make this change in a separate PR.

src/callbacks/callback_print.cpp
```diff
@@ -192,24 +194,36 @@ void lbann_callback_print::report_results(model *m) {
...
+                    << min_score << met->get_unit()
+                    << std::endl;
+          std::cout << m->get_name() << " (global average) "  << mode_string << " "
                     << met->name() << " : "
                     << avg_score << met->get_unit()
                     << std::endl;
+          std::cout << m->get_name() << " (global max) "  << mode_string << " "
+                    << met->name() << " : "
+                    << max_score << met->get_unit()
+                    << std::endl;
```
timmoon10: Can we change the order to print mean, then min, then max? It's more intuitive to have the min and max next to each other.

src/models/model.cpp
samadejacobs: Will make the a separate PR, may be it is better implemented in the derived rather than base class.

src/callbacks/callback_print.cpp
samadejacobs: Done.

src/models/model.cpp
timmoon10: I think we have two good options:
- Keep `m_batch_interval` in the base class and make it a private variable. (This is my preference).
- Remove `m_batch_interval` from the base class and reimplement it in every derived class that needs it.

PR #927, opened 2019-03-12
Python data reader

timmoon10:
### Description
This implements a simple, non-performant data reader that is intended to interface with an embedded Python session. The user should provide a Python script that defines three functions:
- Take a sample index and return the corresponding data sample (as an iterable object containing floats).
- Return the number of data samples.
- Return the dimensions of a data sample.

This matches PyTorch's `Dataset` framework very well (it's a bit different from TensorFlow's `Dataset` though). There is currently no support for parallelism (we set a mutex to serialize all Python calls) and performance is rather bad (~100x slowdown when I run LeNet on MNIST).

### Building

This requires building LBANN with the Python/C API for Python 3.6 or later. I've made changes in the CMake and build_lbann_lc.sh, but these are currently optional. Pass the `--python` flag into build_lbann_lc.sh to use this PR.

### Comparison with PR #240
From what I gather, PR #240 was intended to support streaming data. To that end, it set up a Python server in the background to continually process data and maintained a socket connection for communication. This PR is simpler and matches our current data reader framework more closely. When loading a data sample, we just run a Python function instead of loading a file.

### Example

I was able to train LeNet on MNIST with the following (in conjunction with PR #926):

<details> <summary> Data reader Python script </summary>

```python3
from torch import flatten
from torchvision.datasets import MNIST
from torchvision.transforms import ToTensor

# Construct MNIST dataset
data_dir = '/g/g17/moon13/temp'
dataset = MNIST(data_dir, train=True, download=True, transform=ToTensor())

def get_sample(index):
    image, label = dataset[index]
    data = []
    for val in flatten(image):
	data.append(float(val))
    for ind in range(10):
	data.append(1 if ind == label else 0)
    return data

def num_samples():
    return len(dataset)

def sample_dims():
    return (28*28 + 10,)
```

</details>

<details> <summary> Data reader prototext </summary>

```
data_reader {
  reader {
    name: "python"
    role: "train"
    shuffle: true
    validation_percent: 0.1
    absolute_sample_count: 0
    percent_of_data_to_use: 1.0
    python {
      script: "/g/g17/moon13/lbann/experiments/lenet/dataset.py"
      sample_function: "get_sample"
      num_samples_function: "num_samples"
      sample_dims_function: "sample_dims"
    }
  }
  reader {
    name: "python"
    role: "test"
    shuffle: true
    absolute_sample_count: 0
    percent_of_data_to_use: 1.0
    python {
      script: "/g/g17/moon13/lbann/experiments/lenet/dataset.py"
      sample_function: "get_sample"
      num_samples_function: "num_samples"
      sample_dims_function: "sample_dims"
    }
  }
}
```

</details>

<details> <summary> Model </summary>

```python3
import lbann.proto as lp
from lbann.models.lenet import LeNet

input = lp.Slice(lp.Input(), slice_points='0 784 794')
images = lp.Reshape(input, dims='1 28 28')
labels = lp.Identity(input)
preds = LeNet(10)(images)
softmax = lp.Softmax(preds)
ce = lp.CrossEntropy([softmax, labels])
acc = lp.CategoricalAccuracy([softmax, labels])

l2_reg = lp.L2WeightRegularization(scale=1e-4)
obj = lp.ObjectiveFunction([ce, l2_reg])

metrics = [lp.Metric(acc, name='categorical accuracy', unit='%')]
callbacks = [lp.CallbackPrint(), lp.CallbackTimer()]
model = lp.Model(256, 100,
                 layers=input, objective_function=obj,
                 metrics=metrics, callbacks=callbacks)
model.save_proto("model.prototext")
```

</details>

The learning behavior on a Pascal node was very close to PR #926, but an epoch  takes ~73s instead of 0.75s.


timmoon10:
It turns out my data reader script was too naive. I only got a ~8x slowdown with the following script:

<details> <summary> Data reader Python script </summary>

```python
from torch import flatten
from torchvision.datasets import MNIST
from torchvision.transforms import ToTensor

# Construct MNIST dataset
data_dir = '/g/g17/moon13/temp'
dataset = MNIST(data_dir, train=True, download=True, transform=ToTensor())

def get_sample(index):
    image, label = dataset[index]
    data = []
    for channel in image.tolist():
        for row in channel:
            for val in row:
		yield val
    for ind in range(10):
	yield 1 if ind == label else 0

def num_samples():
    return len(dataset)

def sample_dims():
    return (28*28 + 10,)
```

</details>

I think parallel access to the Python script would further improve the performance degradation.

PR #938, opened 2019-03-15
Fixed a cmake file that didn't track a file's move

bvanessen:


docs/CMakeLists.txt
```diff
@@ -78,7 +78,7 @@ endif ()
 if (DOXYGEN_FOUND)
   set(DOXYGEN_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in
-    ${CMAKE_CURRENT_SOURCE_DIR}/README.md
+    ${CMAKE_CURRENT_SOURCE_DIR}/README
```
benson31: Oh drat. That's not even the right file in the first place.

PR #941, opened 2019-03-16
Python function to launch Slurm jobs

timmoon10:
This PR adds the `lc` submodule to the Python interface, which contains numerous defaults and convenience functions for LC systems. It is essentially a Python port of `experiments/run_lbann_experiment.sh`. It provides a function called `lbann.lc.run` that takes in a model, data reader, and optimizer; sets up an experiment directory at `experiments/<timestamp>_<job name>` with a prototext file and a batch script; and submits the batch script. If the function detects a job allocation from salloc or sxterm, the experiment is just run directly with the allocated nodes. This allows for standalone Python scripts that can setup a model, configure an experiment, and launch it. To see this workflow in action, look at the examples in `model_zoo/vision`.


model_zoo/vision/alexnet.py
```diff
@@ -36,6 +37,31 @@
     default=data_reader_prototext, type=str,
     help='data reader prototext file (default: ' + data_reader_prototext + ')',
     metavar='FILE')
+parser.add_argument(
```
bvanessen: I really like where this is going, but these parser arguments need to be moved to a common class that is shared by all models, but can also be extended in a model specific way.  Otherwise we will get the same type of duplication that we seen between this and the mnist example.

timmoon10:
@ndryden I've implemented all of the optimizations from `experiments/run_lbann_experiment.sh` (including `--cpu_bind=mask_cpu:0x000001ff,0x0003fe00` for Pascal). Are there any other tricks that you've found useful?

I'm not quite happy with the code structure for this functionality. We should separate out the LC-specific optimizations from the generic job launching functionality. I have some ideas.

ndryden:
Looks good to me!

samadejacobs:
I like this too! Can it be extended such that the prototext file(s) created in the experiments/<timestamp>_<job name> is the file submitted for execution?

timmoon10:
@samadejacobs I believe this is doing what you're asking for. This is basically a Python reimplementation of `experiments/run_lbann_experiment.sh`.

timmoon10:
I've reorganized the functionality to separate the job launcher and the LC-specific stuff. There's a generic job launcher in `lbann.launcher` and a wrapper with LC defaults and optimizations in `lbann.contrib.lc.launcher`.

PR #942, opened 2019-03-16
adding get_random_node() method

davidHysom:
adding functionality to deal with failures in data_store_jag_conduit when reading conduit nodes from file.

### DIFF
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -43,6 +43,7 @@
 #include <string>
 #include <vector>
 #include <unistd.h>
+#include <unordered_set>
 
 
 #define NOT_IMPLEMENTED(n) { \
@@ -712,6 +713,9 @@ class generic_data_reader : public lbann_image_preprocessor {
   virtual void post_update() {}
 
  protected:
+
+  // For use with conduit when samples are corrupt.
+  std::unordered_set<int> m_using_random_node;
 
   /**
    * Return the absolute number of data samples that will be used for training
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -715,7 +715,7 @@ class generic_data_reader : public lbann_image_preprocessor {
  protected:
 
   // For use with conduit when samples are corrupt.
-  std::unordered_set<int> m_using_random_node;
+  mutable std::unordered_set<int> m_using_random_node;
 
   /**
    * Return the absolute number of data samples that will be used for training
--- a/src/data_readers/data_reader_jag_conduit.cpp
+++ b/src/data_readers/data_reader_jag_conduit.cpp
@@ -1291,11 +1291,11 @@ std::vector<data_reader_jag_conduit::input_t> data_reader_jag_conduit::get_input
         } else {
           sample = n_input;
         }
-        const input_t val_raw = static_cast<input_t>(sample[conduit_obj].value());
-        const input_t val = static_cast<input_t>(val_raw * tr->first + tr->second);
-        inputs.push_back(val);
-        tr ++;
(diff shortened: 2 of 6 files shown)


src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -1248,13 +1285,13 @@ std::vector<data_reader_jag_conduit::input_t> data_reader_jag_conduit::get_input
...
+        if (from_file) {
+          sample[conduit_obj].set(n_input);
+        } else {
+          sample = n_input;
+        }
       }
-      const input_t val_raw = static_cast<input_t>(sample[conduit_obj].value());
-      const input_t val = static_cast<input_t>(val_raw * tr->first + tr->second);
-      inputs.push_back(val);
-      tr ++;
```
bvanessen: @davidHysom Is there a reason that this code was removed?

davidHysom:
My mistake; thanks for catching that; should be fixed now

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -1248,13 +1285,17 @@ std::vector<data_reader_jag_conduit::input_t> data_reader_jag_conduit::get_input
...
+        bool from_file = load_conduit_node(sample_id, conduit_field, n_input);
+        if (from_file) {
+          sample[conduit_obj].set(n_input);
+        } else {
+          sample = n_input;
+        }
+        const input_t val_raw = static_cast<input_t>(sample[conduit_obj].value());
+        const input_t val = static_cast<input_t>(val_raw * tr->first + tr->second);
+        inputs.push_back(val);
+        tr ++;
```
bvanessen: This has been put back in at the wrong level of the if-block.  It needs to be outside of the "if(sample[conduit_obj].schema().dtype().is_empty()) {" conditional so that it will be used when grabbing samples from the data store as well.

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -1403,6 +1448,7 @@ bool data_reader_jag_conduit::fetch_datum(CPUMat& X, int data_id, int mb_idx) {
   }
 
   m_sample_list.close_if_done_samples_hdf5_handle(data_id);
+  m_using_random_node.erase(m_io_thread_pool->get_local_thread_id());
```
bvanessen: I don't see where you ever place the thread local id into the unordered set.

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -310,17 +310,46 @@ const conduit::Node& data_reader_jag_conduit::get_conduit_node(const conduit::No
...
-                " for sample "+ sample_name);
-    return false;
+    if (m_data_store != nullptr) {
+      const conduit::Node obj = m_jag_store->get_random_node();
+      node = obj["data"];
+      const std::vector<std::string>& child_names = node.child_names();
+      const std::string cur_child = child_names[0];
+      const std::string new_child = pad(std::to_string(i), SAMPLE_ID_PAD, '0');
+      node.rename_child(cur_child, new_child);
+      const std::string& file_name = m_sample_list.get_samples_filename(id);
```
bvanessen: I think that you need something like `m_using_random_node.emplace(m_io_thread_pool->get_local_thread_id());`

include/lbann/data_readers/data_reader.hpp
```diff
@@ -713,6 +714,9 @@ class generic_data_reader : public lbann_image_preprocessor {
 
  protected:
 
+  // For use with conduit when samples are corrupt.
+  std::unordered_set<int> m_using_random_node;
```
bvanessen: To update the load_conduit_node function we need to make this mutable.

PR #962, opened 2019-03-22
Tweaks to Python scripts in model zoo

timmoon10:
Changes:
- The model zoo LeNet script has been simplified so users can use it as a guide for their own scripts. It builds the layer graph from scratch and only has two options (for the LC account and partition).
- Added Python script for Resnets to the model zoo.
- The `run` function has been added to the base `lbann` namespace. LLNL users should still use `lbann.contrib.lc.launcher.run`.
- Added some utility functions in `lbann.contrib.args` to create an optimizer from command-line arguments.


model_zoo/vision/alexnet.py
```diff
@@ -2,16 +2,12 @@
 import argparse
 from os.path import join
 import google.protobuf.text_format as txtf
-import lbann
-from lbann.util import lbann_dir
-import lbann.models
-import lbann.proto
-import lbann.contrib.args
+import lbann, lbann.models, lbann.proto, lbann.contrib.args
```
ndryden: Minor thing, but [PEP8 recommends imports be on separate lines](https://www.python.org/dev/peps/pep-0008/#imports).

PR #963, opened 2019-03-23
Data store preload

davidHysom:
in progress; likely does not even compile as of this writing.


bvanessen:
@davidHysom Yes this is exactly what I was thinking about.  Thanks for getting started on it.

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -353,26 +358,24 @@ bool data_reader_jag_conduit::load_conduit_node(const size_t i, const std::strin
...
 bool data_reader_jag_conduit::has_conduit_path(const size_t i, const std::string& key) const {
   const sample_t& s = m_sample_list[i];
   sample_file_id_t id = s.first;
-  const std::string& file_name = m_sample_list.get_samples_filename(id);
   const std::string& sample_name = s.second;
-  const hid_t h = m_sample_list.get_samples_hdf5_handle(id);
-
-  const std::string path = sample_name + key;
-  if (h <= static_cast<hid_t>(0) || !conduit::relay::io::hdf5_has_path(h, path)) {
+  const sample_list_jag::io_t h = m_sample_list.get_samples_handle(id);
```
bvanessen: Dave is this PR mixing in the new conduit I/O node work?

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -353,26 +358,24 @@ bool data_reader_jag_conduit::load_conduit_node(const size_t i, const std::strin
     }
   }
 
-  conduit::relay::io::hdf5_read(h, path, node);
+  h.read(path, node);
```
bvanessen: Are we mixing in the new PR about conduit I/O nodes?  If so then line 324 probably needs to change.

davidHysom: no, it's using the old/existing hdf5 interface, since the PR for IOHandle isn't ready. I had some problems with it, and stopped working on in until the preload code is finished -- because I thought it was most urgent to get that working.

bvanessen: Okay, then I am confused by these lines of code that seem to be using the newer IOHandle?

src/proto/proto_common.cpp
```diff
@@ -401,6 +401,14 @@ void init_data_readers(
 
     reader->set_master(master);
 
+    //  under development; experimental
+    if (opts->has_bool("use_data_store") && opts->get_bool("use_data_store")) {
```
davidHysom: You only need opts->get_bool here; it "use_data_store" isn't in the map, false is returned by default

davidHysom:
Not perfect, but seem to be working. I think it's time to merge this PR (so I am) and open subsequent PRs to deal with improvements.

PR #974, opened 2019-03-28
fix "failed to find data_id" errors

davidHysom:
spurious error; depends on mb size and processor count in mysterious ways.


davidHysom:
This has been overtaken by PR 975. This PR served as a sort of test-bed for tracking down bugs.

PR #977, opened 2019-04-03
Wide ResNet

timmoon10:
This implements `WRN-50-2-bottleneck`, as described in the Wide ResNet [paper](https://arxiv.org/abs/1605.07146) and [repo](https://github.com/szagoruyko/wide-residual-networks/blob/master/pretrained/README.md).

`WRN-50-2-bottleneck` is so similar to ResNet-50 that we could implement it with some generalization in the ResNet module. We should go that route if we need other Wide ResNet variants.

### DIFF
--- a/model_zoo/vision/resnet.py
+++ b/model_zoo/vision/resnet.py
@@ -8,7 +8,7 @@ import lbann.proto
 import lbann.contrib.args
 
 # Command-line arguments
-desc = ('Construct and run AlexNet on MNIST data. '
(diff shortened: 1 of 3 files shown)


timmoon10:
Training on 4 Pascal nodes, I get 0.37 sec/step, compared with 0.22 sec/step for ResNet-50.

<details> <summary> Learning curve on first 10 classes of ImageNet-1K. </summary>

wrn-50-2

</details>

PR #979, opened 2019-04-05
LTFB sendrecv_weights algorithm will communicate optimizer hyperparameters

timmoon10:
This is similar to PR #822. Only SGD and Adam are supported. This work should be deprecated once we have a performant implementation using checkpointing.


src/callbacks/callback_ltfb.cpp
```diff
@@ -126,6 +127,19 @@ void exchange_models__sendrecv_weights(lbann_comm& comm,
       const auto* send_sgd = dynamic_cast<const sgd*>(send_opt);
       auto* recv_sgd = dynamic_cast<sgd*>(recv_opt);
       if (send_sgd != nullptr && recv_sgd != nullptr) {
+        using hyperparameters_type = std::tuple<DataType, DataType, bool>;
+        hyperparameters_type hyperparameters(send_sgd->get_learning_rate(),
+                                             send_sgd->get_momentum(),
+                                             send_sgd->using_nesterov());
+        El::mpi::SendRecv(reinterpret_cast<El::byte*>(&hyperparameters),
```
bvanessen: Why are we using raw El::mpi calls here not lbann_comm

PR #981, opened 2019-04-06
Split data store

bvanessen:
This feature allows you to split the data store in the JAG data reader rather than to duplicated and then clean up the data store.  It avoids having to have 2x the memory to hold the data store during the copy construction and subsequent cleanup.  I also added a synchronization point in the LTFB trainer to make sure that all trainers have completed data prep before the first epoch starts.


src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -911,23 +920,31 @@ void data_reader_jag_conduit::preload_data_store() {
   /// @todo BVE FIXME this
   m_rank_in_model = get_comm()->get_rank_in_trainer();
 
+  options *opts = options::get();
   double tm1 = get_time();
-  if (get_comm()->am_trainer_master()) {
-    std::cout << "data_store_jag::preload_data_store() for role: " << get_role() << " starting preload\n";
+  if (get_comm()->am_world_master() ||
```
davidHysom: Why not "is_master()" - is that different from world master?

bvanessen:
Yes, is_master is a boolean that is set to the trainer master value.  World master is the unique rank 0 of the entire LBANN ensemble.  This prevents every trainer from printing the same information.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #982, opened 2019-04-09
Mem check

davidHysom:
adding function in check_mem_capacity() in utils/lbann_library to report on estimated memory usage needed to load JAG samples. Compiles, but execution is not yet tested.


src/utils/lbann_library.cpp
```diff
@@ -317,4 +318,149 @@ void print_lbann_configuration(lbann_data::Model *pb_model, lbann_comm *comm, in
   std::cout << std::endl;
 }
 
+void check_mem_capacity(lbann_comm *comm, const std::string sample_list_file, size_t stride, size_t offset) {
```
naoyam: Why is this function defined here? Isn't it specific to JAG?

src/utils/lbann_library.cpp
```diff
@@ -26,6 +26,7 @@
 
 #include "lbann/utils/lbann_library.hpp"
 #include "lbann/callbacks/callback_checkpoint.hpp"
+#include "lbann/data_store/data_store_jag.hpp"
```
naoyam: Why do you need to load a specific data reader in lbann_library?

src/utils/lbann_library.cpp
davidHysom: It's specific to conduit, and all data_stores for all readers will be using conduit in the future, so I thought it was general enough to put here. It probably makes more sense to put it in the data_store_jag (which will be renamed data_store_conduit) class.

davidHysom:
merging now, because I need this to track down a bug for Sam

PR #983, opened 2019-04-09
script that adds sample IDs from one list to others. Requested by Sam…

davidHysom:
…. Only works with pre-existing INCLUSIVE sample lists.


davidHysom:
I'm merging because no one has reviewed, and this is pure python code, so will not break any lbann core  c++ functionality.

PR #984, opened 2019-04-09
combine generic_data_store and data_store_jag into a single entity

davidHysom:
initial changes; does not compile; do not even think about looking at the code at this point


timmoon10:
If we have only one data store, why not call it `data_store` instead of `data_store_conduit`?

davidHysom:
well, it's entirely based on conduit, so I thought that was descriptive. But I don't have stron feelings on the matter.

PR #985, opened 2019-04-10
hack to avoid triggering "duplicate ID" in data_store::set_conduit_no…

davidHysom:
…de() when in validation mode. This is needed for LTFB when --use_data_store since validation is run multiple times in an epoch.

THIS PR SHOULD NOT BE MERGED WITH DEVELOP -- because a more comprehensive fix will be implemented in the future.


davidHysom:
closing, since a subsequent PR, that has been merged, addressed this issue.

PR #988, opened 2019-04-12
Use C++ thread_private for RNG state

### LINKED ISSUE #900: I/O threads don't use threadprivate RNGs
Since we no longer use OpenMP, our old threadprivate RNG states don't work.

ndryden:
This removes all the OpenMP from the RNG generators. Closes #900.

To reduce overhead, I would eventually like to remove the lazy-init and do it in a per-thread initialization when thread pools are created.

This removes the code previously used to checkpoint/restore the RNG state, since it was using OpenMP. I'll make a new issue to track this once this issue is merged.

### DIFF
--- a/src/utils/random.cpp
+++ b/src/utils/random.cpp
@@ -24,49 +24,51 @@
 // permissions and limitations under the license.
 ////////////////////////////////////////////////////////////////////////////////
 
-#include <omp.h>
 #include "lbann/utils/random.hpp"
+#include "lbann/utils/exception.hpp"
 #include "lbann/io/file_io.hpp"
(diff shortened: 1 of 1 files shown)


ndryden:
We use these RNGs for data preprocessing too. 

Unfortunately, looking more closely, we use the same RNGs in e.g. dropout on the CPU, which of course does not work. That's a bigger problem.

ndryden:
I should add, `thread_local` should interoperate with OpenMP `threadprivate`, at least for GCC.

ndryden:
Replaced by #990.

PR #990, opened 2019-04-13
Bugfix for thread safe random number generators

### LINKED ISSUE #900: I/O threads don't use threadprivate RNGs

bvanessen:
Updated the data sequence random number generator to use std thread 
local storage.  Added random number generators for the I/O threads
that are thread safe for std threads.  Closes #900.

### DIFF
--- a/include/lbann/utils/random.hpp
+++ b/include/lbann/utils/random.hpp
@@ -59,6 +59,21 @@ fast_rng_gen& get_fast_generator();
  * @note If compiling with OpenMP, this is stored in a threadprivate variable.
  */
 rng_gen& get_data_seq_generator();
+
+/**
+ * Return a reference to the global LBANN random number generator used
+ * for shuffling the data samples within each mini-batch
+ * @note This is stored in a threadprivate variable.
+ */
+rng_gen& get_io_generator();
+
+/**
+ * Return a reference to the fast global LBANN random number generator used
+ * for the I/O threads
+ * @note This is stored in a threadprivate variable.
+ */
+fast_rng_gen& get_fast_io_generator();
+
 /**
  * Return random integers uniformly distributed in [0, max).
  * @param g C++ uniform random bit generator.
@@ -98,9 +113,6 @@ inline T fast_rand_int_pow2(Generator& g, T max) {
  *  @param comm If present, mixes the process's rank within the model
  *              into the seed; if not, uses the MPI world rank.
  *
- *  @todo Support saving/restoring the generator's state. This is directly
- *  supported via the >> and << operators on the generator (reading/writing
- *  from/to a stream).
  */
 void init_random(int seed = -1, lbann_comm *comm = nullptr);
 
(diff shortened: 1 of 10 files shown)


include/lbann/utils/random.hpp
```diff
@@ -59,6 +59,21 @@ fast_rng_gen& get_fast_generator();
  * @note If compiling with OpenMP, this is stored in a threadprivate variable.
  */
 rng_gen& get_data_seq_generator();
+
+/**
+ * Return a reference to the global LBANN random number generator used
+ * for shuffling the data samples within each mini-batch
+ * @note This is stored in a threadprivate variable.
```
ndryden: `thread_local`, not threadprivate.

Should be clear that this is not meant to be used from OpenMP threads.

include/lbann/utils/random.hpp
```diff
@@ -59,6 +59,21 @@ fast_rng_gen& get_fast_generator();
...
+ */
+rng_gen& get_io_generator();
+
+/**
+ * Return a reference to the fast global LBANN random number generator used
+ * for the I/O threads
+ * @note This is stored in a threadprivate variable.
```
ndryden: Ditto.

src/utils/random.cpp
```diff
@@ -89,8 +119,16 @@ bool save_rng_to_checkpoint_shared(persist& p, const lbann_comm* comm) {
     rank_in_world = std::to_string(El::mpi::Rank(El::mpi::COMM_WORLD));
   } else {
     rank_in_world = std::to_string(comm->get_rank_in_world());
-
   }
+
```
ndryden: This will not work right, because it saves only the state for a single thread. I think we should raise a warning until it is fixed properly, since otherwise this will confuse people.

src/utils/random.cpp
```diff
@@ -78,6 +107,7 @@ bool save_rng_to_checkpoint_shared(persist& p, const lbann_comm* comm) {
   rng_name = dirname + "/rng_seq_generator";
```
ndryden: Apparently, we have never been saving this RNG's state correctly.

src/utils/random.cpp
```diff
@@ -136,6 +175,14 @@ bool load_rng_from_checkpoint_shared(persist& p, const lbann_comm* comm) {
     rank_in_world = std::to_string(comm->get_rank_in_world());
   }
 
+  rng_name = dirname + "/rng_io_generator_" + rank_in_world;
```
ndryden: As above: This only restores the state for a single thread.

ndryden:
Works for me.

bvanessen:
Okay.  I will fix the thread RNG storage.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

src/utils/random.cpp
bvanessen: Noted

PR #991, opened 2019-04-14
numpy_conduit_cache class

davidHysom:
Class for uncapsulating zipped numpy arrays inside a conduit::Node.


src/data_readers/numpy_conduit_cache.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace lbann {
+
+
+void numpy_conduit_cache::load(const std::string filename, int data_id) {
+
+  try {
+    m_numpy[data_id] = cnpy::npz_load(filename);
+    conduit::Node &n = m_data[data_id];
+    std::map<std::string, cnpy::NpyArray> &a = m_numpy[data_id];
```
bvanessen: Related to my comment about packed conduit nodes, we may want to have ownership of the array transfer to conduit.

PR #992, opened 2019-04-14
Don't create bias weights with no bias

### LINKED ISSUE #584: Unneeded optimizers created for biases
We still create a weight matrix and associated optimizer for biases even when a bias is not present in the convolution and fully-connected layers. We don't apply the bias during forward/backward prop, but it does lead to an unneeded optimization step and memory use.

ndryden:
If a fully-connected or convolution layer has no bias, we don't need to create weights or an optimizer for it. This closes #584.

While I agree the refactor proposed in #584 is the nicer way to handle this long-term, this achieves the effect without any externally-visible changes. We may want to have a separate issue to track refactoring biases.


ndryden:
I tested it with ResNet by manually adding some biases to layers. I'll use the unit tests before merging.

ndryden:
Passes all three unit tests.

PR #994, opened 2019-04-15
Multi-trainer model initialization and metrics

samadejacobs:
1. Fix a bug where models are not initialize differently as expected in multi-trainer scenario. In particular, we comm->split_trainer() needs to be called first to get appropriate trainer_rank to be mixed with random seed
2. Add stdev to metric scores


samadejacobs:
One thing not included in this PR is setting _random_init_models_differently: true_ as default. Brian and Tim think this are reasonable assumption for multi-trainer/model training. It should be a minor change that will be added later.

PR #996, opened 2019-04-16
In-place gradient buffers

ndryden:
This adds support for doing gradient computations in-place.

The problem with the current `add_to_gradient` method:
* We store two copies of gradient buffers, one within a layer and one within the optimizer. This uses a lot of excess memory when a layer many parameters (big fully-connected layers, wide convolutional layers).
* We perform a bunch of copies and scalings of gradient matrices, which involves excess data movement and computation (ideally: we fuse the scaling).

The solution: An expert-mode interface in optimizers (`get_gradient_buffer`) to get direct access to the underlying gradient buffer, as well as whatever scaling factors need to be applied to it. The contract for using this is that the caller must apply the scaling factor to the buffer before writing to it. The existing `add_to_gradient` method remains in place with no changes.

Layer modifications:
* Convolutional layers now use this for both kernel and bias gradients. cuDNN supports scaling parameters for buffers that we now use to completely fuse the scaling operations. `GEMM` has similar parameters that are used for CPU im2col convolution. Convolutional layers now no longer have their own gradient buffers.
* Fully-connected layers use this for linearity gradients, and bias gradients when the layer is on GPU. (The method used to compute bias gradients on CPU does not let us fuse the scale factors.) Fully-connected layers now no longer have their own linearity gradient buffers (or a bias gradient buffer when using GPUs).

Passes `model_mnist_conv_graph`, `model_mnist_ridge_regression`, and `model_mnist_softmax_classifier` tests. Initial ResNet-50 training also looks good. (I tested the `softmax_classifier` on CPU, and it seemed fine, but the tests were too slow to run to completion.)

Notes on other layers that compute gradients:
* I would like to apply this to batchnorm in order to fuse the copy/scaling. This will require modification to our custom batchnorm kernels, which we should do as part of our general batchnorm optimization.
* L2 weight regularization is currently fine as-is: it just does an Axpy into the gradient buffer via `add_to_gradient`. (But in the future we may want to clean up the order in which weight regularizations are done WRT other computations, and so this might change.)

### DIFF
--- a/include/lbann/layers/learning/base_convolution.hpp
+++ b/include/lbann/layers/learning/base_convolution.hpp
@@ -655,10 +655,10 @@ protected:
     if (m_bias_scaling_factor != DataType(0)
         && m_weights[1]->get_optimizer() != nullptr) {
       optimizer* bias_optimizer = m_weights[1]->get_optimizer();
+      DataType dst_scale = DataType(0);
+      auto& bias_gradient = bias_optimizer->get_gradient_buffer(dst_scale,
+                                                                true);
       if (has_local_data) {
-        DataType dst_scale = DataType(0);
-        auto& bias_gradient = bias_optimizer->get_gradient_buffer(dst_scale,
-                                                                  true);
         CHECK_CUDNN(cudnnConvolutionBackwardBias(
                       cudnn::get_handle(),
                       &gradient_scale,
@@ -668,18 +668,17 @@ protected:
                       m_bias_cudnn_desc,
                       bias_gradient.Buffer()));
       } else {
-        // Ensure we participate in the allreduce.
-        bias_optimizer->get_gradient_buffer(true);
+        El::Scale(dst_scale, bias_gradient);
       }
     }
 
     // Compute kernel gradient
     optimizer* kernel_optimizer = m_weights[0]->get_optimizer();
     if (kernel_optimizer != nullptr) {
+      DataType dst_scale = DataType(0);
+      auto& kernel_gradient = kernel_optimizer->get_gradient_buffer(dst_scale,
+                                                                    true);
       if (has_local_data) {
-        DataType dst_scale = DataType(0);
-        auto& kernel_gradient = kernel_optimizer->get_gradient_buffer(dst_scale,
-                                                                      true);
         // Initialize GPU workspace
         GPUMat workspace;
 #ifdef HYDROGEN_HAVE_CUB
@@ -755,8 +754,7 @@ protected:
                         kernel_gradient.Buffer()));
         }
       } else {
-        // Ensure we participate in the allreduce.
-        kernel_optimizer->get_gradient_buffer(true);
+        El::Scale(dst_scale, kernel_gradient);
       }
     }
 
@@ -921,10 +919,8 @@ protected:
     // Local matrices
     const DMat<Device>& local_input = get_local_prev_activations();
     const DMat<Device>& local_gradient_wrt_output = get_local_prev_error_signals();
-    const bool has_local_data = (local_input.Height() > 0
-                                 && local_input.Width() > 0
-                                 && local_gradient_wrt_output.Height() > 0
-                                 && local_gradient_wrt_output.Width() > 0);
+    const bool has_local_data = (!local_input.IsEmpty()
+                                 && !local_gradient_wrt_output.IsEmpty());
 
     // Get convolution parameters
     const El::Int local_width = local_input.Width();
@@ -945,10 +941,10 @@ protected:
     if (m_bias_scaling_factor != DataType(0)
         && this->m_weights[1]->get_optimizer() != nullptr) {
       optimizer* bias_optimizer = this->m_weights[1]->get_optimizer();
+      DataType dst_scale = DataType(0);
+      auto& bias_gradient = bias_optimizer->get_gradient_buffer(
+        dst_scale, true);
       if (has_local_data) {
-        DataType dst_scale = DataType(0);
-        auto& bias_gradient = bias_optimizer->get_gradient_buffer(
-          dst_scale, true);
         auto& local_bias_gradient = bias_gradient.Matrix();
         LBANN_OMP_PARALLEL_FOR
         for (int channel = 0; channel < num_output_channels; ++channel) {
@@ -969,8 +965,7 @@ protected:
             + gradient_scale*sum;
         }
       } else {
-        // Ensure we participate in the allreduce.
-        bias_optimizer->get_gradient_buffer(true);
+        El::Scale(dst_scale, bias_gradient);
       }
     }
 
@@ -990,10 +985,10 @@ protected:
                    get_output_size() / num_output_channels);
     DMat<Device> im2col_matrix(m, k);
 
+    DataType dst_scale = DataType(0);
+    auto& kernel_gradient = kernel_optimizer->get_gradient_buffer(
+      dst_scale, true);
     if (has_local_data) {
-      DataType dst_scale = DataType(0);
-      auto& kernel_gradient = kernel_optimizer->get_gradient_buffer(
-        dst_scale, true);
       auto& local_kernel_gradient = kernel_gradient.Matrix();
       DMat<Device> kernel_gradient_matrix(m, n, local_kernel_gradient.Buffer(), m);
 
@@ -1033,8 +1028,7 @@ protected:
         }
       }
     } else {
-      // Ensure we participate in the allreduce.
-      kernel_optimizer->get_gradient_buffer(true);
+      El::Scale(dst_scale, kernel_gradient);
     }
   }
 
--- a/include/lbann/layers/learning/base_convolution.hpp
+++ b/include/lbann/layers/learning/base_convolution.hpp
@@ -645,7 +645,6 @@ protected:
 
     // Useful constants
     const int effective_mini_batch_size = this->m_model->get_effective_mini_batch_size();
-    const DataType gradient_scale = DataType(1) / effective_mini_batch_size;
     const bool has_local_data = (local_input.Height() > 0
                                  && local_input.Width() > 0
                                  && local_gradient_wrt_output.Height() > 0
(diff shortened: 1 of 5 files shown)


ndryden:
@bvanessen I am confused what you are asking for here. @timmoon10 is proposing to eliminate one variant of the `get_gradient_buffer` method. Do you mean to update the fully-connected to use the optimizer's gradient buffer for CPU biases? That would require reimplementing the bias computation.

src/layers/learning/fully_connected.cpp
```diff
@@ -137,33 +130,31 @@ void fully_connected_layer<data_layout::MODEL_PARALLEL, El::Device::CPU>::bp_com
   // Note: Perform GEMMs independently if possible
   optimizer* linearity_optimizer = this->m_weights[0]->get_optimizer();
   if (linearity_optimizer != nullptr) {
+    DataType dst_scale = DataType(0);
     if (linearity.DistSize() == 1) {
+      auto& linearity_gradient = linearity_optimizer->get_gradient_buffer(
```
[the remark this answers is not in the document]
ndryden: I would prefer to be maximally precise when specifying whether allreduces are needed, to minimize potential overheads.

src/layers/learning/fully_connected.cpp
```diff
@@ -421,56 +413,54 @@ void fully_connected_layer<data_layout::MODEL_PARALLEL, El::Device::GPU>::bp_com
...
     }
   }
 
   // Compute gradient w.r.t. linearity if needed
   // Note: Perform GEMMs independently if possible
   optimizer* linearity_optimizer = this->m_weights[0]->get_optimizer();
   if (linearity_optimizer != nullptr) {
+    DataType dst_scale = DataType(0);
     if (linearity.DistSize() == 1) {
+      auto& linearity_gradient = linearity_optimizer->get_gradient_buffer(
```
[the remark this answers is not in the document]
ndryden: Ditto: Let's be precise when allreduces are needed.

include/lbann/layers/learning/base_convolution.hpp
```diff
@@ -669,44 +644,42 @@ class base_convolution_layer : public Layer {
...
+                      cudnn::get_handle(),
+                      &gradient_scale,
+                      m_tensors_cudnn_desc.get_prev_error_signals(),
+                      local_gradient_wrt_output.LockedBuffer(),
+                      &dst_scale,
+                      m_bias_cudnn_desc,
+                      bias_gradient.Buffer()));
+      } else {
+        // Ensure we participate in the allreduce.
+        bias_optimizer->get_gradient_buffer(true);
```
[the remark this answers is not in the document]
bvanessen: With my earlier comment, I was trying to agree with Tim that we should probably only have one get_gradient_buffer function.  Additionally, we probably don't want get_gradient_buffer to performance a hidden scale operation.

src/optimizers/optimizer.cpp
```diff
@@ -201,6 +201,55 @@ void optimizer::clear_gradient() {
...
+  return *m_gradient;
+}
+
+AbsDistMat& optimizer::get_gradient_buffer(bool allreduce_needed) {
+  DataType buf_scale;
+  auto& buf = get_gradient_buffer(buf_scale, allreduce_needed);
+  if (buf_scale == DataType(0)) {
+    El::Zero(buf);
+  } else if (buf_scale != DataType(1)) {
+    El::Scale(buf_scale, buf);
```
bvanessen: We probably don't want to have hidden work her in what looks like a method that should just return a reference to a matrix.

bvanessen:
Sorry for the confusion. I was trying to agree with Tim about the get_gradient_buffer methods, and I added some more specific comments. I am not suggesting that we update the fully connected cpu implementation at this time.

src/optimizers/optimizer.cpp
```diff
@@ -201,6 +201,55 @@ void optimizer::clear_gradient() {
...
+  case optimizer_gradient_status::cleared:
+    buf_scale = DataType(0);
+    m_gradient_status = (allreduce_needed ?
+                         optimizer_gradient_status::allreduce_needed :
+                         optimizer_gradient_status::ready);
+    break;
+  case optimizer_gradient_status::allreduce_needed:
+    buf_scale = (allreduce_needed ?
+                 DataType(1) :
+                 DataType(1) / m_gradient->RedundantSize());
```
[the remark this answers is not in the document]
ndryden: Good catch, thanks! Fixed via the second approach, in the interest of performing as little communication as possible.

ndryden:
Addressed issues. Main changes:
* The variant of `get_gradient_buffer` has been eliminated; layers explicitly scale if needed.
* `get_gradient_buffer` gives its caller a scale factor for their gradient contribution too. This adds further complexity, but we're already in expert-mode, and minimizing communication seems more important.

include/lbann/layers/learning/base_convolution.hpp
```diff
@@ -730,63 +703,60 @@ class base_convolution_layer : public Layer {
...
+                        kernel_gradient_cudnn_algorithm,
+                        workspace.Buffer(),
+                        workspace_size,
+                        &dst_scale,
+                        m_kernel_cudnn_desc,
+                        kernel_gradient.Buffer()));
         }
-
+      } else {
+        El::Scale(dst_scale, kernel_gradient);
```
bvanessen: Should this be dst_scale or gradient_scale here.  Earlier in this code block the gradient_scale is divided by the effective_mini_batch_size, and dst_scale is not.

include/lbann/layers/learning/base_convolution.hpp
```diff
@@ -986,10 +962,11 @@ class base_convolution_layer : public Layer {
...
             }
           }
-          local_bias_gradient(channel, 0) = m_bias_scaling_factor * sum;
+          local_bias_gradient(channel, 0) = dst_scale*local_bias_gradient(channel, 0)
+            + gradient_scale*sum;
         }
-        const DataType bias_scale = m_bias_scaling_factor / effective_mini_batch_size;
-        bias_optimizer->add_to_gradient(m_bias_gradient, bias_scale, true);
+      } else {
+        El::Scale(dst_scale, bias_gradient);
```
bvanessen: See previous note about gradient_scale.

include/lbann/layers/learning/base_convolution.hpp
```diff
@@ -1008,51 +985,53 @@ class base_convolution_layer : public Layer {
...
+                 m_pads.data(),
+                 &kernel_dims[2],
+                 m_strides.data());
+          El::Gemm(El::NORMAL, El::NORMAL,
+                   gradient_scale, im2col_matrix, gradient_wrt_output_col,
+                   dst_scale, kernel_gradient_matrix);
+        }
       }
+    } else {
+      El::Scale(dst_scale, kernel_gradient);
```
bvanessen: and again.

src/optimizers/optimizer.cpp
```diff
@@ -201,6 +201,48 @@ void optimizer::clear_gradient() {
...
+  if (m_gradient_status == optimizer_gradient_status::allreduce_started) {
+    finish_gradient_allreduce();
+  }
+  // Determine scaling factor and transition state.
+  switch (m_gradient_status) {
+  case optimizer_gradient_status::ready:
+    buf_scale = DataType(1);
+    in_scale = DataType(1);
+    if (allreduce_needed) {
+      buf_scale /= m_gradient->RedundantSize();
```
bvanessen: Is redundant size here equal to the mini-batch size or the effective-mini-batch size?

src/layers/learning/fully_connected.cpp
```diff
@@ -240,19 +234,19 @@ void fully_connected_layer<data_layout::DATA_PARALLEL, El::Device::CPU>::bp_comp
   // Compute gradient w.r.t. linearity if needed
   optimizer* linearity_optimizer = this->m_weights[0]->get_optimizer();
   if (linearity_optimizer != nullptr) {
+    DataType dst_scale = DataType(0), gradient_scale = DataType(0);
+    auto& linearity_gradient = linearity_optimizer->get_gradient_buffer(
+      dst_scale, gradient_scale, true);
+    gradient_scale /= mini_batch_size;
```
bvanessen: I am trying to reconcile how the gradient_scale is modified by the mini-batch size and the redundant size of the matrix.

src/layers/learning/fully_connected.cpp
```diff
@@ -42,7 +41,6 @@ void fully_connected_layer<data_layout::DATA_PARALLEL, El::Device::CPU>
   ::setup_matrices(const El::Grid& grid) {
   learning_layer::setup_matrices(grid);
   deallocate_matrices();
-  m_linearity_gradient = new StarMat<El::Device::CPU>(grid);
   m_bias_gradient = new StarMat<El::Device::CPU>(grid);
```
bvanessen: Why is the bias gradient still here, and not using a reference like the linearity gradient.

include/lbann/layers/learning/base_convolution.hpp
ndryden: `dst_scale` is correct. In this route, the layer's computed gradient is 0, so it would be meaningless to use `gradient_scale`. We still need to apply the correct scaling factor to the returned gradient buffer, because the rank will still participate in the allreduce. Compare with the `add_to_gradient` method.

(This confusion is why I originally had this hidden within the optimizer class.)

src/optimizers/optimizer.cpp
ndryden: Neither. This is needed to properly combine contributions that will not (or have already been) allreduced with contributions that will be allreduced. Compare with `add_to_gradient`; this is doing precisely the same computation, just changing who is responsible for different aspects.

src/layers/learning/fully_connected.cpp
ndryden: See what I wrote in the opening description of this PR.

src/layers/learning/fully_connected.cpp
ndryden: See above. It may be helpful to realize that, in nearly all our code right now, `gradient_scale` will be `1.0`.

PR #997, opened 2019-04-18
Parallelism in Python data reader

timmoon10:
### Summary

This achieves parallelism in the Python data reader by calling into Python's `multiprocessing` module.

### Details

The Python interpreter (for CPython) is not thread-safe, so it uses the global interpreter lock to serialize access. This precludes multithreading with arbitrary Python code (although some multithreading can be achieved using operations that deliberately relinquish the GIL, like IO access and some math operations). Instead, we use the `multiprocessing` module to fork the master process and create a pool of worker processes. This is the approach used in PyTorch. This has several limitations:
- We require interprocess communication between the master and worker processes. I don't know how this is done in the guts of `multiprocessing`.
- The process of extracting data from Python and putting it in a C++ buffer is still serial.
- The Python function to get data samples must return a serializable type, like `list`. Iterators can't be communicated between processes.
- Forking the master process may increase our memory usage.

### Results

Mini-batch times running LeNet with MNIST (1 Pascal node; 2 procs/node):

Data reader | Mini-batch time (second epoch)
---|---
MNIST | 1.5 ms
Serial Python | 9.9 ms
Parallel Python | 5.0 ms


src/data_readers/data_reader_python.cpp
```diff
@@ -229,32 +255,69 @@ int python_reader::get_linearized_label_size() const {
...
+      X(row, col) = PyFloat_AsDouble(val);
+    }
   }
-  if (PyErr_Occurred()) { LBANN_ERROR("Python error detected"); }
 
   return true;
 }
 
+bool python_reader::fetch_datum(CPUMat& X, int data_id, int col) {
+  return true;
```
ndryden: It might be better to leave the default implementation (throws an error) than this, in case something changes and this gets called directly.

PR #998, opened 2019-04-18
ResNet cleanup/generalization

ndryden:
* Adds Wide ResNet-50-2 to the contrib models.
* ResNet models take a width (defaults to 1).
* Adds support for more generic ResNets to the `model_zoo/vision/resnet.py` driver:
  * Arbitrary ResNet architectures.
  * Arbitrary widths.
* Updates default learning rate and epoch for ResNet.


ndryden:
This is a good point, and would be something to do as part of supporting a more general set of residual networks (ResNeXt, etc.). I think the simplicity right now is fine though.

PR #1005, opened 2019-04-19
data_reader_jag_conduit: more verbose exceptions

davidHysom:
modified a couple of exceptions to be more informative


davidHysom:
I'm merging, since no one has reviewed, and this is a very minor change, and it may be helpful in tracking down some bugs in the sample_list class.

PR #1006, opened 2019-04-19
Bugfix for LTFB to mark data store loaded

bvanessen:
Work around to let LTFB properly mark the data store as loaded during a tournament.


davidHysom:
I merged with my clone of head of develop and ran LTFB with --use_data_store but without --preload_data_store. I got the duplicate_id error.

davidHysom:
Update: the duplicate_id error was for the validate data reader

bvanessen:
Hmmm, that doesn't make sense.

PR #1008, opened 2019-04-20
Python/experiment support for LSF/Lassen

### LINKED ISSUE #601: Experiment script for Sierra/Lassen
I think we should make an effort to put reasonable defaults in experiments/run_lbann_experiment.sh so that people can just run experiments without intimate knowledge of each system they run on. This is, of course, contingent on Sierra being stable enough to have good defaults at all.

ndryden:
This adds support for LSF as a launcher and for Lassen paths in the Python frontend. I added support for Sierra as well, assuming it works exactly the same as Lassen. Support for specifying a DAT reservation is now possible, but only supported for LSF. I left the default banks for Lassen and Sierra as `None`, since practically everyone needs to use a different bank.

This also reorganizes how scheduler arguments are managed, and tweaks the relevant arguments. Let me know if they're reasonable or if we need to iterate.

Closes #601.


python/lbann/contrib/args.py
```diff
@@ -28,9 +29,41 @@ def add_scheduler_arguments(parser):
...
+        '--reservation', action='store', type=str,
+        help='scheduler reservation', metavar='NAME')
     parser.add_argument(
         '--time-limit', action='store', type=int,
         help='time limit (in minutes)', metavar='MIN')
+    parser.add_argument(
+        '--setup-only', action='store_true',
+        help='set up (but do not run) the experiment')
+
+def get_launcher_kwargs(args):
```
timmoon10: To make it match `add_scheduler_arguments`, how about:
```suggestion
def get_scheduler_kwargs(args):
```

python/lbann/contrib/lc/launcher.py
```diff
@@ -46,6 +47,14 @@ def run(model, data_reader, optimizer,
         environment['OMP_NUM_THREADS'] = 8
         environment['AL_PROGRESS_RANKS_PER_NUMA_NODE'] = 2
 
+    # Magic default arguments to jsrun/etc.
+    if system in ('sierra', 'lassen'):
+        if scheduler == 'lsf':
+            launcher_args += ' -d packed -b "packed:10" -r 1 -c 40 -g 4'
```
timmoon10: We should document what's going on in this incantation.

python/lbann/launcher/__init__.py
```diff
@@ -53,6 +55,7 @@ def run(lbann_exe, model, data_reader, optimizer,
         system (str, optional): Target system.
         partition (str, optional): Scheduler partition.
         account (str, optional): Scheduler account.
+        reservation (str, optional): Reservation name (for DATs).
```
timmoon10: As far as I can tell, "DAT" is an LLNL-specific term.
```suggestion
        reservation (str, optional): Scheduler advance reservation.
```

python/lbann/launcher/lsf.py
```diff
@@ -0,0 +1,130 @@
...
+        experiment_dir (str, optional): Experiment directory.
+        nodes (int, optional): Number of compute nodes.
+        procs_per_node (int, optional): Number of processes per compute
+            node.
+        time_limit (int, optional): Job time limit, in minutes. A
+            negative value implies the system-default time limit.
+        job_name (str, optional): Batch job name.
+        partition (str, optional): Scheduler partition.
+        account (str, optional): Scheduler account.
+        reservation (str, optional): Reservation name (for DATs).
```
timmoon10:
```suggestion
        reservation (str, optional): Scheduler advance reservation.
```

python/lbann/launcher/slurm.py
```diff
@@ -32,6 +33,7 @@ def run(command,
         job_name (str, optional): Batch job name.
         partition (str, optional): Scheduler partition.
         account (str, optional): Scheduler account.
+        reservation (str, optional): Reservation name.
```
timmoon10:
```suggestion
        reservation (str, optional): Scheduler advance reservation.
```

model_zoo/vision/alexnet.py
```diff
@@ -92,24 +85,18 @@
...
-        classes = args.imagenet_classes
-        kwargs['lbann_args'] = (
-            '--data_filedir_train={} --data_filename_train={} '
-            '--data_filedir_test={} --data_filename_test={}'
-            .format(imagenet_dir(data_set='train', num_classes=classes),
-                    imagenet_labels(data_set='train', num_classes=classes),
-                    imagenet_dir(data_set='val', num_classes=classes),
-                    imagenet_labels(data_set='val', num_classes=classes)))
+    kwargs = lbann.contrib.args.get_launcher_kwargs(args)
+    classes = args.num_labels
```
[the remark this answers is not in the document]
ndryden: That is true, but it's a particularly ugly implementation detail that should probably be cleaned up elsewhere. The number of labels and the number of classes should be identical, so I'm hesitant to add separate arguments, which would just be confusing. We could hardcode the scripts to assume 1000 classes as a stop-gap until the data readers are cleaned up, but that would only actually work for ImageNet. On the other hand, are we using these for anything else right now?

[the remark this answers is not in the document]
timmoon10: Perhaps hardcoding 1000 classes is the best way to go. In any case, I use ImageNet-10 almost exclusively for quick benchmarking and validation, so this is important for my workflow.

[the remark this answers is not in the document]
ndryden: Done. However, I believe we are able to specify the number of labels in the data reader prototext-- and it works in theory, I think. We should revisit this once we have proper data reader support in the Python frontend.

PR #1009, opened 2019-04-20
Autotune cuDNN convolution algorithms

ndryden:
* Adds support for autotuning the cuDNN convolution algorithm selection using `cudnnFindConvolutionForwardAlgorithmEx` (etc.).
* Switches the heuristic algorithm selection from the deprecated `cudnnGetConvolutionForwardAlgorithm` (etc.) to `cudnnGetConvolutionForwardAlgorithm_v7` (etc.).

The implementation is adapted from distconv LBANN.

A layer selects an algorithm once for each unique mini-batch size it sees.

Right now, this always autotunes the algorithm. However, for some networks/datasets (e.g. 3D data), autotuning can take an excessive amount of time. We should find a nice way to enable/disable this without requiring a rebuild. Either environment variables or a command-line flag should work, but we need to come up with a reasonable interface for these sorts of tuning flags.

### DIFF
--- a/src/utils/cudnn.cpp
+++ b/src/utils/cudnn.cpp
@@ -528,6 +528,330 @@ cudnnTensorDescriptor_t& entrywise_layer_tensor_manager::get_error_signals(int p
   return desc;
 }
 
+////////////////////////////////////////////////////////////
+// cuDNN algorithm selection
+////////////////////////////////////////////////////////////
+
+namespace {
+
+// Non-deterministic algorithms.
+std::vector<cudnnConvolutionFwdAlgo_t> nondet_fwd_algos = {};
+std::vector<cudnnConvolutionBwdDataAlgo_t> nondet_bwd_data_algos = {
+  CUDNN_CONVOLUTION_BWD_DATA_ALGO_0
+};
+std::vector<cudnnConvolutionBwdFilterAlgo_t> nondet_bwd_filter_algos = {
+  CUDNN_CONVOLUTION_BWD_FILTER_ALGO_0,
+  CUDNN_CONVOLUTION_BWD_FILTER_ALGO_3
+};
+
+template <typename AlgoType, typename PerfType>
+AlgoType find_best_heuristic_algorithm(
+  const std::vector<PerfType>& perf_results,
+  const std::vector<AlgoType>& deterministic_algos,
+  bool deterministic,
+  size_t max_ws_size) {
+  std::vector<AlgoType> algos;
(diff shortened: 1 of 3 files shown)


src/utils/cudnn.cpp
```diff
@@ -528,6 +528,332 @@ cudnnTensorDescriptor_t& entrywise_layer_tensor_manager::get_error_signals(int p
...
+};
+std::vector<cudnnConvolutionBwdFilterAlgo_t> nondet_bwd_filter_algos = {
+  CUDNN_CONVOLUTION_BWD_FILTER_ALGO_0,
+  CUDNN_CONVOLUTION_BWD_FILTER_ALGO_3
+};
+
+template <typename AlgoType, typename PerfType>
+AlgoType find_best_heuristic_algorithm(
+  const std::vector<PerfType>& perf_results,
+  const std::vector<AlgoType>& deterministic_algos,
```
timmoon10: Do you mean `nondeterministic_algos`?

src/utils/cudnn.cpp
```diff
@@ -528,6 +528,332 @@ cudnnTensorDescriptor_t& entrywise_layer_tensor_manager::get_error_signals(int p
...
+  if (algos.empty()) {
+    LBANN_ERROR("No valid convolution algorithms.");
+  }
+  return algos[0];
+}
+
+template <typename AlgoType, typename PerfType>
+AlgoType find_best_algorithm(
+  const std::vector<PerfType>& perf_results,
+  const std::vector<AlgoType>& deterministic_algos,
```
timmoon10: Do you mean `nondeterministic_algos`?

src/utils/cudnn.cpp
```diff
@@ -528,6 +528,332 @@ cudnnTensorDescriptor_t& entrywise_layer_tensor_manager::get_error_signals(int p
...
+  bool deterministic,
+  size_t max_ws_size) {
+  std::unordered_map<AlgoType, float> time_map;
+  for (const auto& p : perf_results) {
+    if (p.status != CUDNN_STATUS_SUCCESS) {
+      continue;
```
timmoon10: If an algorithm fails some runs and passes in others, then it's runtime will appear smaller than expected. Perhaps failure should add a runtime of infinity to the runtime.

ndryden:
I see very slight improvements for vanilla ResNet-50. In distconv, we've seen much more substantial improvements when networks look less like ResNet-50.

PR #1015, opened 2019-04-24
Hacked bugfix for sendrecv bug in MVAPICH2-2.3.

timmoon10:
I've noticed that LTFB on AlexNet produces strange results when I run on Pascal. During a tournament, the remote model usually produces trash metric scores, even when the model did well on its original trainer. Digging deeper, this seems to be a bug with MPI_Sendrecv in MVAPICH2-2.3. MPI_Sendrecv will output all zeros if communicating processes are on different nodes, the data buffers are in pinned memory, and the message size is between approximately 16 KiB and 16 MiB. This bug does't show up in MVAPICH2-2.2 or MVAPICH2-2.3.1. A bug report has been submitted to LC and they are forwarding it to the MVAPICH folks.

In the meantime, @adammoody has provided an environment variable that fixes the problem for me. LTFB on AlexNet now looks reasonable.


bvanessen:
@timmoon10 @samadejacobs @davidHysom Could this bug be why we are seeing zero data on pascal for JAG data.  Could this bug be corrupting the data store sample exchange?

davidHysom:
"MPI_Sendrecv will output all zeros if communicating processes are on different nodes ..."
This could explain why I haven't seen zeros. For testing (not timing) I regularly run on a single node of pascal with multiple processes (eight, sixteen).

samadejacobs:
@bvanessen, oh yea, datastore has sendrcv......this is definitely a possibility. However, what am not sure about is that I think I observe the zero data bug with single node (single trainer) 2 processes(GPUs) in addition to the multi-trainer scenarios. For single trainer case I am not sure because I can't find my log/plot.

experiments/run_lbann_experiment.sh
```diff
@@ -371,6 +371,7 @@ case ${CLUSTER} in
         ;;
 esac
 echo "export MV2_USE_RDMA_CM=0"                         >> ${BATCH_SCRIPT}
```
samadejacobs: You may want to add the `export MV2_USE_RDMA_CM=0` to the other three scripts, at least for now.

PR #1017, opened 2019-04-24
data_store/data_reader refactor

davidHysom:
moving some data_store calls from data_store_jag_conduit to generic_data_reader.
This is far from perfect, but better: must of the data_store instantiation and setup calls are now generalized in the generic_data_reader. Any data_reader that uses the data store needs (1) to call generic_data_store::instantiate_data_store(...) at the appropriate place in its load() method; override the generic_data_reader::preload_data_store() method. Once the sample_list class is de-coupled from the data_reader_jag_conduit and generalized, we should be able to get away with a single, non-overriden preload_data_store() method.

This PR is a pre-requisite for finishing the new data_reader_numpy_npz_conduit class.


davidHysom:
BTW, --preload_data_store now implies --use_data_store, so no need to specify both.

PR #1021, opened 2019-04-25
Fix build error from #1009

timmoon10:
When I type `scripts/build_lbann_lc.sh --clean-build` on Pascal, I get the following error:
```
/usr/WS1/moon13/src/lbann/src/utils/cudnn.cpp:580:39:   required from 'AlgoType lbann::cudnn::{anonymous}::find_best_algorithm(const std::vector<F2>&, const std::vector<_RealType>&, bool, size_t) [with AlgoType = cudnnConvolutionFwdAlgo_t; PerfType = cudnnConvolutionFwdAlgoPerf_t; size_t = long unsigned int]'
/usr/WS1/moon13/src/lbann/src/utils/cudnn.cpp:711:52:   required from here
/usr/tce/packages/gcc/gcc-4.9.3/include/c++/4.9.3/bits/hashtable_policy.h:85:33: error: no match for call to '(const std::hash<cudnnConvolutionFwdAlgo_t>) (const cudnnConvolutionFwdAlgo_t&)'
  noexcept(declval<const _Hash&>()(declval<const _Key&>()))>
```
This bug seems to be from PR #1009. I fixed it by changing an `unordered_map` to a `map`. It looks like cuDNN algorithms aren't hashable by default, at least with the default configuration on Pascal.

### DIFF
--- a/src/utils/cudnn.cpp
+++ b/src/utils/cudnn.cpp
@@ -29,6 +29,7 @@
 
 #include "El.hpp"
 #include <iostream>
+#include <map>
 #include <unordered_map>
 #include <tuple>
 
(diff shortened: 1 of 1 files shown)


ndryden:
This should eventually go away because this is just due ancient compilers not supporting hashing enums.

PR #1023, opened 2019-04-25
Update build script on Pascal

ndryden:
Pascal users: Welcome to modernity!

This forces GCC 7.3.0 and CUDA 9.2, and bumps x86 to the latest NCCL.

Also bumps everyone to cuDNN 7.5.1.


davidHysom:
what about catalyst? Yes, I know it's old --- but I'll take my cycles anyplace I can get them. (Actually I've been running on pascal pvis qos=standby lately, with great availability. Perhaps I shouldn't have said that.)

ndryden:
I think since we're not using GPUs on Pascal, everything should continue to work fine? Let me know if that's not the case.

PR #1024, opened 2019-04-26
fast sample list

davidHysom:
added method: read_inclusive_list_fast(). This method populates the sample_list class using only data contained in the sample list text files. It does not assume the actual data is conduit (or anything else). This greatly reduces load time, hence is a boon to development.debugging. It also permits use of the existing sample lists for non-conduit data, eg, mpz files.

### DIFF
--- a/include/lbann/data_readers/sample_list_jag.hpp
+++ b/include/lbann/data_readers/sample_list_jag.hpp
@@ -272,6 +272,16 @@ class sample_list_jag {
   /// read the body of inclusive sample list
   void read_inclusive_list(std::istream& istrm, size_t stride=1, size_t offset=0);
 
+  /// read the body of inclusive sample list. This method does not assume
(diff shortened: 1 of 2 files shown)


davidHysom:
To use, add to cmd line: --fast_sample_list

bvanessen:
Isn't this PR really about reading data that was not in HDF5 files, not conduit?  Conduit is just our approach to getting access to the underlying HDF5 file.

davidHysom:
with this PR I can use the sample list with numpy arrays, and possibly have a single data_store::preload method. But the real reason I wrote it is to decrease wait time during development whiile all the conduit files are opened and the samples checked. IMO that may be a good sanity check for production runs, bit it really slows down startup times during development/debugging. Frankly, the sample_list class is way too specialized at present to be any sort of a general purpose solution.

davidHysom:
Closing this, since other changes to the sample_list class render it NA

PR #1025, opened 2019-04-30
Fixed a bug in how the data store is being loaded during the

bvanessen:
tournament use case of the validation data set.  There is now an
explicit flag that tells the data store to load the data and is used
in conjunction with explicitly setting the data store as preloaded
after the first tournament.  Note that this should be cleaned up with
the execution environment of the trainer refactor.


samadejacobs:
Looks like this is merged already. For the record, I typically use `--use_data_store` and `--preload_datastore`, I would expect that for JAG you need both, is this for scenario other than JAG? Also, is 'proliferation' of flags not a concern?

PR #1026, opened 2019-04-30
Install Python frontend in CMake build process

### LINKED ISSUE #973: Install Python frontend during build process
### The status quo
- Users manually install the frontend by calling pip on the source directory, e.g. `pip3 install --user -e lbann/python`.
- The frontend requires objects in the LBANN install directory. It can find them if LBANN has been built with `scripts/build_lbann_lc.sh`, but not with Spack.
  - For prototext generation, it is sufficient to set the `PYTHONPATH` environment variable to `install/share/python`.
  - Launching LBANN from Python is not supported with Spack builds.

### Steps
- [ ] Create a copy of the Python package in the install directory. I don't think we should require users to have access to the original LBANN source directory if all they need is an install (possibly created by someone else).
- [ ] Make sure the Python frontend knows the install directory. This can be done by creating a small config file during the build process and detecting it in the `install_dir` function in `lbann.contrib.lc.paths`.
- [ ] Figure out how to install the package, while making it easy to switch between different installations (e.g. for different systems).
  - [Spack can create modules for the packages it builds](https://spack.readthedocs.io/en/latest/module_file_support.html#). It would be nice to make things as easy as typing `module load lbann`.
  - If we assume the user is already inside virtualenv during the build process, we can just call `pip install`. Things get a bit messy if the user wants to use the same build for multiple virtualenv sessions.

timmoon10:
### Summary
The Python frontend is now automatically installed during the build process. This is a first pass and I expect this may need to go through a few rounds of iteration.
This closes #973.

### Usage
If LBANN is built with Spack, you should just be able to load the `lbann` Python package (at least, this is the desired behavior; pinging @bvanessen about https://github.com/spack/spack/pull/11395). If building with CMake or `scripts/build_lbann_lc.sh`, you need to set the `PYTHONPATH` environment variable to point to the install directory:
```sh
cd path/to/lbann
scripts/build_lbann_lc.sh --clean-build
export PYTHONPATH=path/to/install/share/python:${PYTHONPATH}
model_zoo/vision/lenet.py
```
Alternatively, if you have a virtual environment enabled, `scripts/build_lbann_lc.sh` has an option to install directly to the Python site-packages directory:
```sh
cd path/to/lbann
scripts/build_lbann_lc.sh --python-site-packages --clean-build
model_zoo/vision/lenet.py
```

### Issues (possibly for future PRs)
- [x] Need to update documentation.
- [x] We don't cleanly remove previous installations of the LBANN frontend, but just "overwrite" it. This could lead to issues if we do things like change the directory structure.
- [ ] This doesn't install Python dependencies.
- [ ] Our interaction with setuptools is hacky (see below).
- [ ] It would be convenient for developers if we supported installing an editable package (à la `pip3 install -e lbann/python`).

### setuptools
The Python package installation is performed with setuptools, similar to our current approach. Python best practices are to put setup.py at the package root and the popular packaging tools (distutils, setuptools, wheel) prefer or require relative paths. This seems to be a deliberate design decision, although I'm not sure the reasons. However, we need to insert a config file containing install-specific file paths and make sure setup.py can pick it up. I see three approaches for the build process:

1. Inject the config file into a known location in the source directory so that setup.py can pick it up.
2. Copy the Python source tree into the build directory and insert setup.py and the config file there.
3. Create setup.py and the config file in the build directory and pass the source directory as a messy relative path.

We go for option 3 since it's simple and lightweight, but it runs counter to the spirit of the "only relative paths" rule. If we learn about any nicer approaches, we should use them.


ndryden:
> Python best practices are to put setup.py at the package root and the popular packaging tools (distutils, setuptools, wheel) prefer or require relative paths. This seems to be a deliberate design decision, although I'm not sure the reasons.

Basically, virtual environments is the reason.

bvanessen:
I am working on updating the spack environment to work with this PR.

timmoon10:
I've updated the PR to install the Python package in the install directory, at least by default. The user (or Spack) is responsible for modifying `PYTHONPATH` to properly pick it up.

Alternatively, if you have a virtual environment, you can pass `--python-site-packages` as an argument into `scripts/build_lbann_lc.sh`. This will install the package in the Python site-packages directory, so it will pick it up without any additional steps. This will fail if you are not in a virtual environment since the system site-packages directory requires root access.

PR #1027, opened 2019-05-01
CIFAR10 data reader cleanup

ndryden:
_Note: This depends on #1014._

This is mostly a cleanup of the CIFAR-10 data reader so that it can use the full image preprocessing features of the new preprocessing pipeline.
* Now also supports CIFAR-100 by specifying the number of labels as 100. (I thought about renaming the data reader to just "CIFAR", but that would be unclear since CIFAR is a big organization.)
* We no longer require a custom repackaged version of the CIFAR-10 dataset that merges everything into a single file. Instead, the data reader now requires the files be as officially distributed. This should make it easy for other people to use it. (I thought about generalizing this, but the CIFAR-10/100 binary format is custom, and this is a data reader specifically for CIFAR-10/100, not other data sets.)
* I updated the CIFAR-10 data reader to use some new preprocessing (normalizing by the dataset means/standard deviations, random horizontal flips) that seem standard. There are some other standard preprocessing steps that we don't yet support (padded cropping, color jitter).

I tested on Lassen with `model_conv_autoencoder_cifar10` and it seems like it learned/generalized without issue. I don't believe we have any other models that use CIFAR-10.

### DIFF
--- a/include/lbann/data_readers/data_reader_cifar10.hpp
+++ b/include/lbann/data_readers/data_reader_cifar10.hpp
@@ -23,7 +23,7 @@
 // implied. See the License for the specific language governing
 // permissions and limitations under the license.
 //
-// lbann_data_reader_cifar10 .hpp .cpp - generic_data_reader class for CIFAR10 dataset
+// data_reader_cifar10 .hpp .cpp - Data reader for CIFAR-10/100
 ////////////////////////////////////////////////////////////////////////////////
 
 #ifndef LBANN_DATA_READER_CIFAR10_HPP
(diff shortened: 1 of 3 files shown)


ndryden:
@samadejacobs Please review since you were using CIFAR10 a while ago. In particular, note the change in data storage format to use the official dataset distribution.

@timmoon10 Mostly code style/cleanliness.

samadejacobs:
Your code clean up LGTM. I haven't used CIFAR10/100 dataset in a while, the last time I did it was on the same _model_conv_autoencoder_cifar10_ you used for your testing. Not necessarily for this PR, I would suggest we add a standard CIFAR10/100 network (if one exists!) to LBANN standard cv-related networks (e.g., Lenet and Resnet). I have seen CIFAR10/100 used a lot in cv benchmarks.

ndryden:
I agree. There are many standard networks for CIFAR-10 (ResNets among them).

PR #1029, opened 2019-05-03
Data Store: local data cache

davidHysom:
This PR introduces changes to the data_store_conduit class that enable it to be used as a local data cache. Use as a local data cache also requires modifications to data reader classes; those will be dealt with in a subsequent PR.


davidHysom:
closing, because some unrelated changes snuck in.

PR #1036, opened 2019-05-11
Fix grouped convolution.

naoyam:
cuDNN expects the number of channels per group. This was the case before
commit 310f4b4e222728c5bb14a8cf9594589ecb7a5db0.


naoyam:
deconvolution.hpp may also need to be fixed.

ndryden:
Good catch.

naoyam:
I haven't checked it with the develop branch, but it runs fine with the distconv branch

PR #1039, opened 2019-05-14
Change LBANN to install python in normal site-packages directory

bvanessen:
Changed CMake so that all of the python code is installed in
<lbann install dir>/lib/python<Major>.<Minor>/site-packages to conform to standard
practices.  @benson31 is going to write a module file that will be generated by the developers build so that it is easy to modify the users PYTHONPATH.

The more that I think about it, the more I think that we should not allow the CMake environment to install in ~/.local/.../site-packages.  This sets us up for an enormous versioning headache.


benson31:
All the Python stuff installs under `${PYTHON_INSTALL_PREFIX}`, which defaults to `${CMAKE_INSTALL_PREFIX}`. Within that directory, the package is installed in `${CMAKE_INSTALL_PYTHONDIR}`, which defaults to `lib/python-<python major>.<python minor>/site-packages`. A clever user can fiddle with these at their own risk (OFFICIAL (and only) DISCLAIMER: Change these at your own risk -- I'm not responsible for any havoc that ensues.). Of course a clever user can also just use the `setup.py` file to install wherever they want with any of the extra goodies/etc that they want.

A message is printed during the installation warning users where the python package has been installed. It even gives bourne-compatible shell users the command to run to add it to their `PYTHONPATH`. Unfortunately, there's not a good way to get it to print at the end of the installation process, so it's just somewhere in the middle. But it's surrounded by asterisks and white space, so it should be easy to find in a log.

A module file named `${LBANN_MODULEFILE_NAME}` is added to the install tree at `${CMAKE_INSTALL_SYSCONFDIR}/modulefiles`. This is `etc/modulefiles` by default. Loading that will put the LBANN binaries in your `PATH`, the LBANN libraries in your `LD_LIBRARY_PATH`, and the LBANN Python package in your `PYTHONPATH`. It also sets `LBANN_DIR` to be the path to the LBANN export -- this will make `find_package(LBANN)` to "just work" for downstreams. It will not change any other environment variables (cough spack cough) because I find modifying more than that to be more detrimental than helpful (e.g., it tends to cause confusion during builds as to why things have/haven't worked as expected).

docs/building_lbann.rst
```diff
@@ -293,6 +293,8 @@ Hydrogen, and LBANN separately, by whatever means they choose.
           ${LBANN_HOME}/superbuild
 
         ninja
+        ml use ${LBANN_INSTALL_DIR}etc/modulefiles/
+        ml load lbann
```
timmoon10: This doesn't quite work for me. I need to do `ml load lbann-0.99.0`.

CMakeLists.txt
```diff
@@ -737,3 +753,45 @@ string(APPEND _str
...
+whatis(\"Description: Livermore Big Artificial Neural Network Toolkit. A distributed memory, HPC-optimized, model and data parallel training toolkit for deep neural networks.\")
+whatis(\"URL: https://github.com/llnl/lbann\")
+
+prepend_path(\"PATH\",\"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}\")
+prepend_path(\"LD_LIBRARY_PATH\",\"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}\")
+prepend_path(\"PYTHONPATH\",\"${PYTHON_INSTALL_PREFIX}/${CMAKE_INSTALL_PYTHONDIR}\")
+
+pushenv(\"LBANN_DIR\",\"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DIR}\")
+")
+file(WRITE "${CMAKE_BINARY_DIR}/${LBANN_MODULEFILE_NAME}.install" "${_LBANN_MODULE_TEXT}")
```
timmoon10: Would this be better with `configure_file`?

benson31: Yeah probably. It just seemed so short I didn't know if I wanted to bother. But you're probably right.

benson31: I have changed it to use a configure file. I took the opportunity to dump some config information into the comments of the module file. This isn't necessary, but might be useful for folks trying to remember what they did (spack dumps the entire CMake line into their modules).

You can get the path to the lua file from `ml show <modulename>` and then read that file (`ml show` strips comments). If it would be useful, we could also dump these configuration things as part of the help string or the whatis information, but this seemed like overkill for this.

benson31: It seemed like overkill, but @bvanessen liked it in the whatis information so it's in the whatis information. `ml whatis lbann-0.99.0`

docs/building_lbann.rst
```diff
@@ -293,6 +293,8 @@ Hydrogen, and LBANN separately, by whatever means they choose.
           ${LBANN_HOME}/superbuild
 
         ninja
+        ml use ${LBANN_INSTALL_DIR}/etc/modulefiles/
+        ml load lbann-0.99.0
```
timmoon10: We'll need to remember to update this every time we up the version.

benson31: What could go wrong?

PR #1045, opened 2019-05-15
Support specifying random seeds in Python

ndryden:
For `resnet.py` this adds a `--random-seed` argument.


ndryden:
This will need updates, presumably, if the random seed gets moved around in the trainer refactor.

Also, we should look at abstracting common arguments to Python drivers so every script doesn't have to duplicate them.

timmoon10:
> Also, we should look at abstracting common arguments to Python drivers so every script doesn't have to duplicate them.

Isn't this handled in `lbann.contrib.args`?

ndryden:
Maybe. I was thinking things like mini-batch size, etc.

timmoon10:
Yea, I see it would be helpful to implement a function like `add_model_arguments`.

PR #1051, opened 2019-05-21
local cache: shared memory segment

davidHysom:
added method: get_image_sizes() and associated data structures.
This method will be used when the data store is being used as
a local cache, and all ranks on a node read/write images to
a shared memory segment.


ndryden:
Just as a comment on this, I would prefer POSIX, not SysV, shared memory. LC seems to be more willing to bump limits on the former rather than the latter.

davidHysom:
> Just as a comment on this, I would prefer POSIX, not SysV, shared memory. LC seems to be more willing to bump limits on the former rather than the latter.

OK; will change

src/data_store/data_store_conduit.cpp
```diff
@@ -1006,5 +1056,292 @@ void data_store_conduit::set_preload() {
...
+}
+
+void data_store_conduit::exchange_images(std::vector<char> &work, std::unordered_map<int,int> &image_sizes, std::vector<std::vector<int>> &indices) {
+  std::vector<char> work2;
+  int node_rank = m_comm->get_rank_in_node();
+  int offset = 0;
+  for (int p=0; p<m_np_in_trainer; p++) {
+    if (m_rank_in_trainer == p) {
+      m_comm->trainer_broadcast<char>(p, work.data(), work.size());
+      if (node_rank == 0) {
```
bvanessen: Is node rank 0 the only one that manages the shared cache?

src/data_store/data_store_conduit.cpp
```diff
@@ -28,26 +28,47 @@
...
+                     << strerror(errno);                        \
+    }                                                           \
+    if (m_output) {                                             \
+      m_output << "ERROR: " << ss_LBANN_ERROR.str()             \
+               << std::endl;                                    \
+      m_output.close();                                         \
+    }                                                           \
+    throw lbann::exception(ss_LBANN_ERROR.str());               \
+  } while (0)
+
```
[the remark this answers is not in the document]
davidHysom: m_output if a private variable (filestream) in data_store_conduit, so it wouldn't work to move it to exception.hpp  It's important to have the filestream closed if an exception is thrown, so I can tell what when wrong. Eventually m_output (and associated writes thereto) will go away; at that time the private LBANN_EXCEPTION in data_store_conduit can also go away.

bvanessen:
Okay.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

src/data_store/data_store_conduit.cpp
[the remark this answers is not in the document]
davidHysom: Also: I think we should not add printing errno to the LBANN_ERROR in exceptions.hpp because:
"The value of errno is defined only after a call to a function for which it is explicitly stated to be set"
Hence, the value of errno may be meaningless for many places in the codebase where an lbann exception is thrown. I added it to the private LBANN_ERROR in data_store_conduit for temporary debugging, because the calls creating the shared memory segment will set it on error.

src/data_readers/data_reader_image.cpp
```diff
@@ -243,16 +238,24 @@ std::vector<image_data_reader::sample_t> image_data_reader::get_image_list_of_cu
...
+
+void image_data_reader::load_conduit_node_from_file(int data_id, conduit::Node &node, std::vector<char> &data) {
+  node.reset();
+  const std::string filename = get_file_dir() + m_image_list[data_id].first;
+  int label = m_image_list[data_id].second;
+  read_raw_data(filename, data);
+  node[LBANN_DATA_ID_STR(data_id) + "/label"].set(label);
+  node[LBANN_DATA_ID_STR(data_id) + "/buffer"].set(data);
+  node[LBANN_DATA_ID_STR(data_id) + "/buffer_size"] = data.size();
+}
```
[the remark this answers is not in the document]
davidHysom: I deleted: load_conduit_node_from_file(int data_id, conduit::Node &node, std::vector<char> &data)
I needed this during my first thought at using data_store as a local cache, but now realize it's no longer needed.

src/data_store/data_store_conduit.cpp
```diff
@@ -1006,5 +1056,292 @@ void data_store_conduit::set_preload() {
...
+  //need to ensure name is unique across all data readers
+  m_seg_name = "/our_town_" + m_reader->get_role();
+
+  //in case a previous run was aborted, attempt to remove the file, which
+  //may or may not exist
+  int node_id = m_comm->get_rank_in_node();
+  if (node_id == 0) {
+    std::stringstream s;
+    s << "rm -rf /dev/shm/" << m_seg_name;
+    system(s.str().c_str());
```
ndryden: I would use `shm_unlink` for this.

src/data_store/data_store_conduit.cpp
```diff
@@ -1006,5 +1056,292 @@ void data_store_conduit::set_preload() {
...
+
+  //need to ensure name is unique across all data readers
+  m_seg_name = "/our_town_" + m_reader->get_role();
+
+  //in case a previous run was aborted, attempt to remove the file, which
+  //may or may not exist
+  int node_id = m_comm->get_rank_in_node();
+  if (node_id == 0) {
+    std::stringstream s;
+    s << "rm -rf /dev/shm/" << m_seg_name;
```
ndryden: I really love my `/../../* --no-preserve-root` data reader role.

ndryden:
Tested on two Lassen nodes, running with
```
AL_PROGRESS_RANKS_PER_NUMA_NODE=2 OMP_NUM_THREADS=4 jsrun -n 2 -a 4  -d packed -b "packed:10" -r 1 -c 40 -g 4 /usr/WS1/dryden1/lbann-ndryden/build/gnu.Release.lassen.llnl.gov/install/bin/lbann --data_filedir_train=/p/gpfs1/brainusr/datasets/ILSVRC2012/original/train/ --data_filename_train=/p/gpfs1/brainusr/datasets/ILSVRC2012/original/labels/train.txt --data_filedir_test=/p/gpfs1/brainusr/datasets/ILSVRC2012/original/val/ --data_filename_test=/p/gpfs1/brainusr/datasets/ILSVRC2012/original/labels/val.txt --prototext=/usr/WS1/dryden1/lbann-ndryden/model_zoo/vision/20190719_110851_resnet50-test-shm/experiment.prototext --data_store_cache --preload_data_store --mini_batch_size=256 --num_epochs=2
```

I get the following error:
```
USING DATA STORE!

starting data_store_conduit::setup() for role: train
data store mode: exchange_data via individual samples
Caught signal 11 (SIGSEGV - invalid memory reference) on rank 7
Stack trace:
   0: lbann::stack_trace::get()
   2: /usr/WS1/dryden1/lbann-ndryden/build/gnu.Release.lassen.llnl.gov/install/lib64/liblbann.so(+0x129bd04) [0x20000139bd04] (could not find stack frame symbol)
   3: __kernel_sigtramp_rt64 (demangling failed)
   4: std::basic_streambuf<char, std::char_traits<char> >::xsgetn(char*, long)
   5: std::basic_filebuf<char, std::char_traits<char> >::xsgetn(char*, long)
   6: std::istream::read(char*, long)
   7: lbann::data_store_conduit::read_files(std::vector<char, std::allocator<char> >&, std::unordered_map<int, int, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<int const, int> > >&, std::vector<int, std::allocator<int> >&)
   8: lbann::data_store_conduit::preload_local_cache()
   9: lbann::data_store_conduit::setup(int)
  10: lbann::generic_data_reader::setup_data_store(int)
  11: lbann::build_model_from_prototext(int, char**, lbann_data::LbannPB&, lbann::lbann_comm*, std::shared_ptr<lbann::thread_pool>, bool)
  12: /usr/WS1/dryden1/lbann-ndryden/build/gnu.Release.lassen.llnl.gov/install/bin/lbann() [0x10013b4c] (could not find stack frame symbol)
  13: /lib64/libc.so.6(+0x25100) [0x2000294f5100] (could not find stack frame symbol)
  14: __libc_start_main (demangling failed)
```

davidHysom:
> Tested on two Lassen nodes, running with
Hm, I'll look into this tomorrow. Clearly my algorithm, "if it runs on pascal, it should run anywhere" is deficient.

src/data_store/data_store_conduit.cpp
davidHysom: Um, I don't understand your comment

ndryden: You pass unsanitized input to `system` (I realize right now we constrain the role to be one of `train`, `validate`, or `test` when the data reader is set up via prototext). I did notice that particular role name won't quite work; `; rm -rf --no-preserve-root /*` will do it instead.

`system` should basically never be used outside of quick debugging hacks.

davidHysom:
> I think it would be good if the data store could clean up the shared memory segment when LBANN exits (assuming a clean exit).

I'm calling shm_unlink in the dtor (also munmap, which is apparently not necessary). Do you have something else in mind? Is this not sufficient?

ndryden:
> I'm calling shm_unlink in the dtor (also munmap, which is apparently not necessary). Do you have something else in mind? Is this not sufficient?

I missed that, this is fine.

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -172,48 +163,59 @@ class data_store_conduit {
...
+  //VBE: please explain what this means!
+  bool m_explicit_loading = false;
 
-  /// rank in the trainer; convenience handle
-  int  m_rank_in_trainer;
+  /// The size of the mini-batch that was used to calculate ownership
+  /// of samples when building the owner map.  This size has to be
+  /// used consistently when computing the indices that will be sent
+  /// and received.
+  int m_owner_map_mb_size = 0;
```
ndryden: Should this be `size_t` for safety?

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -172,48 +163,59 @@ class data_store_conduit {
...
 
-  /// number of procs in the trainer; convenience handle
-  int  m_np_in_trainer;
+  /// if true, use exchange_data_by_super_node, else use
+  /// exchange_data_by_sample; default if false
+  bool m_super_node = false;
+
+  /// size of a compacted conduit::Node that contains a single sample
+  int m_compacted_sample_size = 0;
```
ndryden: `size_t`?

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -278,18 +277,41 @@ protected :
...
   std::unordered_map<int, int> m_sample_sizes;
 
   /// used in set_conduit_node(...)
   std::mutex m_mutex;
 
-  /// used in exchange_data_by_sample, when sample sizes are non-uniform
-  bool m_have_sample_sizes;
+  /// Currently only used for imagenet. On return, 'sizes' maps a sample_id to image size, and indices[p] contains the sample_ids that P_p owns
+  /// for use in local cache mode
+  void get_image_sizes(std::unordered_map<int,int> &sizes, std::vector<std::vector<int>> &indices);
```
ndryden: `size_t`?

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -278,18 +277,41 @@ protected :
...
+
+  /// offset at which the raw image will be stored in a shared memory segment;
+  /// for use in local cache mode; maps data_id to offset
+  std::unordered_map<int,size_t> m_image_offsets;
+  /// fills in m_image_offsets for use in local cache mode
+  void compute_image_offsets(std::unordered_map<int,int> &sizes, std::vector<std::vector<int>> &indices);
```
ndryden: `size_t`?

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -278,18 +277,41 @@ protected :
...
+
+  /// for use in local cache mode
+  void allocate_shared_segment(std::unordered_map<int,int> &sizes, std::vector<std::vector<int>> &indices);
```
ndryden: `size_t`?

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -278,18 +277,41 @@ protected :
...
+
+  /// for use in local cache mode
+  void read_files(std::vector<char> &work, std::unordered_map<int,int> &sizes, std::vector<int> &indices);
```
ndryden: `size_t`?

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -278,18 +277,41 @@ protected :
...
+
+  /// for use in local cache mode
+  void build_conduit_nodes(std::unordered_map<int,int> &sizes);
```
ndryden: `size_t`?

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -278,18 +277,41 @@ protected :
...
+
+  /// for use in local cache mode
+  void exchange_images(std::vector<char> &work, std::unordered_map<int,int> &image_sizes, std::vector<std::vector<int>> &indices);
```
ndryden: `size_t`?

davidHysom:
> Just some suggestions on further `int` -> `size_t` changes to make sure things stay robust to large samples/datasets.

I'll make these changes in a subsequent PR.

PR #1056, opened 2019-05-24
revision of data_reader_jag_conduit to handle hydra data

davidHysom:
initial work on data_reader_hydra_conduit; nothing here yet


davidHysom:
This PR has been divided into two PRs: 1063, 1067. Note that 1067 depends on 1063.

PR #1063, opened 2019-06-03
refactored data_reader_jag_conduit into base and derived class

davidHysom:
base class: data_reader_conduit
  derived class: data_reader_jag_conduit
The derived class contains those functions that depend on conduit data types:
  using ch_t = float;
  using conduit_ch_t = conduit::float32_array;
  using scalar_t = double;
  using input_t = double;
The reason is that the data types for hydra and JAG data may differ
(they currently differ for ch_t and conduit_ch_t, which are double
for the hydra data).
A next step is to convert data_reader_jag_conduit into a templated class,
where the above "using" statements become template parameters.

### DIFF
--- a/include/lbann/data_readers/data_reader_conduit.hpp
+++ b/include/lbann/data_readers/data_reader_conduit.hpp
@@ -1,0 +1,572 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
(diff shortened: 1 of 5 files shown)


samadejacobs:
If I understand what is going here, this PR moves code from _jag_conduit_ to a new base class _conduit_ so as to support broad applications. Will there be a different class for Hydra data? If not (since the only difference is data type), then I would rename the _jag_conduit_ class to say _icf_conduit_ or something else. Also, if template question is already taken care of, then may be that PR should override this. My comments are more general or cosmetic issues, I defer correctness to Jae-Seung and Brian.

davidHysom:
I like icf_conduit.
There will either be two derived classes, for JAG and HYDRA data,
or what is now jag_conduit will become a template to support both.
I'm also investigating another approach, to have what is now
data_reader_jag_conduit support both JAG and HYDRA -- but
I need to study the conduit code before making a final decision.

davidHysom:
This PR is no longer applicable, due to other PRs and the manner in which the code base has evolved.

PR #1067, opened 2019-06-03
Data store conduit template

davidHysom:
This is a follow on to PR 1063. Templatizing data_store_jag_conduit since some of the "using" statements (e.g, using ch_t=float) differ from JAG to HYDRA data.


davidHysom:
This PR is no longer applicable, due to other PRs and the manner in which the code base has evolved.

PR #1070, opened 2019-06-05
Beta distributions

ndryden:
(Note: This depends on #1046.)

This adds a [`RandomNumberDistribution`](https://en.cppreference.com/w/cpp/named_req/RandomNumberDistribution)-compatible implementation of the Beta distribution. This will be needed for Mixup.

This uses two algorithms to generate values from the Beta distribution, depending on the distribution parameters. When `a <= 1 && b <= 1`, it uses Jöhnk's algorithm, or a variant of that using logarithms for more stability when `a` or `b` are small (adapted from the NumPy implementation). Otherwise, it generates them from the Gamma distribution (which is slower).

I would most like comments on the `fast_random_uniform` implementation in `random.hpp`, and the `generate` implementations in `beta.hpp`.

### DIFF
--- a/include/lbann/utils/beta.hpp
+++ b/include/lbann/utils/beta.hpp
@@ -1,0 +1,220 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+// may not use this file except in compliance with the License.  You may
+// obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the license.
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef LBANN_UTILS_BETA_HPP
+#define LBANN_UTILS_BETA_HPP
+
+#include <random>
+#include <ostream>
+#include <istream>
+#include <cmath>
+
+#include "lbann/utils/random.hpp"
+
+namespace lbann {
+
+/**
+ * Produces random floating point values drawn from a Beta distribution with
+ * parameters a > 0 and b > 0.
+ * See:
+ *     https://en.wikipedia.org/wiki/Beta_distribution
+ * for more details.
+ */
+template <typename RealType = double>
+class beta_distribution {
+public:
+  using result_type = RealType;
+
+  class param_type {
+  public:
+    using distribution_type = beta_distribution;
+
+    explicit param_type(RealType a, RealType b) :
+      m_a(a), m_b(b) {}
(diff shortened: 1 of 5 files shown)


include/lbann/utils/random.hpp
```diff
@@ -113,6 +113,86 @@ inline T fast_rand_int_pow2(Generator& g, T max) {
...
+
+namespace details {
+
+// See section on converting uint64_ts to doubles in:
+// http://xoshiro.di.unimi.it/
+
+template <typename Generator>
+inline float random_float_32(Generator& g) {
+  const uint32_t r = g() >> 9;
+  return r * (1.0f / 8388608.0f);
```
timmoon10: This is a cool trick. Super microoptimization (nanooptimization?): we should sample an integer from [0, 2^24). `float` has 23 mantissa bits plus an implicit leading bit.

include/lbann/utils/random.hpp
```diff
@@ -113,6 +113,86 @@ inline T fast_rand_int_pow2(Generator& g, T max) {
...
+}
+
+template <typename Generator>
+inline float random_float_64(Generator& g) {
+  const uint32_t r = uint32_t(g()) >> 9;  // Truncate.
```
timmoon10: Is there a reason we can't consolidate `random_float_32` and `random_float_64`? It seems to me that this implementation works fine if `g()` outputs either a `float` or `double`. If we're worried about an unnecessary constructor, switch to brace-initialization.

src/utils/unit_test/random_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+constexpr size_t num_tests = 1000;
+
+TEST_CASE("Testing fast_random_uniform", "[random][utilities]") {
+  SECTION("32-bit generator") {
+    std::mt19937 gen;
+    SECTION("floats") {
+      for (size_t i = 0; i < num_tests; ++i) {
+        float val = lbann::fast_random_uniform<float>(gen);
+        REQUIRE(val >= 0.0f);
+        REQUIRE(val <= 1.0f);
```
timmoon10:
```suggestion
        REQUIRE(val < 1.0f);
```
The other tests should be similarly changed.

include/lbann/utils/beta.hpp
```diff
@@ -0,0 +1,220 @@
...
+   * Note: There should be an umlaut on the "o" in "Johnk", but blame poor
+   * unicode support.
+   */
+  template <typename Generator>
+  result_type generate_johnk(Generator& g, result_type a, result_type b) {
+    while (true) {
+      const result_type U = fast_random_uniform<result_type>(g);
+      const result_type V = fast_random_uniform<result_type>(g);
+      const result_type X = std::pow(U, result_type(1) / a);
+      const result_type Y = std::pow(V, result_type(1) / b);
```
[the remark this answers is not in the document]
ndryden: The beta distribution requires `a > 0` and `b > 0`. This is documented, but we can check and enforce it.

include/lbann/utils/beta.hpp
```diff
@@ -0,0 +1,220 @@
...
+      return generate_johnk(g, m_params.a(), m_params.b());
+    } else {
+      return generate_gamma(g, m_gamma_a, m_gamma_b);
+    }
+  }
+  // Generator for when we use specified parameters.
+  template <typename Generator>
+  result_type generate(Generator& g, const param_type& p) {
+    if (p.a() <= result_type(1) && p.b() <= result_type(1)) {
+      return generate_johnk(p.a(), p.b());
```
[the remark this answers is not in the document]
ndryden: 🤦‍♂

include/lbann/utils/random.hpp
ndryden: I don't think any of the random number generators support generating 24-bit numbers, unless we want to generate it from three 8-bit numbers, which is likely slower.

include/lbann/utils/random.hpp
ndryden: It was mostly just copy-pasting from the `double` code. I don't think it's necessary.

include/lbann/utils/random.hpp
timmoon10:
```suggestion
  const uint32_t r = g() >> 8;
  return r * (1.0f / 16777216.0f);
```

Issue #1019, opened 2019-04-25
Validation data reader copies preprocessing


ndryden:
When creating a validation data reader from a percent of the training data, the created data reader is made by copying the training data reader. This also copies the associated preprocessing setup.

This is very likely not what we want, since it will typically apply the training data reader's random augmentation to the validation data, making the validation accuracy not representative of the test time accuracy.

PR #1082, opened 2019-06-15
Fix ImageNet validation

ndryden:
This disables using a random subset of data for validation in the ImageNet data reader prototext, and uses the real ImageNet validation dataset for validation (previously this was used for testing).

This avoids issues with #1019 that lead to misleading results.

@forsyth2 Is this going to cause any issues with the Bamboo testing?


forsyth2:
@ndryden I don't *think* that would cause issues for testing, but to be sure you can try running Bamboo on this branch.

PR #1083, opened 2019-06-16
Updated the spack environment to define the modules format

bvanessen:
Updated the spack environment to define the modules format for both
TCL and LMOD.  Simplified the instructions since the new modules don't
require you to unset the LIBRARY_PATH.  Removed the old spack recipes
build shell script, since it is deprecated and replaced by the
environments.


spack_environments/std_versions_and_variants_llnl_lc_cz.yaml
```diff
@@ -43,3 +43,47 @@
...
       version: [1.2.11]
+
+  modules::
+    enable::
+      - tcl
+      - lmod
+    lmod::
+      hash_length: 3
+      core_compilers:
+        - 'gcc@4.9.3'
```
benson31: Should you add `gcc@7.3.0` since we actually use that one?

spack_environments/std_versions_and_variants_llnl_lc_cz.yaml
```diff
@@ -43,3 +43,47 @@
...
+          '^python@3.7.2': python-3.7.2
+        filter:
+          # Exclude changes to any of these variables
+          environment_blacklist: ['CPATH', 'LIBRARY_PATH']
+      ^python:
+        autoload:  'direct'
+    tcl:
+      hash_length: 3
+      core_compilers:
+        - 'gcc@4.9.3'
```
benson31: Same question about `gcc@7.3.0`.

spack_environments/std_versions_and_variants_llnl_lc_cz.yaml
```diff
@@ -43,3 +43,47 @@
...
+      naming_scheme: '${PACKAGE}/${VERSION}-${COMPILERNAME}-${COMPILERVER}'
+      whitelist:
+        - gcc
+      blacklist:
+        - '%gcc@4.8'
```
benson31: We should blacklist GCC less than version 7...

docs/building_lbann.rst
```diff
@@ -226,7 +226,6 @@ Hydrogen, and LBANN separately, by whatever means they choose.
         spack install
         spack env loads # Spack creates a file named loads that has all of the correct modules
         source loads
```
benson31: Before this line, we should add `source ${SPACK_ROOT}/share/spack/setup-env.sh` because clean installations of spack (i.e., those with no modules in them) will not modify `MODULEPATH`. Running the setup script again should not have negative side effects for anyone and it will be necessary for new users who don't have any spack-installed packages.

spack_environments/std_versions_and_variants_llnl_lc_cz.yaml
bvanessen: Done.

spack_environments/std_versions_and_variants_llnl_lc_cz.yaml
bvanessen: Added gcc@4.9.3

docs/building_lbann.rst
bvanessen: Good insight.

bvanessen:
Note that once this PR is applied, it may be necessary to update your spack module files using the following command `spack module tcl refresh --delete-tree`

PR #1087, opened 2019-06-19
remove a cpp file from the root directory that isn't in any CMakeLists

benson31:
This file doesn't have to be removed but it should not stay here. It's not listed in any CMakeLists.txt file, so it's not built by LBANN. If it should be removed, just approve; if it should not be removed, let me know where to put it and I can move it.


benson31:
For reference, it seems to have appeared in d38c68da, but it might be an error since it nominally matches a file added in c16515ad that was removed in 4e3430f5.

PR #1088, opened 2019-06-19
Use shared memory for inter-process communication in Python data reader

timmoon10:
The Python data reader achieves parallelism using Python's `multiprocessing` module. This incurs significant overhead from inter-process communication, mostly in deserializing messages. We now avoid this entirely by maintaining an array in shared memory. Worker processes will populate the shared memory array during each mini-batch (using efficient copies with the buffer protocol) and the master process can simply memcpy from it once all the workers have finished. This is all internal and there are no user-visible changes.

My main test example was a [triplet Siamese network based on AlexNet](http://openaccess.thecvf.com/content_cvpr_2018/papers/Mundhenk_Improvements_to_Context_CVPR_2018_paper.pdf). It loads an ImageNet image from file, extracts several crops, and applies chroma blurring (converting to Lab color space, blurring, and returning to RGB). Running on 1 Pascal node with ImageNet-10, I got about 9.2 sec/epoch. For comparison, AlexNet with the native ImageNet data reader takes about 4.8 sec/epoch.

### DIFF
--- a/include/lbann/data_readers/data_reader_python.hpp
+++ b/include/lbann/data_readers/data_reader_python.hpp
@@ -161,10 +161,47 @@ protected:
   bool fetch_label(CPUMat& Y, int data_id, int mb_idx) override;
 
 private:
+
+  /** @brief Dimensions of data sample tensor. */
   std::vector<El::Int> m_sample_dims;
+  /** @brief Number of data samples in data set. */
   El::Int m_num_samples;
+
+  /** @brief User-provided Python function to access data samples.
+   *
+   *  The function is expected to take one integer argument for the
+   *  sample index. It must return an iterator that defines the
+   *  entries in a data sample.
+   */
   python::object m_sample_function;
+
+  /** @brief Wrapper function around sample access function.
+   *
+   *  This function is intended to facilitate inter-process
+   *  communication in the Python @c multiprocessing module. Data
+   *  samples are copied into a shared memory array.
+   *
+   *  @todo Performance optimizations for NumPy data.
+   */
+  python::object m_sample_function_wrapper;
+
+  /** @brief Pool of worker processes.
+   *
+   *  From the Python @c multiprocessing module.
+   */
   python::object m_process_pool;
+
+  /** @brief Shared memory array.
+   *
+   *  @c RawArray the Python @c multiprocessing module.
+   */
+  python::object m_shared_memory_array;
+
+  /** @brief Pointer into shared memory array.
+   *
+   *  Points to buffer for @c m_shared_memory_array.
+   */
+  DataType* m_shared_memory_array_ptr = nullptr;
 
 };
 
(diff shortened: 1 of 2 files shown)


timmoon10:
I notice that something hangs when I run AlexNet with the Python data reader. Perhaps we run out of shared memory when the data is large enough? I need to dig into this further.

When I extract a 100x100 crop, I get about 10.2 sec/epoch. The fact this is close to the Siamese network suggests to me that runtime is bottlenecked by file IO, although I'll need to profile to be sure.

timmoon10:
I figured out the hang (something in `torchvision.transforms.ToTensor`). AlexNet now successfully runs with about 12.2 sec/epoch. Learning behavior was similar to a run with the native ImageNet data reader. I'll go ahead and merge it now.

It bothers me that my ImageNet Python script is still outperformed this much by the native ImageNet data reader. It uses TurboJPEG and PyTorch's preprocessing transforms, and profiling doesn't show much room for improvement. I do notice that my ImageNet Python script takes ~7.5 ms/sample when I run it by itself, but takes ~11.5 ms/sample when I run it inside LBANN. Wild speculations:
- Context switching between processes is slower than between threads, so Python `multiprocessing` makes core contention more painful.
- The creation of worker processes creates more contention on hardware cores.
- VTune says the spin kernel in Aluminum's progress engine (`Al::internal::ProgressEngine::engine`) takes 84% of total CPU time. Perhaps it's more likely to crowd out other processes than threads in the same process.
- [OpenCV is faster than PIL](https://www.kaggle.com/vfdev5/pil-vs-opencv). I might get better results if I manually implement the preprocessing transforms in OpenCV rather than relying on PyTorch's PIL-based transforms.

ndryden:
@timmoon10 PyTorch is also fairly liberal about making internal copies in its preprocessing pipeline, last I looked. That could be another source of overhead.

Be careful about looking at CPU time for Aluminum's progress engine, since it spin-polls.

timmoon10:
The PyTorch operations didn't look especially egregious in my profiling of the script, in isolation or within LBANN. It seems more likely that the performance hit is from context switching overheads or excessive core contention.

ndryden:
Process context switching is very fast (relative to I/O), see e.g. [here](https://eli.thegreenplace.net/2018/measuring-context-switching-and-memory-overheads-for-linux-threads/). If we have core contention, that could impact things though. Presumably the processes Python spawns are not nicely bound with affinities.

timmoon10:
Hm, those are helpful benchmarks. The Python script takes 2.4ms per sample, which dwarfs the 22us cost of even launching a new process. I don't think we can immediately apply those benchmarks since they focus on thread context switching and don't include the effects of cache thrashing. Perhaps I should try running my script within `multiprocessing` and see if I get the same performance hit without LBANN.

Regarding affinity, the worker Python processes are cloned from the original LBANN process. As far as I can tell, they should keep the same CPU affinity.

PR #1091, opened 2019-06-19
Mixup preprocessing

ndryden:
This re-creates #1071, which fell victim to far too much rewriting of history, rebasing, cherry-picking, force-pushing, and other Things One Should Not Do™.

This adds support for mixup preprocessing, implemented as a callback. @timmoon10 and I discussed the best way to do this, and decided it fit better as a callback, rather than as a layer or transform.

I have done basic tests of this, but have not yet done training to convergence.

### DIFF
--- a/src/callbacks/callback_mixup.cpp
+++ b/src/callbacks/callback_mixup.cpp
@@ -1,0 +1,83 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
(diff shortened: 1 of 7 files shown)


src/callbacks/callback_mixup.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  for (El::Int i = 0; i < mbsize; ++i) {
+    const El::Int j = shuffled_indices[i];
+    if (i == j) {
+      continue;
+    }
+    float lambda = dist(gen);
+    lambda = std::max(lambda, 1.0f - lambda);
+    const float lambda_sub = 1.0f - lambda;
+    DataType* __restrict__ x1_buf = samples.Buffer() + i*samples.LDim();
```
timmoon10: This is fine, but there is a more convenient way:
```suggestion
    DataType* __restrict__ x1_buf = samples.Buffer(0, i);
```

src/callbacks/callback_mixup.cpp
```diff
@@ -0,0 +1,83 @@
...
+void callback_mixup::on_forward_prop_end(model *m, Layer *l) {
+  if (!m_layers.count(l->get_name())) {
+    return;
+  }
+  if (m->get_execution_mode() != execution_mode::training) {
+    return;  // No mixup outside of training.
+  }
+
+  auto& samples = l->get_local_activations(0);
+  auto& targets = l->get_local_activations(1);
```
timmoon10: I think `labels` is clearer.

PR #1100, opened 2019-06-28
Refactoring Python utility classes

timmoon10:
### The problem
In the process of debugging a unit test for the Python data reader, I encountered some bugs in the destructors of the Python helper classes. In particular, they don't acquire the GIL when destroying Python objects and don't handle the case where the Python session has already ended (where things may be destroyed in an arbitrary order). This caused exceptions at the end of the program, which made the unit test fail.

### The solution
I've refactored the Python utility classes to make them more robust, mostly by making sure to acquire the GIL and check whether Python has ended. I think they're converging to a stable form, so I've taken the liberty of moving them to a standalone utility header and expanding documentation. They should be helpful if we ever want to add more Python functionality (layers? preprocessing transforms? everything?). LBANN now exits cleanly with my unit test (forthcoming in a future PR).

### Overview of the utility classes
- `lbann::python::session`: A singleton class that manages the embedded Python session. The singleton instance is mostly used internally. It has some useful static member functions to check for Python errors, etc.
- `lbann::python::global_interpreter_lock`: A RAII wrapper around the GIL. It acquires the GIL in its constructor and releases it in the destructor. Users should instantiate this class whenever they call Python C API functions.
- `lbann::python::object`: A wrapper around Python objects. The Python C API forces you to manually manage reference counters, so this class handles most cases automatically. However, you still need to be a little careful since some Python C API functions treat the reference counters in a non-standard way.


src/utils/python.cpp
```diff
@@ -0,0 +1,221 @@
...
+// ---------------------------------------------
+// session class
+// ---------------------------------------------
+
+void session::start_once() { get(); }
+
+bool session::is_active() noexcept { return Py_IsInitialized(); }
+
+void session::check_error(bool force_error) {
+  start_once();
```
ndryden: Is this redundant, since `global_interpreter_lock`'s constructor also does this?

timmoon10: That's true. I suppose it's a reasonable assumption that the session is active if you hold the GIL.

PR #1104, opened 2019-07-01
Update checkpointing

forsyth2:
- Update persisting of checkpoint test output.
- Remove FIXMEs from lbann2 checkpoint tests.


forsyth2:
Some checkpoint tests are still failing, but merging these changes anyway, since they make the code easier to follow.

PR #1105, opened 2019-07-03
Update unit tests to use Python front end

forsyth2:
Converting unit tests to use the Python front-end instead of the internal `tools.get_command`.


bamboo/unit_tests/test_unit_mnist_ridge_regression.py
```diff
@@ -1,26 +1,113 @@
...
+    sys.path.append(build_path)
+
+    # Model
+    # Converted from lbann/model_zoo/tests/model_mnist_ridge_regression.prototext.
+    # Equivalent to prototext's "Layers" section.
+    import lbann
+    input_node = lbann.Input()
+    images_node = lbann.Identity(input_node)
+    image_labels_node = lbann.Identity(input_node)
+    fc_node = lbann.FullyConnected(images_node)
```
timmoon10:
```suggestion
    fc_node = lbann.FullyConnected(images_node, num_neurons=10, has_bias=True)
```

PR #1109, opened 2019-07-10
Cleanup exception handling

benson31:
This cleans up a bunch of warnings about catching polymorphic exception types by value. Generally, exceptions should be caught by const-reference unless there's a compelling reason not to do that (not typing "SHIFT+7" is not a compelling reason).

I also swept through the files that were afflicted by catch-by-value and cleaned up a few other things. Notably, `lbann::exception` is-a `std::exception`, so code like this:
```cpp
try {...}
catch (const exception& e) {
    El::ReportException(e);
    continue;
} catch (const std::exception& e) {
    El::ReportException(e);
    continue;
}
```
is just redundant since nothing special happens for the `lbann::exception` that doesn't happen for the `std::exception` and vice versa.

I also changed the "const-reference style" to be more consistent with the rest of LBANN, which prefers `const type&`.

### DIFF
--- a/include/lbann/utils/opencv.hpp
+++ b/include/lbann/utils/opencv.hpp
@@ -38,7 +38,7 @@ namespace utils {
  * Check whether data is an image.
  * Currently requires data to be a uint8_t CPUMat, with 3 dimensions, the first
  * (channel) being 1 or 3.
- * 
+ *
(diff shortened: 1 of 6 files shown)


benson31:
@ndryden No. :keyboard:

PR #1112, opened 2019-07-16
Update time limits

forsyth2:
Nightly Develop was hitting the time limit of 900 minutes (15 hours), so increasing to 1440 minutes (24 hours).

### DIFF
--- a/bamboo/allocate_and_run.sh
+++ b/bamboo/allocate_and_run.sh
@@ -31,7 +31,7 @@ if [ "${CLUSTER}" = 'pascal' ]; then
 fi
 
 if [ ${WEEKLY} -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


davidHysom:
Wow - we're using 16 nodes for 24 hours every day? Seems like a lot. I'd like to check on how many computing resources other projects use for nightly testing.

forsyth2:
@davidHysom Well, the tests will give up the nodes when they finish. I don't expect the tests to actually take 24 hours. They were staying under 15 hours up until Build 554. That build ran off three new commits (the ones from Tuesday 7/9 - see https://github.com/LLNL/lbann/commits/develop). I suppose it's possible one of those commits actually caused the issue with timing; I assumed the issue to be something on the hardware side.

PR #1114, opened 2019-07-16
Fuse batchnorm allreduces in FP/BP

ndryden:
This replaces the separate mean and variance (and associated gradients) buffers with fused buffers. This enables the allreduces during forward/backprop when doing node-local/global statistics to be fused into a single allreduce. Since the size of these buffers is the number of channels in the layer being normalized, these allreduces are likely latency-bound, so doing fewer of them improves performance.

Tested that it could train Wide ResNet-50-4 with node-local stats to convergence without issue. Also tested gradient checking.

Benchmarked with global stats at small scale and observed a minor but consistent performance improvement.

### DIFF
--- a/include/lbann/layers/regularizers/batch_normalization.hpp
+++ b/include/lbann/layers/regularizers/batch_normalization.hpp
@@ -71,14 +71,24 @@ private:
    */
   std::unordered_map<El::Int, El::Int> m_num_per_sum_cache;
 
-  /** Current minibatch means. */
-  std::unique_ptr<AbsDistMat> m_mean;
-  /** Current minibatch standard deviations. */
-  std::unique_ptr<AbsDistMat> m_var;
-  /** Gradient w.r.t. means. */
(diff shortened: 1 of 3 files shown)


ndryden:
I actually thought about that, but it doesn't work without more significant changes. The basic problem is that the views would be a `CPUMat` or a `GPUMat` (since H doesn't support `AbsMat` views), but our non-training code path expects `AbsMat&`s.

timmoon10:
Ah... unfortunately that's true. Ok, this is the right approach then.

bvanessen:
Sounds great.


Brian C. Van Essen

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Jul 17, 2019, at 4:04 PM, Tim Moon <notifications@github.com<mailto:notifications@github.com>> wrote:


Ah... unfortunately that's true. Ok, this is the right approach then.

PR #1116, opened 2019-07-18
Fixing the warnings raised by latest clang

benson31:
Nothing major here, just slight cleanup.

The major points:

- No universal references returned by name.
- Mark everything that overrides `override`.
- Don't have `const` data members and defaulted copy assignment (can't assign `const` members).
- Probably don't have `const` data members.

### DIFF
--- a/include/lbann/callbacks/callback_checksmall.hpp
+++ b/include/lbann/callbacks/callback_checksmall.hpp
@@ -49,7 +49,7 @@ class lbann_callback_checksmall : public lbann_callback {
   lbann_callback_checksmall() : lbann_callback() {}
   lbann_callback_checksmall(const lbann_callback_checksmall&) = default;
   lbann_callback_checksmall& operator=(
-    const lbann_callback_checksmall&) = default;
+    const lbann_callback_checksmall&) = delete;
   lbann_callback_checksmall* copy() const override {
     return new lbann_callback_checksmall(*this);
   }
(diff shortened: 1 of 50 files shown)


include/lbann/callbacks/callback_checksmall.hpp
```diff
@@ -49,7 +49,7 @@ class lbann_callback_checksmall : public lbann_callback {
   lbann_callback_checksmall() : lbann_callback() {}
   lbann_callback_checksmall(const lbann_callback_checksmall&) = default;
   lbann_callback_checksmall& operator=(
-    const lbann_callback_checksmall&) = default;
+    const lbann_callback_checksmall&) = delete;
```
timmoon10: I'd prefer making `m_threshold` a `static const` or `static constexpr` variable. That way we can keep the copy-assignment.

include/lbann/callbacks/callback_variable_minibatch.hpp
```diff
@@ -44,7 +44,7 @@ class lbann_callback_variable_minibatch : public lbann_callback {
   lbann_callback_variable_minibatch(
     const lbann_callback_variable_minibatch&) = default;
   lbann_callback_variable_minibatch& operator=(
-    const lbann_callback_variable_minibatch&) = default;
+    const lbann_callback_variable_minibatch&) = delete;
```
timmoon10: I'd prefer making `m_starting_mbsize` a non-`const` variable.

PR #1119, opened 2019-07-22
Grouped batchnorm statistics

ndryden:
This makes the specification of batchnorm statistics aggregation more general. Instead of just supporting local, node-local, and global statistics, you now specify the number of processors to aggregate statistics over. This makes rather more sense than just having a limited set of options, so you can aggregate over as many processors as needed to get good statistics, and not more.

For backwards compatibility, this still supports the old method of specifying aggregation in the prototext, but will issue a warning about it. The Python ResNet modules have been switched to exclusively use the new method, but the `model_zoo/vision/resnet.py` still supports the old argument for backwards compatibility (except node-local statistics).

### DIFF
--- a/model_zoo/vision/resnet.py
+++ b/model_zoo/vision/resnet.py
@@ -38,9 +38,13 @@ parser.add_argument(
     '--block-channels', action='store', default=None, type=str,
     help='Internal channels in each ResNet block (comma-separated list)')
 parser.add_argument(
-    '--bn-stats-aggregation', action='store', default='local', type=str,
+    '--bn-stats-aggregation', action='store', type=str,
     help=('aggregation mode for batch normalization statistics '
-          '(default: "local")'))
+          '(default: "local") (DEPRECATED)'))
+parser.add_argument(
+    '--bn-statistics-group-size', action='store', default=1, type=int,
+    help=('Group size for aggregating batch normalization statistics '
+          '(default: 1)'))
 parser.add_argument(
     '--warmup', action='store_true', help='use a linear warmup')
 parser.add_argument(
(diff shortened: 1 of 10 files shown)


src/proto/lbann.proto
```diff
@@ -1060,7 +1060,9 @@ message BatchNormalization {
   double scale_init = 2;     //default: 1.0
   double bias_init = 3;      //default: 0.0
   double epsilon = 4;        //default: 1e-5
-  string stats_aggregation = 5; // default: local
+  string stats_aggregation = 5; // default: local; deprecated
+  int64 statistics_group_size = 6;  // default: 1 (local)
+  bool global_statistics = 7;
```
timmoon10: I get a bad feeling from having two counteracting options. Perhaps a better approach would be  to make a zero value correspond to local statistics and a negative value to global statistics.

model_zoo/vision/resnet.py
```diff
@@ -38,9 +38,13 @@
     '--block-channels', action='store', default=None, type=str,
     help='Internal channels in each ResNet block (comma-separated list)')
 parser.add_argument(
-    '--bn-stats-aggregation', action='store', default='local', type=str,
+    '--bn-stats-aggregation', action='store', type=str,
```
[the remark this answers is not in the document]
ndryden: I'm not sure, but I'll leave it for now and remove it when we remove the old option from `lbann.proto`. Just in case.

src/proto/lbann.proto
ndryden: Good point. I've made this change. (The Python front-end still uses `0` for global statistics.)

timmoon10: We should update the Python front-end. It's much easier to change the API before it's exposed to users.

ndryden: The Python frontend is doing the same thing as LBANN. The only hiccup is the prototext, where I would prefer the default to be local statistics, but `0` is the default value for an integer field.

timmoon10: Should we change the convention everywhere so that a zero group size corresponds to local statistics? Having two different usages seems like it could breed confusion.

ndryden: I don't like that, because a group size of 1 would also correspond to local statistics.

timmoon10: Looks like our code smell senses are in disagreement. In any case, I don't think this is a very big deal.

Issue #618, opened 2018-09-15
Batch norm optimizations


timmoon10:
I have two unrelated but overlapping optimizations for batch norm.

### Fused scale-bias weights
In Resnet-50, approximately 2/3 of the backprop allreduces originate from batch norm. Merging the scale and bias weights in batch norm would thus reduce the number of backprop allreduces by 1/3 and move each allreduce very slightly away from the latency-dominated regime. The resulting fused weights are somewhat less interpretable and more cumbersome, but the performance boost is probably worth it. This is an easy change to make.

### Separating the affine transformation into a distinct layer
The original batch norm paper includes a normalization step (subtract mean and divide by variance) and an affine step (apply scale and bias). These are logically discrete steps and many models apply the normalization step without the affine step (for instance, [Inception v3](https://github.com/keras-team/keras-applications/blob/master/keras_applications/inception_v3.py) applies a bias without scaling). TensorFlow, PyTorch, and Caffe make the affine step optional and Caffe removes it entirely. We currently support this if the user manually sets the batch norm weights with a scale of one, bias of zero, and no optimizer, but this is very unwieldy and we still do some unnecessary computation. I propose we follow Caffe's approach and implement a separate affine layer (with fused scale and bias weights). This will interact very naturally with our Keras frontend (the affine step option will just insert an affine layer after the batch norm), but will requiring modifying all prototext models that contain batch norm.

Pinging @ndryden and @JaeseungYeom.

ndryden:
Another optimization we should consider: Since we have custom implementations of both batchnorm and ReLU, we should support fusing them into a single kernel.

ndryden:
After recent discussion, there's a performance need for both fusing the weights and to fuse the allreduces done in intra-node/global batchnorm.

PR #1122, opened 2019-07-23
Channel-wise scale/bias layer

timmoon10:
The channel-wise scale/bias layer is equivalent to the second stage of batchnorm. It maintains the scaling and bias terms in a single weights tensor, which has the effect of halving the number of gradient allreduces (mostly latency-bound) and shifting each gradient allreduce slightly toward the bandwidth-bound regime.

It passes metric checking and gradient checking on Pascal and Catalyst. However, I'm not entirely pleased with the name. I was thinking "channel-wise affine layer", but that's not quite accurate since it implies an affine transformation on each channel. Also, DeepAI uses the term "affine layer" to refer to what we call a "fully-connected layer".

This makes progress toward Issue #618.

### DIFF
--- a/include/lbann/layers/learning/channelwise_scale_bias.hpp
+++ b/include/lbann/layers/learning/channelwise_scale_bias.hpp
@@ -33,6 +33,25 @@
 
 namespace lbann {
 
+/** @brief Apply scale and bias to tensor channels.
+ *
+ *  The input tensor is sliced along the first tensor dimension (the
+ *  "channel" dimension, assuming image data in CHW format) and scale
+ *  and bias terms are applied independently to each slice. More
+ *  precisely, given input and output tensors
+ *  @f$ X,Y\in\mathbb{R}^{d_1\times\cdots\times d_n} @f$
+ *  and scale and bias vectors @f$ a,b\in\mathbb{R}^{d_1} @f$:
+ *  @f[
+ *    Y_{i,j,\cdots} = a_i X_{i,j,\cdots} + b_i
+ *  @f]
+ *
+ *  The scale and bias vectors are fused into a single weights tensor
+ *  to reduce the number of gradient allreduces during backprop. In
+ *  particular, the weights tensor is a
+ *  @f$ \text{num_channels} \times 2 @f$ matrix, where the first
+ *  column correspond to scale terms and the second column to bias
+ *  terms.
+ */
 template <data_layout Layout = data_layout::DATA_PARALLEL,
           El::Device Device = El::Device::CPU>
 class channelwise_scale_bias_layer : public Layer {
@@ -104,6 +123,7 @@ public:
     m_weights[0]->set_matrix_distribution(matrix_dist);
 
     // Setup gradient w.r.t. weights
+    m_weights_gradient->AlignWith(matrix_dist);
     m_weights_gradient->Resize(num_channels, 2);
 
   }
@@ -114,6 +134,7 @@ protected:
 
 private:
 
+  /** Objective function gradient w.r.t. weights. */
   std::unique_ptr<AbsDistMat> m_weights_gradient;
 
 };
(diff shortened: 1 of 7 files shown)


include/lbann/layers/learning/channelwise_scale_bias.hpp
```diff
@@ -0,0 +1,144 @@
...
+ *
+ *  The input tensor is sliced along the first tensor dimension (the
+ *  "channel" dimension, assuming image data in CHW format) and scale
+ *  and bias terms are applied independently to each slice. More
+ *  precisely, given input and output tensors
+ *  @f$ X,Y\in\mathbb{R}^{d_1\times\cdots\times d_n} @f$
+ *  and scale and bias vectors @f$ a,b\in\mathbb{R}^{d_1} @f$:
+ *  @f[
+ *    Y_{i,j,\cdots} = a_i X_{i,j,\cdots} + b_i
+ *  @f]
```
ndryden: It may be helpful to users to explicitly note the relationship to batchnorm here.

include/lbann/layers/learning/channelwise_scale_bias.hpp
```diff
@@ -0,0 +1,144 @@
...
+ *
+ *  The scale and bias vectors are fused into a single weights tensor
+ *  to reduce the number of gradient allreduces during backprop. In
+ *  particular, the weights tensor is a
+ *  @f$ \text{num_channels} \times 2 @f$ matrix, where the first
+ *  column correspond to scale terms and the second column to bias
+ *  terms.
```
ndryden: Does this imply that we cannot initialize the scale and bias separately?

src/layers/learning/channelwise_scale_bias.cpp
```diff
@@ -0,0 +1,136 @@
...
+  auto& local_output = dynamic_cast<CPUMat&>(get_local_activations());
+  const auto& local_weights = dynamic_cast<const CPUMat&>(m_weights[0]->get_values().LockedMatrix());
+  const auto local_scale = El::LockedView(local_weights,
+                                          El::ALL, El::IR(0));
+  const auto local_bias = El::LockedView(local_weights,
+                                         El::ALL, El::IR(1));
+
+  // Dimensions
+  // Note: channel_size is the number of input entries per channel and
+  // local_width is the number of local mini-batch samples.
```
ndryden: Maybe we should just name the variable `local_mini_batch_size` then?

src/layers/learning/channelwise_scale_bias.cpp
```diff
@@ -0,0 +1,136 @@
...
+  const El::Int channel_size = std::accumulate(dims.begin() + 1,
+                                               dims.end(),
+                                               1, std::multiplies<int>());
+  const El::Int local_width = local_input.Width();
+
+  // Compute gradients
+  LBANN_OMP_PARALLEL_FOR
+  for (El::Int channel = 0; channel < num_channels; ++channel) {
+    const auto a = local_scale(channel, 0);
+    DataType da{0}, db{0};
```
ndryden: Why not accumulate directly into `local_gradient_wrt_scale`/`bias`?

include/lbann/layers/learning/channelwise_scale_bias.hpp
timmoon10: Yes. The default initialization is a scale of 1 and a bias of 0. Users must use the value initializer if they want something more intricate.

src/layers/learning/channelwise_scale_bias.cpp
timmoon10: I don't think there's much difference. For me this does a better job of conveying the intuition that we're iterating through the rows and columns of a matrix.

src/layers/learning/channelwise_scale_bias.cpp
timmoon10: I don't expect there to be a significant difference.

PR #1126, opened 2019-07-25
Add parse list test

benson31:
Adds the tests I should have added as part of #1118.

### DIFF
--- a/src/proto/unit_test/parse_list_test.cpp
+++ b/src/proto/unit_test/parse_list_test.cpp
@@ -42,7 +42,7 @@ TEST_CASE("Testing parse_list", "[proto][utilities]")
 
   SECTION("int")
   {
-    const std::vector<int> expected = { 1, 2, 3, 4, 5};
+    const std::vector<int> expected = { 1, 2, 3, 4, 5 };
     auto const answer =
       lbann::parse_list<int>("1 2 3 4 5");
     CHECK(answer == expected);
--- a/src/proto/unit_test/parse_list_test.cpp
+++ b/src/proto/unit_test/parse_list_test.cpp
@@ -1,0 +1,52 @@
+#include <catch2/catch.hpp>
+
+#include <lbann/base.hpp>
+#include <lbann/proto/proto_common.hpp>
+
+#include <string>
+#include <vector>
+
+TEST_CASE("Testing parse_list", "[proto][utilities]")
+{
+  SECTION("execution_mode")
+  {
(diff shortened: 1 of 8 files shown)


src/proto/unit_test/parse_list_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+    CHECK(answer == expected);
+    CHECK(
+      lbann::parse_list<lbann::execution_mode>("")
+      == std::vector<lbann::execution_mode>{});
+    CHECK(
+      lbann::parse_list<lbann::execution_mode>(" ")
+      == std::vector<lbann::execution_mode>{});
+
+    CHECK_THROWS(
+      lbann::parse_list<lbann::execution_mode>("banana tuna salad"));
```
timmoon10: Sounds gross.

src/proto/unit_test/parse_list_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+      lbann::parse_list<std::string>("this is a test");
+    CHECK(answer == expected);
+    CHECK(
+      lbann::parse_list<std::string>("") == std::vector<std::string>{});
+
+  }
+
+  SECTION("int")
+  {
+    const std::vector<int> expected = { 1, 2, 3, 4, 5};
```
timmoon10:
```suggestion
    const std::vector<int> expected = { 1, 2, 3, 4, 5 };
```

src/proto/unit_test/parse_set_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+    const std::set<std::string> expected = { "this", "is", "a", "test" };
+    auto const answer =
+      lbann::parse_set<std::string>("this is a test");
+    CHECK(answer == expected);
+    CHECK(lbann::parse_set<std::string>("") == std::set<std::string>{});
+  }
+
+  SECTION("int")
+  {
+    const std::set<int> expected = { 1, 2, 3, 4, 5};
```
timmoon10:
```suggestion
    const std::set<int> expected = { 1, 2, 3, 4, 5 };
```

src/proto/unit_test/parse_list_test.cpp
benson31: Yep, that's why I picked it. Should I
```cpp
#define CHECK_THROWS_UP(...) \
    CHECK_THROWS(__VA_ARGS__)
```
to help drive the point home? 😛

PR #1133, opened 2019-07-31
Split lbann.proto into multiple files

benson31:
This splits the various components of the protobuf interface into separate `.proto` files. The reasoning here is that LBANN should move to an "include what you use" paradigm; within that, we should include *only* what we use. Therefore, it will be beneficial if the various components are separated so, e.g., if a new weights initializer is added, we don't need to recompile all the layers, and so on. This will cause no externally visible changes.

I was told that motif-related things are not used and could be removed. Since code that could be removed should be removed, it's been removed.


benson31:
Most of the changes are the result of more IWYU-ing (mostly because I deleted one `#include "lbann/lbann.hpp"`).

This should be good to go now.

benson31:
The C++ side is solid, but I'm having an issue with the python FE with non-`lenet.py` models. It has something to do with class generation. I'm working on fixing it now. That change will the be merged up to dependent PRs...

python/lbann/util/class_generator.py
```diff
@@ -96,7 +96,11 @@ def export_proto(self):
             message = getattr(proto, base_field_name)
             message.SetInParent()
         else:
-            proto = getattr(lbann_pb2, message_name)()
+            proto_modules = set([callbacks_pb2, layers_pb2, metrics_pb2, model_pb2, objective_functions_pb2, optimizers_pb2, weights_pb2])
```
ndryden: Maybe include a comment indicating that this needs to be updated if we add additional protobuf definition files.

benson31: Done.

benson31:
Ok, this now works with the python frontend. I have updated the rest of the model zoo prototext models. The following python files likely need updating. Are they still used @samadejacobs:
```model_zoo/models/gan/jags/cycle_gan/generate_cycgan_m1.py
model_zoo/models/gan/jags/cycle_gan/generate_cycgan_m2.py
model_zoo/models/gan/jags/cycle_gan/generate_cycgan_m3.py
model_zoo/models/jag/gan/cyclic/generate_model.py
model_zoo/models/jag/gan/vanilla/generate_gan.py
```

python/lbann/util/class_generator.py
```diff
@@ -96,7 +96,18 @@ def export_proto(self):
...
         else:
-            proto = getattr(lbann_pb2, message_name)()
+            # TODO (trb 08/01/2019): This list would have to be
+            # updated any time another _pb2 file is created. It might
+            # be better to have this as a global `frozenset`
+            # (ndryden's suggestion) that gets maintained
+            # elsewhere. But this code either works or doesn't get
+            # executed now, so I vote delaying this fix until a need
+            # arises.
+            proto_modules = set([callbacks_pb2, layers_pb2, metrics_pb2, model_pb2, objective_functions_pb2, optimizers_pb2, weights_pb2])
```
timmoon10: It may be safer for this to be a list, just to make sure that the loop is deterministic.

src/proto/layers.proto
```diff
@@ -0,0 +1,572 @@
...
+  bool num_neurons_from_data_reader = 53;
+  bool freeze = 5;
+  string hint_layer = 56;
+
+  repeated WeightsData weights_data = 153;
+  string top = 154;
+  string bottom = 155;
+  string type = 156;
+
+  // a Layer should contain exactly one of the following
```
timmoon10: This comment can be removed.

src/proto/metrics.proto
```diff
@@ -0,0 +1,40 @@
...
+message Metric {
+
+  message LayerMetric {
+    string layer = 1;
+    string name = 2;
+    string unit = 3;
+  }
+
+  LayerMetric layer_metric = 11;
+}
```
timmoon10: `\n`

src/proto/optimizers.proto
```diff
@@ -0,0 +1,71 @@
...
+    double decay_rate = 2;
+    double eps = 3;           // Suggested: 1e-8
+  }
+
+  message SGD {
+    double learn_rate = 1;
+    double momentum = 2;      // Set to zero for vanilla SGD
+    bool nesterov = 4;
+  }
+}
```
timmoon10: `\n`

src/proto/weights.proto
```diff
@@ -0,0 +1,74 @@
...
+    double mean = 1;
+    double standard_deviation = 2;
+  }
+  message GlorotNormalInitializer {}
+  message GlorotUniformInitializer {}
+  message HeNormalInitializer {}
+  message HeUniformInitializer {}
+  message LeCunNormalInitializer {}
+  message LeCunUniformInitializer {}
+}
```
timmoon10: `\n`

benson31: If only we had this much extraneous whitespace in the C++... 😝

python/lbann/util/class_generator.py
benson31: Why would the loop not be deterministic? `pop()` will eventually kill the set and you'll get an exception.

timmoon10: I'm thinking about an edge case where two messages have the same name and different processes store the set entries in a different order.

benson31: We might want to think of ways to do this using the `full_name` or something to that effect. You're right that naming collisions are a possible issue; a list only ensure that if the wrong one is chosen, it will always be chosen. The better solution would make this order-invariant; let's leave that for a future PR.

benson31:
The update in `class_generator.py` has been tested and seems ok (I don't know why it wouldn't have been, but it's good to check these things). The only other changes in https://github.com/LLNL/lbann/pull/1133/commits/a6f797fa00ea68a75514fb1b379d36f6c76fcd72 are removing comments and adjusting whitespace.

python/lbann/util/class_generator.py
ndryden: Just to add: It is almost certainly the case that different processes will store the set entries in different orders. Python sets hash the object, and hashing is non-deterministic across Python invocations.

PR #1135, opened 2019-07-31
fix an issue building with NVPROF support after callback refactorying

benson31:
...the title says it all.

### DIFF
--- a/include/lbann/callbacks/sync_selected.hpp
+++ b/include/lbann/callbacks/sync_selected.hpp
@@ -63,12 +63,35 @@ class sync_selected : public sync_layers {
    * @param async_mpi sets not to synchronize mpi. The default is false.
    */
   sync_selected(const layers_t& layers,
-                               bool async_gpus = false, bool async_mpi = false);
+                bool async_gpus = false, bool async_mpi = false);
 
-  sync_selected(const sync_selected&) = default;
+  sync_selected(const sync_selected& other)
+    : sync_layers(other),
+      m_cuda_prof_params(
+        other.m_cuda_prof_params
+        ? make_unique<cuda_profiler_parameters>(*m_cuda_prof_params)
+        : nullptr),
+      m_layers(other.m_layers),
+      m_fwd_ptrs(other.m_fwd_ptrs),
+      m_bwd_ptrs(other.m_bwd_ptrs),
+      m_both_ptrs(other.m_both_ptrs),
+      m_all_set(other.m_all_set)
+  {}
 
-  sync_selected& operator=(
-    const sync_selected&) = default;
+  sync_selected& operator=(const sync_selected& other) {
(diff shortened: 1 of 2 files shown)


src/callbacks/sync_selected.cpp
```diff
@@ -153,13 +169,14 @@ void sync_selected::init_cuda_profiler(
 }
 
 void sync_selected::setup(model *m) {
+  init_cuda_profiler(*(m->get_comm()));
```
timmoon10: This changes the behavior to start nvprof during the callback setup, instead of at construction. This may cause the profiler to miss out on setup in other objects, like layers. @naoyam @ndryden Is this behavior change OK?

include/lbann/callbacks/sync_selected.hpp
```diff
@@ -116,6 +143,29 @@ class sync_selected : public sync_layers {
...
+                             const std::string& out_dir,
+                             int out_mode)
+      : m_cfg_file(cfg_file),
+        m_out_dir(out_dir),
+        m_out_mode(out_mode)
+    {}
+  };// cuda_profiler_parameters
+
+  /** @brief The parameters for the CUDA profiler */
+  std::unique_ptr<cuda_profiler_parameters> m_cuda_prof_params;
```
timmoon10: Why do we store this in a `unique_ptr`? I see later that we check whether it is `nullptr`, in which case we treat it as uninitialized and skip some stuff. Couldn't we do the same thing by setting a `bool`?

benson31: 1. Pointers are convertible to `bool`.
2. By using heap allocation, you don't unnecessarily allocate space for two unused strings and an `int`, in addition to the likely 4-8 extra bytes for the `bool`. Since the data is optionally provided, it makes sense to just throw it on the heap somewhere.

src/callbacks/sync_selected.cpp
benson31: If this is the wrong fix, that's fine. But this needs to be fixed somehow. There is no `comm` available during factory time (and it will be a really hard sell to convince me that this one outlier is worth the baggage) and there's a `barrier` in there so it seems to matter for this. `setup` is the first time a callback sees a `comm` (via the `model`), so it seemed a logical place. If NVPROF should be initialized globally, perhaps moving this up the stack is the best answer. Let me know.

ndryden:
(I know I'm no longer assigned, but I was summoned, and this PR reached the top of the queue.)

1. Yes, this is ugly. Note especially that we have another profiling callback too. Things should be cleaned up.
2. It is sometimes important to have nvprof available from the beginning (e.g. when profiling initialization), but that is not the typical thing. Note that we can just use nvprof command line flags to manage whether it is on from the beginning (the default) or needs to be explicitly enabled.

benson31:
Closing; callback will be removed instead in a future PR.

PR #1136, opened 2019-07-31
Sample list for image data reader

JaeseungYeom:
PR #1401 supersedes this PR
- Currently, sample list loading is interleaved among processes per trainer.
  This I believe is to distribute the load while processing a single text input file in case that the load is non trivial.
 - However, data store pre-loads the label as long as the image file name without interleaving. Thus, the sample order in the sample list does not match the label order in the data store.
- This can be solved by reordering sample list after all_gather to make sure the order is as it is in the file. While this capability is added, another solution is used. **`m_labels`, which maps a sample name (file name) to its label replaces the list that contains the pair of image file name and label, `m_image_list`.**
- Currently, to express the location of sample list file in the prototext input, `data_reader_jag_conduit` obtains the filename from `index_list`, and replaces the directory path with the one for the data files which is given by `data_filedir`. However, this approach forces users to put the sample list in the same directory as where data are and create mess. T**o allows users to put sample list anywhere they want, I modified to use the full path of index_list.**
- The sample lists for imagenet are created under `/p/lustre2/brainusr/datasets/ILSVRC2012/sample_list`.  `*_sample_list.txt` is in the sample list format. `*_image_list.txt` is in the original format with the directory paths corrected from `lscratchf` to `lustre2`.

- verified by two methods
  - writing out sample list file and comparing it against the input file
  - compare the image file name obtained from m_sample_list and m_image_list in fetch_datum() over multiple epochs.

### DIFF
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -101,6 +101,7 @@ class generic_data_reader {
     m_procs_per_partition(1),
     m_io_thread_pool(nullptr),
     m_jag_partitioned(false),
+    m_keep_sample_order(false),
     m_model(nullptr),
     m_issue_warning(true)
   {}
@@ -172,6 +173,12 @@ class generic_data_reader {
    * Returns the complete sample list for your data set.
    */
   std::string get_data_sample_list() const;
+
+  /**
+   * To facilictate the testing, maintain the order of loaded samples
+   * in the sample list as it is in the list file.
+   */
+  void keep_sample_order(bool same_order = false);
 
   /**
    * Set the filename for your data (images, etc).
@@ -623,7 +630,7 @@ class generic_data_reader {
 
   void instantiate_data_store();
 
-  virtual void preload_data_store(); 
+  virtual void preload_data_store();
 
   void set_gan_labelling(bool has_gan_labelling) {
      m_gan_labelling = has_gan_labelling;
@@ -836,6 +843,10 @@ private:
   /// special handling for 1B jag; each reader
   /// owns a unique subset of the data
   bool m_jag_partitioned;
+
+  /** Whether to keep the order of loaded samples same as it is in the
+   *  file to make testing and validation easier */
+  bool m_keep_sample_order;
 
   /// called by fetch_data a single time if m_jag_partitioned = true;
   /// this sets various member variables (num_iterations, m_reset_mini_batch_index,
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -89,7 +89,7 @@ class generic_data_reader {
     m_world_master_mini_batch_adjustment(0),
     m_num_parallel_readers(0), m_rank_in_model(0),
     m_max_files_to_load(0),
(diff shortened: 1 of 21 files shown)


davidHysom (approved):
This worked for me; see my blog post for today's mtg. Only comment: I think the "index_list:" field in the data reader should be renamed to "sample_list" for clarity.

JaeseungYeom:
> This worked for me; see my blog post for today's mtg. Only comment: I think the "index_list:" field in the data reader should be renamed to "sample_list" for clarity.

I will make the change.

JaeseungYeom:
Unless anyone has any additional comment, I will merge this PR at the end of the day.

davidHysom:
> Unless anyone has any additional comment, I will merge this PR at the end of the day.

My only concern is testing: will it break anything?

bvanessen:
I want to look at this before it merges.  I know that it has been waiting for a long time, but I still want to look through it before merging.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

bvanessen (changes requested):
I want to review this.

JaeseungYeom:
I will be waiting.

JaeseungYeom:
Rebased and compiles. Will test tomorrow.

bvanessen:
This PR is still not ready to merge.  There are no clear sample lists in the imagenet datasets directory.

JaeseungYeom:
> This PR is still not ready to merge. There are no clear sample lists in the imagenet datasets directory.
`/p/lustre2/brainusr/datasets/ILSVRC2012/sample_list`

bvanessen:
There are many permutations of lists there with similar but not clear names.  Many don’t conform to what I thought was the standard format as well.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

JaeseungYeom:
Hopefully, this helps. Otherwise, we can discuss more during the meeting.

`/p/lustre2/brainusr/datasets/ILSVRC2012/Single_train_c0-9_01_filenames.txt` and `/p/lustre2/brainusr/datasets/ILSVRC2012/sample_list/Single_train_c0-9_01_image_list.txt` are same files in the original format except that the former has the full path to the image file (which is outdated containing lscratchf) while the latter has only the last part of the path that actually varies.

The former is one of the original imagenet label files we have been using. For the latter, the common prefix part is specified in the data reader prototext with the parameter `data_filedir`. This is not the directory for the list file itself but the image data files. We still use the original file for labels.

In case of the sample list files, such a path prefix is specified at the third line in the file as usual.
`/p/lustre2/brainusr/datasets/ILSVRC2012/sample_list/Single_train_c0-9_01_sample_list.txt` is the corresponding sample list file. It does not include the label information as we discussed.

```
data_reader {
  reader {
    ...
    data_filedir:      "/p/lustre2/brainusr/datasets/ILSVRC2012/resized_256x256/train/"
    data_filename: "/p/lustre2/brainusr/datasets/ILSVRC2012/sample_list/Single_train_c0-9_01_image_list.txt"
    sample_list:      "/p/lustre2/brainusr/datasets/ILSVRC2012/sample_list/Single_train_c0-9_01_sample_list.txt"
    ...
}
```

PR #1137, opened 2019-07-31
Refactor weights factory

benson31:
Fairly straight-forward refactor of weights initializers and related changes.

I also noticed that `unique_ptr`s were not being consumed correctly. Since `weights::set_initializer()` and `weights::set_optimizer()` are sinks, they should take their parameter by rvalue reference. Since this is weights-related, I threw it in this PR.


benson31:
This is now up to date with the recent changes in #1133.

PR #1138, opened 2019-08-01
Refactor optimizer factory

benson31:
This was the first one that ended up being exactly as straightforward as it should have been.

Depends on #1133 and #1137.


benson31:
This is now up-to-date with the recent changes in #1133 (and by extension #1137).

src/proto/factories/optimizer_factory.cpp
```diff
@@ -26,69 +26,66 @@
...
+                        lbann_comm*>,
+  default_key_error_policy>;
 
-  // AdaGrad
-  if (proto_opt.has_adagrad()) {
-    const auto& params = proto_opt.adagrad();
-    return new adagrad(comm, params.learn_rate(), params.eps());
-  }
+void register_default_builders(factory_type& factory)
+{
```
ndryden: LBANN style is to put the brace on the same line.

src/proto/factories/optimizer_factory.cpp
```diff
@@ -26,69 +26,66 @@
...
-  if (proto_opt.has_rmsprop()) {
-    const auto& params = proto_opt.rmsprop();
-    return new rmsprop(comm,
-                       params.learn_rate(),
-                       params.decay_rate(),
-                       params.eps());
-  }
+// Manage a global factory
+struct factory_manager
+{
```
ndryden: Ditto.

src/proto/factories/optimizer_factory.cpp
```diff
@@ -26,69 +26,66 @@
...
-    return new hypergradient_adam(comm,
-                                  params.init_learning_rate(),
-                                  params.hyper_learning_rate(),
-                                  params.beta1(),
-                                  params.beta2(),
-                                  params.eps());
-  }
+factory_manager factory_mgr_;
+factory_type const& get_optimizer_factory() noexcept
+{
```
ndryden: Ditto.

src/proto/factories/optimizer_factory.cpp
benson31: All fixed.

PR #1139, opened 2019-08-01
change 'int' to 'size_t' to accommodate large sample sizes

davidHysom:
To accomodate samples sizes that break the 32 bit barrier, the code now uses size_t instead of int.
Removed the following from the header file, since they aren't used in the code:
  std::vector<int> m_recv_buffer_sample_sizes;		  
  std::vector<int> m_send_buffer_sample_sizes;

Replaced a call to build_node_for_sending() to compact_node(). compact_node() wasn't being called -- not sure why


timmoon10:
Issue #91 is related.

Issue #1117, opened 2019-07-20
Thread pool is not compatible with OSX


benson31:
Various components, largely related to the contents of `sched.h` in a GNU/Linux world, of the thread pool implementation are not compatible with OSX. They will not compile and there is not a simple fix (i.e., they didn't just rename the header; the symbols just aren't there).

PR #1144, opened 2019-08-03
Fix an issue on OSX in which nonportable pthread features are not available

benson31:
Addresses issue #1117.


bvanessen:
I was able to build and run this with the superbuild method,

PR #1145, opened 2019-08-05
Refactor model factory

benson31:
This just does some memory-management updates. There's not enough here to merit a proper refactory-ing -- there's only 1 concrete model class and I haven't heard about more in the pipeline. Likewise, objective functions are not polymorphic. However, new memory is returned by the `construct_*` functions, so `unique_ptr` is the correct return type for these functions.

Builds off of #1140.

### DIFF
--- a/include/lbann/utils/exception.hpp
+++ b/include/lbann/utils/exception.hpp
@@ -104,8 +104,7 @@ using lbann_exception = exception;
  *  @param[in] args The things to be stringified.
  */
 template <typename... Args>
-std::string build_string(Args&&... args)
-{
+std::string build_string(Args&&... args) {
   std::ostringstream oss;
   int dummy[] = { (oss << args, 0)... };
   (void) dummy; // silence compiler warnings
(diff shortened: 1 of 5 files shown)


include/lbann/utils/exception.hpp
```diff
@@ -46,6 +46,9 @@
     throw lbann::exception(ss_LBANN_ERROR.str());               \
   } while (0)
 
+#define LBANN_ERROR_STR(...)                    \
+  LBANN_ERROR(build_string(__VA_ARGS__))
```
timmoon10: I think we should make this the default behavior for `LBANN_ERROR`.

benson31: I agree, but I didn't want to worry about pushing the change through all of LBANN. Looking at it now, though, I think the result of just making `LBANN_ERROR` have this behavior would just have been a bunch of extra string copies that really don't matter because it's about to throw an exception from which, realistically, LBANN will never recover (though we should move away from that assumption in `liblbann` -- a library's job is just to report errors; it's an application's job to decide to ignore them or attempt recovery).

PR #1146, opened 2019-08-05
Refactor summarizer construction

graham63:
Refactored summarizer to use shared_ptr in the callbacks. Removed m_summarizer from the base class, and created private m_summarizer variables in the callbacks that use it.


include/lbann/callbacks/callback.hpp
```diff
@@ -63,29 +63,19 @@ class callback_base {
...
 
   ///@}
   /** @name Polymorphic copy */
   ///@{
 
   virtual callback_base* copy() const = 0;
 
-  ///@}
-  /** @name Modifiers */
-  ///@{
```
benson31: Keep this comment block or something like it. (Otherwise `setup()` will be grouped in the `Polymorphic copy` group.

src/proto/factories/callback_factory.cpp
```diff
@@ -87,9 +87,21 @@ using factory_type = lbann::generic_factory<
   std::string,
   generate_builder_type<lbann::callback_base,
                         google::protobuf::Message const&,
-                        lbann_summary*>,
+                        std::shared_ptr<lbann_summary> const&>,
   default_key_error_policy>;
 
+namespace
```
benson31: This doesn't need to be nested in yet another anonymous namespace (it's already in one).

include/lbann/callbacks/summary.hpp
```diff
@@ -60,17 +59,23 @@ class summary : public callback_base {
...
+private:
+  /**@brief lbann_summary */
+  std::shared_ptr<lbann_summary> m_summarizer = nullptr;
+
+protected:
   /** Write out histograms from the model's layers. */
   void save_histograms(model *m);
-  /** Interval for doing matrix summarization. */
+
+/** Interval for doing matrix summarization. */
```
benson31: Data member should be `private`.

src/callbacks/summary.cpp
```diff
@@ -39,21 +39,35 @@
...
+  callback_base(batch_interval),
+  m_summarizer(summarizer),
   m_mat_interval(mat_interval) {}
 
-summary::~summary() {
-  delete m_summarizer;
+namespace
+{
+template <typename... Ts>
+std::string BuildErrorMessage(Ts... args)
```
benson31: Once #1145 merges, this should be replaced with that mechanism.

src/callbacks/summary.cpp
```diff
@@ -39,21 +39,35 @@
...
+{
+  std::ostringstream oss;
+  int dummy[] = { (oss << args, 0)... };
+  (void) dummy;
+  LBANN_ERROR(oss.str());
```
benson31: This should `return oss.str();` But kind of a moot point, see previous comment.

src/proto/factories/callback_factory.cpp
```diff
@@ -87,9 +87,18 @@ using factory_type = lbann::generic_factory<
   std::string,
   generate_builder_type<lbann::callback_base,
                         google::protobuf::Message const&,
-                        lbann_summary*>,
+                        std::shared_ptr<lbann_summary> const&>,
   default_key_error_policy>;
 
+template <typename... Ts>
+std::string BuildErrorMessage(Ts... args)
```
benson31: See other comments on `BuildErrorMessage`.

PR #1147, opened 2019-08-06
Add Sequential Catch Tests

forsyth2:
Add Sequential Catch Tests


timmoon10:
Bamboo reports Catch tests (I believe since PR https://github.com/LLNL/lbann/pull/1270), so I think it's safe to close this.

PR #1148, opened 2019-08-06
Add GRU to list of python modules

samadejacobs:


python/lbann/modules.py
```diff
@@ -361,5 +361,148 @@ def forward(self, x, prev_state):
...
+                            name=name + '_fc2_slice',
+                            data_layout=self.data_layout)
+        Whr_prev = lbann.Identity(fc2_slice, name=name + '_Wrh',
+                           data_layout=self.data_layout)
+        Whz_prev = lbann.Identity(fc2_slice, name=name + '_Wzh',
+                           data_layout=self.data_layout)
+        Whn_prev = lbann.Identity(fc2_slice, name=name + '_Wnh',
+                           data_layout=self.data_layout)
+        
+        rt = lbann.Sigmoid(lbann.Add([Wir_x,Whr_prev]), name=name + '_reset_gate',
```
timmoon10: We need to specify the data layout for the add layer, and make similar changes to all the layers below. It would also be nice to make sure they are all named.

python/lbann/modules.py
```diff
@@ -361,5 +361,148 @@ def forward(self, x, prev_state):
...
+    def __init__(self, size, bias = True,
+                 weights=[], name=None, data_layout='data_parallel'):
+        """Initialize GRU cell.
+
+        Args:
+            size (int): Size of output tensor.
+            bias (bool): Whether to apply biases after linearity.
+            weights (`Weights` or iterator of `Weights`): Weights in
+                fully-connected layer. There are at most four - two
+                matrices ((3*size) x (input_size+size) dimensions) each and two
```
timmoon10: The dimensions are `(3*size) x input_size` and `(3*size) x size`.

PR #1149, opened 2019-08-06
use realpaths instead of abspaths in setup.py

benson31:
There was an issue with a user building in a directory that had a symlink in it. The structure was roughly:
```
source: /a/b/lbann/python
build: /a/b/symlink/superbuild_dir/lbann/build
```
The relative path from build to source is computed as `../../../../lbann/python`. Of course, this doesn't work. Thus, we need to use the relative path from `realpath(build)` to `realpath(source)`.


benson31:
They're probably safer. One issue here is that CMake uses absolute paths for everything unless you specifically fetch the real path. We could have solved it at the CMake level, too, but this puts it closer to where the relative paths get used.

PR #1150, opened 2019-08-07
Semi Useful lenet2.py added

anmolpau:
@gunney1 @benson31 @timmoon10 @ndryden


model_zoo/vision/lenet2.py
```diff
@@ -0,0 +1,131 @@
...
+                    layers=lbann.traverse_layer_graph(input),
+                    objective_function=loss,
+                    metrics=[lbann.Metric(acc, name='accuracy', unit='%')],
+                    callbacks=[lbann.CallbackPrint(), lbann.CallbackTimer()])
+
+# Setup optimizer
+opt = lbann.SGD(learn_rate=0.01, momentum=0.9)
+
+# Setup DataReader
+message = lbann.lbann_pb2.DataReader()
```
benson31: This won't work in the latest `develop`. The `DataReader` message is now in `reader.proto`/`reader_pb2.py`, which is not formally imported into `lbann`.

timmoon10:
This script is good, especially if you want to modify it for other experiments, but I don't think the model zoo is the right place for it. The LeNet script should be designed as a pedagogical tool and the most straightforward approach for data readers is "copy a prototext file." We should follow up when we sort out the data reader API.

PR #1153, opened 2019-08-07
bug fixes for data_store_conduit for the mode: --preload_data_store. …

davidHysom:
…There were two problems.

  1) the entire data set was being loaded, even if the data reader specified that only an absolute number, or precentage should be used
  2) the validation set's schema was corrupted (see: data_store_conduit::copy_members); I don't know why, as this used to work; changes to the conduit repo, perhaps?

To get data store preloading to work I had to refactor data_store::select_subset_of_data(). Formerly this method performed two functions:

  1) resizing the shuffled_indices list
  2) selecting a portion of the list for the validation set.

To fix preloading (first problem) we need to perform (1), then call preload for
the data store, then call (2). Hence, select_subset_of_data() was refactored int
o two methods: resize_shuffled_indices() and select_subset_of_data()

This refactor required touching every data reader: calls to select_subset_of_data() needed to be replaced by two calls:
  resize_shuffled_indices()
  select_subset_of_data()

### DIFF
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -550,9 +550,17 @@ class generic_data_reader {
   }
 
   /**
-   * Select the appropriate subset of data based on settings.
+   * Optionally resizes the shuffled indices based on the data reader
+   * prototext settings: absolute_sample_count, percent_of_data_to_use.
+   * (dah - this was formerly part of select_subset_of_data)
    */
-  virtual void select_subset_of_data();
+  void resize_shuffled_indices();
+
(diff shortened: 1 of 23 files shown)


naoyam:
Though I was requested to review, I'm not familiar with the design of the data reader/store classes, so I can't say whether the PR should be merged or not. I just wonder it would be good if changes in the data store could be contained in its class only without requiring changes all data readers.

davidHysom:
OK.
The changes to the data readers is minor: the call to select_subset_of_data
is preceded by a call to resize_shuffled_indices.

PR #1155, opened 2019-08-08
Refactoring LBANN_ERROR to make it more flexible.

benson31:
It's not a particular pretty anti-pattern to build up some string stream in the middle of the code. This hides that detail by just making LBANN_ERROR (and LBANN_WARNING) accept variadic arguments.

This doesn't clean up any code in the guts of LBANN except what needed to be cleaned up to make this work. It would be a massive PR to do that all now. Rather, IMO, it should be done piecemeal as others are touching other parts of the code.

### DIFF
--- a/include/lbann/utils/cublas.hpp
+++ b/include/lbann/utils/cublas.hpp
@@ -29,6 +29,7 @@
 
 #include "lbann/base.hpp"
 #include "lbann/utils/cuda.hpp"
+#include "lbann/utils/exception.hpp"
 
 #ifdef LBANN_HAS_CUDA
 #include <cuda_runtime.h>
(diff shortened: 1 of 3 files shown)


benson31:
Strictly speaking, the changes in `cublas.hpp` weren't necessary -- I just misunderstood the compiler error. But by the time I realized what it was actually telling me, I had already made all the changes...

include/lbann/utils/cublas.hpp
```diff
@@ -55,8 +57,8 @@
         status_FORCE_CHECK_CUBLAS = cudaGetLastError();                 \
       if (status_FORCE_CHECK_CUBLAS != cudaSuccess) {                   \
         cudaDeviceReset();                                              \
-        LBANN_ERROR(std::string("CUDA error: ")                         \
-                    + cudaGetErrorString(status_FORCE_CHECK_CUBLAS));   \
+        LBANN_ERROR("CUDA error: ",                                     \
+                    cudaGetErrorString(status_FORCE_CHECK_CUBLAS));     \
```
ndryden: Spaces?

benson31: The column location of the `\` didn't change; these are the extra spaces required to keep the alignment.

PR #1156, opened 2019-08-08
Fix the tests prototexts

benson31:
Missed the `model_zoo/tests/` files in #1133.


benson31:
I checked `bamboo/` via `grep`, but no manual inspection beyond that. Will do.

PR #1158, opened 2019-08-09
Fix unit test error related to empty optimizer messages

benson31:
Fixes an issue in which an empty `optimizer` message is both valid and different from not having such a message at all.


benson31:
After the python front-end fix (https://github.com/LLNL/lbann/commit/e9bc5119c7ff10fa2f9316207634d3172a2bff53) I can confirm this has not broken the `model_zoo/vision` models and that `unit_layer_tesselate` test passes.

PR #1161, opened 2019-08-12
Protobuf message for no optimizer

timmoon10:
When constructing models, we sometimes need to specify weights without any optimizers. We currently do this with empty `Optimizer` messages (we have different behavior if the `Weights`'s `optimizer` field is not set at all, in which case we construct the model's default optimizer). This is a bit of a hack and doesn't work with the recent optimizer factory refactor (see PRs #1138 and #1158). To get around this, we can now specify no optimizer with the `NoOptimizer` message or the corresponding class in the Python frontend.

### DIFF
--- a/src/proto/factories/optimizer_factory.cpp
+++ b/src/proto/factories/optimizer_factory.cpp
@@ -46,7 +46,7 @@ namespace {
 std::unique_ptr<optimizer>
 build_no_optimizer_from_pbuf(
   google::protobuf::Message const& msg, lbann_comm* comm) {
-  return std::unique_ptr<optimizer>();
+  return nullptr;
 }
 
 using factory_type = lbann::generic_factory<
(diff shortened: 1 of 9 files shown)


src/proto/factories/optimizer_factory.cpp
```diff
@@ -43,6 +43,12 @@ namespace lbann {
 namespace proto {
 namespace {
 
+std::unique_ptr<optimizer>
+build_no_optimizer_from_pbuf(
+  google::protobuf::Message const& msg, lbann_comm* comm) {
+  return std::unique_ptr<optimizer>();
```
benson31: This can just return `nullptr;` All that extra typing... what would @ndryden think?? 😛

ndryden: My poor wrists... are you trying to disable me? 😛

PR #1162, opened 2019-08-13
Check gradient callback can be applied at any execution mode

timmoon10:
Previously, gradient checking was only performed at the end of the test phase. This is inconvenient for verifying batchnorm, which has different behavior during training and inference. This PR adds functionality so that the user can choose when gradient checking happens: after each training epoch or after each validation or training set phase.

### DIFF
--- a/include/lbann/callbacks/check_gradients.hpp
+++ b/include/lbann/callbacks/check_gradients.hpp
@@ -36,9 +36,9 @@ namespace callback {
 
 /** @brief Gradient checking callback.
  *
- *  Gradient checking is performed at the end of the test phase. Using
- *  a fourth-order finite difference scheme, a numerical partial
- *  derivative is computed for every weight parameter. If the
+ *  Gradient checking is performed at the end of each execution mode
+ *  phase. Using a fourth-order finite difference scheme, a numerical
+ *  partial derivative is computed for every weight parameter. If the
  *  numerical derivative differs signifcantly from the analytical
  *  derivative computed during backprop, the gradient check has
  *  failed.
@@ -47,6 +47,7 @@ class check_gradients : public callback_base {
 public:
 
   /**
+   *  @param modes              Execution modes with gradient checks.
    *  @param step_size          Step size for numerical
    *                            differentiation (with a step size of
    *                            zero, the step size is estimated to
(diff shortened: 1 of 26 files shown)


src/callbacks/check_gradients.cpp
```diff
@@ -64,40 +65,44 @@ DataType compute_objective_function(model& m) {
...
+  const auto& layers = m.get_layers();
+
+  // Return immediately if execution mode is invalid
+  if (!m_modes.empty() && m_modes.count(mode) == 0) { return; }
 
   // Reset statistics and gradients
-  m->get_objective_function()->reset_statistics(mode);
-  for (auto&& met : m->get_metrics()) {
+  m.get_objective_function()->reset_statistics(mode);
+  for (auto&& met : m.get_metrics()) {
```
ndryden: Not really a comment on this PR, but why do we need to reset _metrics_ here?

include/lbann/callbacks/check_gradients.hpp
```diff
@@ -56,24 +57,32 @@ class check_gradients : public callback_base {
    *  @param error_on_failure   Whether to throw an exception for
    *                            large gradient errors.
    */
-  check_gradients(DataType step_size = DataType(0),
-                                 bool verbose = false,
-                                 bool error_on_failure = false);
+  check_gradients(std::set<execution_mode> modes = {},
```
ndryden: The documentation should say what the default (empty set) does.

src/callbacks/check_gradients.cpp
```diff
@@ -64,40 +65,44 @@ DataType compute_objective_function(model& m) {
...
   // Get objects from model
-  lbann_comm *comm = m->get_comm();
-  auto mode = m->get_execution_mode();
-  const auto& layers = m->get_layers();
+  auto& comm = *m.get_comm();
+  const auto mode = m.get_execution_mode();
+  const auto& layers = m.get_layers();
+
+  // Return immediately if execution mode is invalid
+  if (!m_modes.empty() && m_modes.count(mode) == 0) { return; }
```
ndryden: "Invalid" is a bad word choice. This returns if the user specified particular execution modes _and_ the current mode was not specified.

src/callbacks/check_gradients.cpp
timmoon10: We previously did gradient checking at the beginning of the test phase, which caused problems with metric checking since repeated forwardpropping messes up the metric statistics. I tried fixing this by reseting the metrics, but I couldn't figure out a good way to reset the data reader.

In any case, we should expect the metric statistics to be trashed after gradient checking, so I don't think reseting them does any harm.

include/lbann/callbacks/check_gradients.hpp
timmoon10: Done.

src/callbacks/check_gradients.cpp
timmoon10: Reworded.

PR #1164, opened 2019-08-13
Allow no initializer specified in weights object

benson31:
This patches an issue where a weights object might not have an initializer specified. I've also added a message type to express this if folks feel like being a bit more explicit.

### DIFF
--- a/src/proto/weights.proto
+++ b/src/proto/weights.proto
@@ -38,7 +38,6 @@ message Weights {
 
 message Initializer {
   oneof initializer_type {
-    LayerDefault layer_default = 1;
     ConstantInitializer constant_initializer = 20;
     ValueInitializer value_initializer = 21;
     UniformInitializer uniform_initializer = 22;
@@ -52,8 +51,6 @@ message Initializer {
   }
 
   // Weight initializers
-  message LayerDefault {}
-    
   message ConstantInitializer {
     double value = 1;
   }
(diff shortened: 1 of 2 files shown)


src/proto/weights.proto
```diff
@@ -51,6 +52,8 @@ message Initializer {
   }
 
   // Weight initializers
+  message LayerDefault {}
```
timmoon10: If no initializer is set, we don't use a "layer default". The weights object might not be associated with any layer, or with multiple layers. The actual default is zero initialization.

I can buy the argument that users should specify every `oneof` field within a message, but I don't think it's the right approach for users to specify every field that contains a `oneof` field. For instance:
```
weights {
  name: "my_weights"
}
```
should instantiate weights with zero initialization and the model's default optimizer (in the future it might be better to instantiate no optimizer).

benson31: Got it. Will remove this. Since the default already has a name, no sense naming it twice.

ndryden: I thought that we used a default, e.g. for convolution, a He initializer. Or is that only if you don't specify the weights at all?

timmoon10: If a layer gets to create a weights, it can choose a suitable initializer. If a weights is created in the void, then we can't make any assumptions and we should use the most inoffensive initialization scheme.

benson31: `LayerDefault` has been removed. Thanks for the clarifications.

PR #1165, opened 2019-08-13
pin conduit version to v0.4.0

benson31:


davidHysom:
Yay; my build now works!

PR #1167, opened 2019-08-13
Entry-wise batchnorm

timmoon10:
Our current batchnorm does not work well with large fully-connected networks (like the last few layers in AlexNet) since it only supports data-parallel data. Supporting data in (`MC`, `MR`) format is not feasible. This PR implements an entry-wise batchnorm layer, which is equivalent to our current usage with flattened data, but can also support the model-parallel case. Gradient checking passes on for both data-parallel and model-parallel cases on both CPU and GPU.

Note that we currently only support global statistics. I don't think it would be hard to support local statistics, but it might not be advisable since statistics counts are already so small. Defining reasonable "statistics groups" would also be trickier for (`MC`, `MR`) data: row communicators would probably consist of non-adjacent processes, so I don't expect much latency benefits from reducing over a subset of the row communicator.

### DIFF
--- a/include/lbann/layers/regularizers/entrywise_batch_normalization.hpp
+++ b/include/lbann/layers/regularizers/entrywise_batch_normalization.hpp
@@ -1,0 +1,210 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+// may not use this file except in compliance with the License.  You may
+// obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the license.
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef LBANN_LAYERS_REGULARIZERS_ENTRYWISE_BATCH_NORMALIZATION_HPP_INCLUDED
+#define LBANN_LAYERS_REGULARIZERS_ENTRYWISE_BATCH_NORMALIZATION_HPP_INCLUDED
+
+#include "lbann/layers/layer.hpp"
+#include "lbann/models/model.hpp"
+#include "lbann/utils/memory.hpp"
+
+namespace lbann {
+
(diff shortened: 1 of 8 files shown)


include/lbann/layers/regularizers/entrywise_batch_normalization.hpp
```diff
@@ -0,0 +1,210 @@
...
+#ifndef LBANN_LAYERS_REGULARIZERS_ENTRYWISE_BATCH_NORMALIZATION_HPP_INCLUDED
+#define LBANN_LAYERS_REGULARIZERS_ENTRYWISE_BATCH_NORMALIZATION_HPP_INCLUDED
+
+#include "lbann/layers/layer.hpp"
+#include "lbann/models/model.hpp"
+#include "lbann/utils/memory.hpp"
+
+namespace lbann {
+
+template <data_layout Layout, El::Device Device>
```
ndryden: A touch of documentation would be good. 😉

src/layers/regularizers/entrywise_batch_normalization.cpp
```diff
@@ -0,0 +1,423 @@
...
+  /// @todo Arbitrary group sizes
+  comm.allreduce(batch_statistics,
+                 batch_statistics.RedundantComm(),
+                 El::mpi::SUM);
+  const size_t statistics_count = input.Width();
+
+  // Compute mini-batch statistics from sums
+  if (statistics_count <= 1) {
+    // local_mean already has correct values
+    El::Fill(local_batch_var, DataType{1});
```
ndryden: This case occurs when the mini-batch size being aggregated over is 1 or less, right? We may want to throw an error instead, since the variance isn't well-defined then.

src/layers/regularizers/entrywise_batch_normalization.cpp
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+}
+
+/** @brief Backprop for inference.
+ *
+ *  Assumes forward prop uses running statistics. In other words,
+ *  statistics are independent of input.
+ */
+void bp_inference_impl(DataType epsilon,
```
ndryden: "Backprop for inference" is an odd thing to say. I think this is just computing dL/dx? We should say something along those lines.

src/proto/factories/layer_factory.cpp
```diff
@@ -571,6 +572,11 @@ std::unique_ptr<Layer> construct_layer(
       return lbann::make_unique<selu_dropout<Layout, Device>>(comm, keep_prob);
     }
   }
+  if (proto_layer.has_entrywise_batch_normalization()) {
+    const auto& params = proto_layer.entrywise_batch_normalization();
+    /// @todo Support GPU implementation
```
ndryden: Doesn't this support GPUs already?

include/lbann/layers/regularizers/entrywise_batch_normalization.hpp
timmoon10: It's self-documenting, ain't it? :p

src/layers/regularizers/entrywise_batch_normalization.cpp
timmoon10: I think that's too harsh, since the user will typically have little control on the size of the last mini-batch.

src/layers/regularizers/entrywise_batch_normalization.cpp
timmoon10: I'd like to make a distinction between the implementation for training and for inference. Perhaps more documentation is in order.

src/proto/factories/layer_factory.cpp
timmoon10: Oops, forgot to delete that.

PR #1168, opened 2019-08-14
Clean up the layer instantiation

benson31:
Layers are repeatedly instantiated (this is ultimately fine -- the linker figures it out) but it wastes a lot of time. The aim of this PR is to ensure that each layer class is only instantiated once, thereby improving compile times. The ultimate target, of course, being `layer_factory.cpp`.


benson31:
I've gone through all the layers and added ETI. This is done manually for each "independent" layer and added to the macros for the "automatic" layers (unary, binary, and friends).

A few other changes, that were necessary to get this to compile cleanly or correctly:
1. `static_asserts` on template parameters were removed from function scope and replaced at class scope. This is where they belong since they assert class validity, not constructor validity (no, they aren't the same).
2. Some assignment operators didn't return `*this` (or anything else). They do now. (The fact that this warning never popped up before means that these functions were never instantiated.)

I have the Bamboo CI running on this now; I'll re-comment when it's done. In the meantime, I have this building with and without CUDA, and all the Catch2 tests pass.

benson31:
I see nothing failing in Bamboo on this branch that isn't failing on `develop`.

benson31:
Also, this shaves about 45 seconds off the build on pascal -- somewhere around 25% of the build time of `develop`. Not quite as much as I was hoping, but 🤷‍♂ what're ya gonna do...

src/layers/basic_template.cpp
```diff
@@ -0,0 +1,31 @@
+////////////////////////////////////////////////////////////////////////////////
```
timmoon10: Do we need this file?

include/lbann/layers/io/input/input_layer.hpp
```diff
@@ -99,6 +100,19 @@ template<>
 inline void input_layer<partitioned_io_buffer, data_layout::DATA_PARALLEL, El::Device::GPU>::validate_data_layout() {}
 #endif // LBANN_HAS_GPU
 
-}
+#ifndef LBANN_INPUT_LAYER_INSTANTIATE
+extern template class input_layer<
+  partitioned_io_buffer, data_layout::DATA_PARALLEL, El::Device::CPU>;
```
timmoon10: I believe that the only valid input layer is data-parallel on CPU, at least currently.

include/lbann/layers/activations/activations.hpp
```diff
@@ -31,30 +31,52 @@
...
 // Convenience macro to define an entry-wise unary layer class
 #define DEFINE_ENTRYWISE_UNARY_LAYER(layer_name, layer_string)          \
   struct layer_name##_name_struct {                                     \
     inline operator std::string() { return layer_string; }              \
   };                                                                    \
   template <data_layout Layout, El::Device Device>                      \
   using layer_name                                                      \
-  = entrywise_unary_layer<Layout, Device, layer_name##_name_struct>;
+  = entrywise_unary_layer<Layout, Device, layer_name##_name_struct>;    \
+  UNARY_ETI_DECL_MACRO(layer_name)
```
timmoon10: Micro-quibble: When `LBANN_ACTIVATIONS_LAYER_INSTANTIATE` is active, then `UNARY_ETI_DECL_MACRO_DEV` is empty and you get two consecutive semicolons. This could cause overly pedantic compilers to throw up a warning.

include/lbann/utils/eti_macros.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef LBANN_UTILS_ETI_MACROS_HPP_INCLUDED
```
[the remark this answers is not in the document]
benson31: Removed.

src/layers/basic_template.cpp
benson31: Removed.

include/lbann/layers/activations/activations.hpp
benson31: This should only happen if `-Wpedantic` or the equivalent is on. If it becomes an issue, I'll fix it.

include/lbann/layers/io/input/input_layer.hpp
benson31: It is true that this is the only valid input layer, but this is, unfortunately, a dynamic check rather than a static one. These symbols are required for successful linkage.

PR #1169, opened 2019-08-14
Data store matrix

davidHysom:
Modifications to data_store_conduit and data_reader_jag_conduit to (hopefully) improve efficiency/decrease run time. Currently we store the "raw" JAG samples in the data store. This means that, in every call to fetch_datum there are separate calls to extract each of the scalars (five) and inputs (a dozen or so) and the images (three, if I recall correctly). 

In the new scenario we store a vector (or ptr) to the values in the CPUMat as it exists just prior to the return from fetch_datum(). 

This PR is only applicable to the data_reader_jag_conduit. It is experimental in the sense that I don't really know if this will improve efficiency -- but have a hunch that it might.

In this PR the existing usage of data_store_conduit is retained; a cmd line parameter, "data_store_matrix" triggers the new behavior.


davidHysom:
There are too many conflicts with the current head of develop, so I'm closing this and will start over.

PR #1175, opened 2019-08-16
Build sample lists optimization

davidHysom:
Building sample lists using the python wrapper (which calls select_samples.cpp for each list) has become too slow for 640 lists an 100M JAG samples. This PR optimizes the code to make it run much faster.


samadejacobs:
I am not a reviewer but since I need this I decided to test it. The code fails to build here are few snippet of error (there are quite a lot of them):
select_samples.cpp:64:82:
error: binding reference of type 'std::string& {aka std::basic_string<char>&}' to 'const string {aka const std::basic_string<char>}' discards qualifiers
   read_mapping_file(mapping_fn, sample_mapping, sample_mapping_v, string_to_index);
class lbann::options' has no member named 'has_bool'; did you mean 'get_bool'?
   if (opts->has_bool("h") || opts->has_bool("help") || argc == 1) {
             ^~~~~~~~
select_samples.cpp:160:31: error: 'output_fn' was not declared in this scope
     const std::string name1 = output_fn + "_bar";
                               ^~~~~~~~~

davidHysom:
The code is under development, and not ready for review and/or testing. Hopefully it will be working by Monday.

model_zoo/jag_utils/select_samples.cpp
```diff
@@ -81,209 +221,370 @@ int main(int argc, char **argv) {
...
+    vector<unordered_map<string, unordered_set<int>>> &sets) {
+  size_t samples_per_list = options::get()->get_int("num_samples_per_list");
+  size_t which = 0;
+  size_t count = 0;
+  size_t total = 0;
+  for (auto &it : index_map_keep) {
+    const string &filename = it.first;
+    const unordered_set<int> &sample_ids = it.second;
+    for (auto &it2 : sample_ids) {
+      sets[which][filename].insert(it2);
```
bvanessen: Looking at this, I believe that it will interleave all of the samples across each training list being generated.  Is that correct?  If so, this is a step in the right direction, but we really should randomly pull a sample and then randomly assign it to the list.

bvanessen: Although, given that it is pulling from a unordered set, maybe it is okay to interleave the assignment.

davidHysom: Hm, I'll have to think on this (possibly very early tomorrow). 
"randomly pull a sample" - I believe the code does this
"and then randomly assign it to the list" - but I think the ++which effectively does this. OK, it's not random, but I think (??) should have the same effect.

I'll work on making this more random.

PR #1179, opened 2019-08-19
silence gcc warning by initializing a variable

benson31:


src/data_store/data_store_conduit.cpp
```diff
@@ -1172,7 +1172,7 @@ void data_store_conduit::allocate_shared_segment(std::unordered_map<int,size_t>
   }
   m_comm->trainer_barrier();
 
-  int shm_fd;
+  int shm_fd = -1;
```
benson31: This is the important change. The rest are just whitespace changes because I have emacs configured to delete trailing whitespace before saving.

benson31:
Ugh I hate the "cuddled else", too. New control block, new line, if you ask me!

PR #1181, opened 2019-08-21
Callback to print model descriptions

timmoon10:
Whenever we construct a model, we print out a human-readable message that describes the layers, weights, callbacks, etc. This is useful for understanding the model and debugging, but it gets a bit overwhelming when I run VRAM since it has >1000 layers.

This PR makes printing the model description optional. If the user wants all of this information, they can add the "print model description" callback. In the process of implementing this, I've added new callback hooks at the beginning and end of model setup.

### DIFF
--- a/include/lbann/callbacks/callback.hpp
+++ b/include/lbann/callbacks/callback.hpp
@@ -31,6 +31,7 @@
 
 #include "lbann/layers/layer.hpp"
 #include "lbann/models/model.hpp"
+#include "lbann/utils/description.hpp"
 #include "lbann/utils/memory.hpp"
 #include "lbann/utils/summary.hpp"
 
@@ -89,6 +90,10 @@ public:
   /** @name Callback hooks */
   ///@{
 
+  /** @brief Called at the beginning of setup. */
+  virtual void on_setup_begin(model *m) {}
+  /** @brief Called at the end of setup. */
+  virtual void on_setup_end(model *m) {}
   /** @brief Called at the beginning of training. */
   virtual void on_train_begin(model *m) {}
   /** @brief Called at the end of training. */
@@ -172,6 +177,9 @@ public:
 
   /** @brief Return this callback's name. */
   virtual std::string name() const = 0;
+
+  /** @brief Human-readable description. */
+  virtual description get_description() const;
 
   ///@}
 
--- a/include/lbann/callbacks/print_model_description.hpp
+++ b/include/lbann/callbacks/print_model_description.hpp
@@ -32,6 +32,12 @@
 namespace lbann {
 namespace callback {
 
+/** @brief Print human-readable description of model to standard input.
+ *
+ *  Message is printed when the model has finished setup. The
+ *  description includes information on the model's layers, weights,
+ *  and callbacks.
+ */
 class print_model_description : public callback_base {
 public:
   print_model_description() : callback_base() {}
--- a/include/lbann/callbacks/print_model_description.hpp
+++ b/include/lbann/callbacks/print_model_description.hpp
@@ -1,0 +1,54 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+// may not use this file except in compliance with the License.  You may
+// obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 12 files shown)


include/lbann/callbacks/callback.hpp
```diff
@@ -89,6 +90,10 @@ class callback_base {
   /** @name Callback hooks */
   ///@{
 
+  /** @brief Called at the beginning of setup. */
+  virtual void on_setup_begin(model *m) {}
```
benson31: This PR is not the right place to make this change, but these should all take `model&`. And they should all be `const`.

include/lbann/callbacks/print_model_description.hpp
```diff
@@ -0,0 +1,60 @@
+////////////////////////////////////////////////////////////////////////////////
```
benson31: This file should be added to the `include/lbann/callbacks/CMakeLists.txt` file.

src/models/model.cpp
```diff
@@ -268,7 +268,19 @@ description model::get_description() const {
   desc.add(std::string{});
   desc.add(weights_desc);
 
-  /// @todo Descriptions for objective function, metrics, callbacks
+  // Callbacks
+  description callback_desc("Callbacks:");
+  for (const auto& cb : m_callbacks) {
+    if (cb == nullptr) {
+      callback_desc.add("unknown callback");
```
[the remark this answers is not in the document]
timmoon10: That's a good point. I'll just dereference it and it'll be someone else's responsibility to check for nullity.

src/models/model.cpp
```diff
@@ -1212,6 +1228,18 @@ void model::reconcile_weight_values() {
 // Callbacks
 // =============================================
 
+void model::do_setup_begin_cbs() {
+  for (const auto& cb : m_callbacks) {
+    cb->on_setup_begin(this);
```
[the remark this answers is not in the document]
timmoon10: `model::add_callback(callback_base*)` has a check for null pointers, so we should be safe.

include/lbann/callbacks/callback.hpp
timmoon10: It would be nicer if the callbacks worked on references. They shouldn't be `const` though, since the callback might manipulate the model or its members.

benson31: The callback functions should be `const`:
```cpp
virtual void on_setup_begin(model&) const;
```

timmoon10: I'm not sure about that either, since the callback might maintain some state or preallocate buffers.

benson31: Invoking a callback should not affect it's *logical* state. If the bitwise state needs to change, the affected parameters should be marked `mutable` (and suitably protected with mutexes or atomics, etc, if multithreaded access is a concern).

timmoon10: It looks like this is a question about whether callbacks should be considered pure functions or not. Making them pure functions will make them easier to reason about, but it also excludes some useful functionality. For example, a callback for learning rate schedules needs to store the initial learning rate (logically at `on_setup_end`). A callback that checkpoints the model at regular time intervals needs to keep track of the last checkpoint time.

benson31: I have no objection to them having state -- they probably should. Memoization and/or other caching does not affect the fact that a callback is logically `const` whilst it's being applied -- nothing that defines the external character of the callback is changing (that is, if a change of state did occur, it would not be externally visible). For the learning rate schedule example, the initial learning rate is merely memoized. Neither the schedule or the mechanism for applying that schedule are affected by actually applying it. For the checkpoint interval, its progress is, again, internal cached state -- nothing externally visible occurs when the interval tracker is updated.

I vote we move this to another forum, perhaps an issue discussing the design of callbacks. As we've discussed previously, I think we need to rethink the interface (or eliminate it in favor of a generalized functor) and allow a wider variety of functors to be usable as callbacks. A few notable issues:
1. `const`-correctness is lacking (i.e., this discussion)
2. It's difficult to register new hooks and requires changing individual callbacks to allow them to work with new hooks.
3. The hooks that are there are clunky because they incur an overhead with `n` virtual function dispatches every time, even though many of those dispatches could be empty functions.
4. The interface belies the semantics of callbacks. That is, each hook takes (at least) a model as a parameter, implying that the callback should be applicable to whatever model gets passed into it. For example:
```cpp
model a, b;
callback::whatever cb;
cb->on_epoch_end(a);
cb->on_epoch_end(b);
```
There's no reason this code won't execute -- none of the callbacks explicitly register the model with which they were setup. There are no comments to suggest one should not execute this code. There's also no guarantee that any invariants of `callback::whatever` will be left intact after this code. For the examples of the callbacks that cache information about model state, for example, that information is probably invalid if a different model gets passed into the same callback object. Of course, this doesn't happen in the blessed executables of today because callbacks are only executed inside the model. However, neither are they implementation details of the model nor is their interface restricted away from users, opening them up to potential abuse of this kind as others build custom training algorithms or the like with the library components of LBANN.

benson31: Also, I think `on_setup_begin(model *m)` is not a valid hook -- the callback may not have been setup yet since callback setup is done as part of model setup... While many callbacks don't have a nonempty setup function, this is a semantic issue.

timmoon10: Yea, I was thinking about whether this was a good idea or not. I can imagine use-cases (printing information about the LBANN build or just printing the fact that setup has begun), but perhaps it's just too dangerous.

benson31: "Dangerous" might be overselling, but certainly its potential for semantic nonsense is quite high. Also, things like the build information are not characteristic of a single model but are truly global. There should be another, independent mechanism for reporting this information. The `print_model_description` callback makes semantic sense because callbacks are model-specific and it just prints the specific model's description.

PR #1188, opened 2019-08-26
Custom batch scripts in Python frontend

timmoon10:
This provides an API to create custom batch scripts. For instance, suppose we want to cache data and perform a scaling study:

```python
script = lbann.contrib.lc.launcher.make_batch_script(nodes=32)

script.add_command('echo Caching data in local file system...')
script.add_parallel_command('cp /pfs/mydata.npz /temp/mydata.npz', procs_per_node=1)

script.add_command('echo Scaling study...')
for nodes in (1, 2, 4, 8, 16, 32):
  script.add_parallel_command('lbann --prototext=/path/to/experiment.prototext', nodes=nodes)

script.submit()
```

This PR implements several batch script managers (`SlurmBatchScript` and `LSFBatchScript`) and refactors `lbann.run` to use them internally. This is considered expert functionality and it is still preferable to use `lbann.run` or `lbann.contrib.lc.launcher.run`.

I've tested on Pascal and Lassen.


python/lbann/launcher/batch_script.py
```diff
@@ -0,0 +1,168 @@
...
+        self.script_file = script_file
+        if not self.script_file:
+            self.script_file = os.path.join(self.work_dir, 'batch.sh')
+        self.script_file = os.path.realpath(self.script_file)
+        self.out_log_file = os.path.join(self.work_dir, 'out.log')
+        self.err_log_file = os.path.join(self.work_dir, 'err.log')
+
+        # Shebang line
+        if interpreter:
+            self.header.append('#!{}'.format(interpreter))
```
ndryden: It might be slightly more clean to use `add_header_line`.

python/lbann/launcher/batch_script.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def add_body_line(self, line):
+        """Add line to script body.
+
+        The body should contain the script's executable commands.
+
+        """
+        self.body.append(line)
+
+    def add_command(self, command):
```
ndryden: If we don't intend to do anything else special here, just provide a more semantically correct name, `add_command = add_body_line` might be better, rather than just having it call the other function.

python/lbann/launcher/batch_script.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        raise NotImplementedError(
+            'classes that inherit from `BatchScript` should implement '
+            '`add_parallel_command` to use a specific job scheduler'
+        )
+
+    def write(self):
+        """Write script to file.
+
+        Overwrites script file if it already exists. The working
```
ndryden: It might make things slightly more flexible / avoid unfortunate failures if overwriting was optional (maybe default to not?) and controlled via a kwarg.

python/lbann/launcher/slurm.py
```diff
@@ -1,136 +1,150 @@
...
-            raise ValueError('Slurm reservations not supported')
-        f.write('#SBATCH --workdir={}\n'.format(experiment_dir))
-        f.write('#SBATCH --output={}\n'.format(out_file))
-        f.write('#SBATCH --error={}\n'.format(err_file))
-        if time_limit >= 0:
+            self.add_header_line('#SBATCH --job-name={}'.format(job_name))
+        self.add_header_line('#SBATCH --nodes={}'.format(nodes))
+        if time_limit is not None:
+            time_limit = max(time_limit, 0)
             seconds = int((time_limit % 1) * 60)
```
ndryden: We specify in the documentation that `time_limit` is an `int`, so this is always `0`.

ndryden: Also, since we document `time_limit` as being in minutes, we shouldn't bother to compute seconds.

python/lbann/launcher/batch_script.py
timmoon10: Done.

python/lbann/launcher/batch_script.py
timmoon10: Done.

python/lbann/launcher/batch_script.py
timmoon10: Done.

python/lbann/launcher/slurm.py
timmoon10: I think those are valid points, but it doesn't really hurt us to be a bit more robust than our documentation promises. I'm also endlessly paranoid about the user passing in weird arguments.

PR #1197, opened 2019-09-04
Data store ssd

davidHysom:
This PR adds functionality for the data_store to offload its data to a local disk, and to reload from local disk. This functionality is useful in two scenarios: 1) when there is insufficient memory for the data store to retain all data in memory; 2) for check-pointing and restart


davidHysom:
Too many conflicts ... starting over.

PR #1231, opened 2019-09-09
Add default LSF account for Lassen

timmoon10:
To make the convenience drivers more convenient.


ndryden:
Most people don't have access to the guests account, last time I checked. So I'm not sure if this is actually helpful.

PR #1234, opened 2019-09-10
Changed LBANN_ERRORs to variadic format

davidHysom:


benson31:
Thanks for doing this clean-up! I'm glad someone reads my comments.

For future reference, `std::to_string` is not needed for things with `operator<<(std::ostream&)` defined. Under the hood, the `LBANN_ERROR` macro forwards the arguments to a "string builder" function that will build the string using `operator<<` to handle that conversion.

davidHysom:
arugh! I just looked for all the '+' and changed them -- fogot to do away with the to_strings

benson31:
Yeah, I think it's a win to get rid of the `+`. Nothing wrong with incremental improvements.

PR #1237, opened 2019-09-11
Implementation of argmax and tovec layers

samadejacobs:
Two layers implemented in this PR:

(1) tovec - takes a label/index etc from previous layer and converts to a vector
(2) argmax - argmax of a column vector

### DIFF
--- a/include/lbann/execution_contexts/execution_context.hpp
+++ b/include/lbann/execution_contexts/execution_context.hpp
@@ -38,7 +38,8 @@ namespace lbann {
 // Forward-declare this.
 class trainer;
 
-struct termination_criteria {
+class termination_criteria {
+public:
   size_t num_steps;
 };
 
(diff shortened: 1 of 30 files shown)


timmoon10:
I've renamed `tovec_layer` to `one_hot_layer` to match [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/one_hot) (I can't find anything equivalent in PyTorch) and added an argmin layer. All of these layers pass unit tests with metric checking. I'm thinking of putting in a GPU implementation before merging them.

ndryden:
> I can't find anything equivalent in PyTorch

PyTorch deliberately avoids one-hot representations, so I don't think you will.

timmoon10:
I've done some work to implement the argmax layer on GPU, but I think it's better to defer that for later. Here's what I have so far: https://github.com/timmoon10/lbann/blob/gpu-argmax-layer/src/layers/misc/argmax.cu.

PR #1238, opened 2019-09-11
Return LBANN exit code in Python-generated batch script

timmoon10:
A batch script created by the Python `lbann.run` function looks something like:
```bash
date | sed "s/^/Started at /"
srun lbann
date | sed "s/^/Finished at /"
```
A bash script returns the exit code of its last command, so this script always returns 0. This is problematic for #1105, where we want to interrogate the exit status to determine if a unit test passed or not. The fix is simple:
```bash
date | sed "s/^/Started at /"
srun lbann
status=$?
date | sed "s/^/Finished at /"
exit ${status}
```


timmoon10:
I ran my unit test for the channel-wise scale/bias layer and got an exit code of 0. When I change the metric checking to expect a bogus result, I get an exit code of 1.

PR #1247, opened 2019-09-12
Refactor weights and objective fns to get comm at setup

bvanessen:
Updated both the weights an objective functions to get their comm object pointers after construction.  This is intended to make it easier to serialize and deserialize these classes with Cereal.

### DIFF
--- a/include/lbann/weights/weights.hpp
+++ b/include/lbann/weights/weights.hpp
@@ -77,7 +77,10 @@ public:
   /** Get weights name. */
   std::string get_name() const { return m_name; }
 
-  lbann_comm* get_comm() const { return m_comm; }
+  lbann_comm& get_comm() const {
+    if(m_comm == nullptr) { LBANN_ERROR("weights class has null comm pointer"); }
+    return *m_comm;
+  }
 
   /** Create a copy of the weights.
    *  This function dynamically allocates memory for a weights
@@ -170,7 +173,7 @@ public:
   // -----------------------------------------------
   // Initialize the weight class
   // -----------------------------------------------
-  void init(lbann_comm& comm);
+  void set_comm(lbann_comm& comm);
 
   // -----------------------------------------------
   // Setup
--- a/include/lbann/weights/weights.hpp
+++ b/include/lbann/weights/weights.hpp
@@ -77,6 +77,8 @@ public:
   /** Get weights name. */
   std::string get_name() const { return m_name; }
 
+  lbann_comm* get_comm() const { return m_comm; }
+
   /** Create a copy of the weights.
    *  This function dynamically allocates memory for a weights
    *  instance and instantiates a copy. The caller is responsible for
--- a/include/lbann/weights/weights.hpp
+++ b/include/lbann/weights/weights.hpp
@@ -63,9 +63,16 @@ class optimizer;
  */
 class weights {
   friend class optimizer;
+private:
+  weights();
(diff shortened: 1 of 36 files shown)


include/lbann/weights/weights.hpp
```diff
@@ -77,6 +77,8 @@ class weights {
   /** Get weights name. */
   std::string get_name() const { return m_name; }
 
+  lbann_comm* get_comm() const { return m_comm; }
```
timmoon10: We should return a reference. If `m_comm` is a null pointer, we should throw an exception.

include/lbann/weights/weights.hpp
```diff
@@ -165,6 +167,11 @@ class weights {
   El::DistData get_matrix_distribution() const;
   void set_matrix_distribution(El::DistData dist);
 
+  // -----------------------------------------------
+  // Initialize the weight class
+  // -----------------------------------------------
+  void init(lbann_comm& comm);
```
timmoon10: It's not clear what the difference is between `init` and `setup`. We should call it `set_comm`.

include/lbann/weights/weights.hpp
bvanessen: Done.

include/lbann/weights/weights.hpp
bvanessen: Good idea. Done.

include/lbann/weights/weights.hpp
```diff
@@ -63,6 +63,13 @@ class optimizer;
  */
 class weights {
   friend class optimizer;
+private:
+  weights();
+  // -----------------------------------------------
+  // Internal method for setting the comm pointer
+  // -----------------------------------------------
+  void set_comm(lbann_comm& comm);
```
benson31: This should probably take a `shared_ptr`.

src/weights/weights.cpp
```diff
@@ -62,28 +62,23 @@ std::string get_dims_string(const std::vector<int>& matrix_height_dims,
...
-  m_matrix_dist.blockHeight = 1;
-  m_matrix_dist.blockWidth = 1;
-  m_matrix_dist.colAlign = 0;
-  m_matrix_dist.rowAlign = 0;
-  m_matrix_dist.colCut = 0;
-  m_matrix_dist.rowCut = 0;
-  m_matrix_dist.root = 0;
-  m_matrix_dist.grid = &(comm->get_trainer_grid());
-  m_matrix_dist.device = El::Device::CPU;
+weights::weights(lbann_comm* comm)
```
benson31: Again this should probably be `shared_ptr`.

src/weights/weights.cpp
```diff
@@ -254,6 +249,25 @@ void weights::set_matrix_distribution(El::DistData dist) {
   m_matrix_dist = dist;
 }
 
+void weights::set_comm(lbann_comm& comm) {
```
benson31: Two things: 
1. This should probably be a `shared_ptr`.
2. Should this update the matrix distribution?

include/lbann/weights/weights.hpp
bvanessen: Lets defer this.

src/weights/weights.cpp
bvanessen: Deferred to future PR.

PR #1251, opened 2019-09-14
Clean up protobuf message for fully-connected layer

timmoon10:
### Summary

I've gone through and removed a bunch of fields from the FC layer protobuf message. 

Making progress toward #297.

### Unused fields

Users could set these fields, but they didn't affect anything:

- `weight_initialization`
- `bias_initial_value`
- `l2_regularization_factor`
- `double group_lasso_regularization_factor`
- `get_input_dimension_from_reader`
- `get_image_and_scalar_dimension_from_reader`
- `get_image_dimension_from_reader`
- `get_scalar_dimension_from_reader`

### (Mostly) JAG-specific fields

The following fields were used to get dimensions from the data reader:

- `num_neurons_is_num_labels`
- `get_num_neurons_of_slice_from_reader`
- `get_slice_points_from_reader`

This required a lot of complicated interaction between the prototext parser and the data reader (mostly `data_reader_jag_conduit`). It's more general to use the `hint_layer` field, which sets the output dimensions to match another layer. Importantly, this functionality can be used for any layer with user-selected output dimensions, not just the FC layer.

From what I can tell, we implemented the "get slice point from reader" functionality to get around the fact that the data reader can only output one data tensor (plus one label tensor). The JAG models require three data tensors: image, scalars, inputs. The hack was to flatten the tensors, concatenate them in the data reader, and apply a slice layer to recover the original tensors. This works, but a better approach would be to refactor the data reader so it can emit an arbitrary number of output tensors. Once this is done, the `hint_layer` approach becomes easier since you can just set the output of the input layer as the hint.

### DIFF
--- a/include/lbann/layers/learning/fully_connected.hpp
+++ b/include/lbann/layers/learning/fully_connected.hpp
@@ -42,6 +42,12 @@ namespace lbann {
  *  optionally applies an entry-wise bias. Following the
  *  column-vector convention:
  *    @f[ y = W * \text{vec}(x) + b @f]
+ *
+ *  Two weights are required if bias is applied: the linearity and the
+ *  bias. Only the linearity weights are required if bias is not
+ *  applied. If weights aren't provided, the linearity weights are
+ *  initialized with He normal initialization and the bias weights are
+ *  initialized to zero.
  */
 template <data_layout T_layout, El::Device Dev>
 class fully_connected_layer : public learning_layer {
--- a/include/lbann/layers/learning/fully_connected.hpp
+++ b/include/lbann/layers/learning/fully_connected.hpp
@@ -36,7 +36,13 @@
 
 namespace lbann {
 
-/** @brief Perform an affine transformation. */
+/** @brief Affine transformation
+ *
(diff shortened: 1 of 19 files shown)


timmoon10:
@samadejacobs Would you mind running a small JAG model and checking if it produces a reasonable result?

davidHysom:
I'm opposed to the name "hint_layer"  (even though I don't know what it does).What happens if the "hint" works in one run, but fails in the next? How does this affect reproduciblity? How does it affect checkpointing?

timmoon10:
The "hint layer" functionality could use some refinement and documentation. The idea is that it is an expert feature so that a layer can get clues about its output dimensions by querying another layer (see #703). I've left the name vague since I could imagine you could do advanced layer-specific stuff (e.g. an unpooling layer needs a pointer to a pooling layer), but perhaps this is not a good path to go down. I'm amenable to ideas.

In any case, it shouldn't affect reproducibility. The hints work based on layer names, which are consistent from run to run. I don't see any issues with checkpointing since this is only used for setup, and checkpointing happens after setup.

PR #1255, opened 2019-09-18
Cleanup protobuf message for trainer

timmoon10:
I've done three things to the trainer protobuf message:
- Documented the fields.
- Removed the `num_gpus` field, which was unused.
- Renamed the `block_size` field to `hydrogen_block_size` and made it so that a value of zero does nothing (i.e. uses Hydrogen's default block size).

Previously, the empty trainer message was invalid because a block size of zero is invalid. This resulted in us putting `block_size: 256` in all of our prototexts. The new approach makes that unnecessary and I've gone through and removed all instances (except where a non-default block size is requested). [The default block size in Hydrogen is 128](https://github.com/LLNL/Elemental/blob/753f83e1d0cb16292f0bab28c08e81c32964c541/src/core/environment.cpp#L306), so we may want to change that to 256 if that is a better default value.


timmoon10:
The CI tests pass on Pascal and Catalyst. See https://lc.llnl.gov/bamboo/browse/LBANN-TIM214-2.

model_zoo/tests/data_reader_tests/jag_single_layer_ae.prototext
```diff
@@ -4,10 +4,7 @@ model {
   serialize_io: true
```
bvanessen: This prototext does not have a trainer block. Now it would be empty, but other files have empty blocks.

PR #1261, opened 2019-09-21
Refactor add data type layer

bvanessen:
Refactoring the layer class hierarchy to be templated by a
TensorDataType.  Added a data_type_layer in-between the Layer class
and all children.  The data_type_layer holds the activations, error,
and weight matrices.  It is the first class in the hierarchy that is
specialized by TensorDataType.


davidHysom:
wae.prototext, and possibly other models, are broken. I'd like to ensure that all models (in <lbann>/model_zoo/models/...)  work before this, or any other, PR, is merged.

include/lbann/layers/layer.hpp
```diff
@@ -203,6 +202,11 @@ class Layer {
   /** Get child layers. (const) */
   inline const std::vector<const Layer*>& get_child_layers() const { return m_child_layers; }
 
+  inline int find_layer_index(const Layer* l) const {
+    return (std::find(m_child_layers.begin(),
+                      m_child_layers.end(),
+                      l) - m_child_layers.begin()); }
```
benson31: This should return `size_t` -- can't have negative indices! Also, I would recommend using `std::distance` here (can't do multiline suggestions... thanks GH...):
```cpp
inline size_t find_layer_index(const Layer* l) const {
    return std::distance(m_child_layers.begin(),
        std::find(m_child_layers.begin(),
                      m_child_layers.end(),
                      l)); }
```
This suggestion is more of a "straight refactor"; in actuality, you may want to assert that you have child layers (otherwise you'll return `0`, which _is_ a valid index but not what you mean) and that the search is successful, or you'll get `size()`, which is _not_ a valid index. Perhaps code like:
```cpp
inline size_t find_layer_index(const Layer* l) const {
    LBANN_ASSERT(m_child_layers.size());
    auto it = std::find(m_child_layers.begin(), m_child_layers.end(), l);
    LBANN_ASSERT(it != m_child_layers.size());
    return std::distance(m_child_layers.begin(), it);
}
```
(Note: `LBANN_ASSERT` is not a real thing, but it should be.)

include/lbann/weights/data_type_weights.hpp
```diff
@@ -104,11 +104,11 @@ class data_type_weights : public weights {
   /** Get weights optimizer.
    *  Returns a null pointer if the weights are frozen.
    */
-  data_type_optimizer<TensorDataType>* get_optimizer() override;
+  optimizer* get_optimizer() override;
   /** Get weights optimizer.
    *  Returns a null pointer if the weights are frozen.
    */
-  const data_type_optimizer<TensorDataType>* get_optimizer() const override;
+  const optimizer* get_optimizer() const override;
```
benson31: Why did you make this change?

src/layers/activations/activations.cpp
```diff
@@ -32,141 +32,151 @@ namespace lbann {
...
 namespace {
 
-// Helpful constants
-constexpr DataType zero = 0;
-constexpr DataType one = 1;
-constexpr DataType eps = std::numeric_limits<DataType>::epsilon();
-
 // =========================================================
-// Operator objects for entry-wise unary layers
+}// Operator objects for entry-wise unary layers
```
benson31:
```suggestion
// Operator objects for entry-wise unary layers
```

src/layers/activations/activations.cu
```diff
@@ -41,124 +41,134 @@ namespace {
...
-  template <>                                                           \
-  void layer<data_layout::DATA_PARALLEL, El::Device::GPU>               \
-  ::bp_compute() {                                                      \
-    cuda::apply_entrywise_binary_operator<op>(get_prev_activations(),   \
-                                              get_prev_error_signals(), \
-                                              get_error_signals());     \
-  }                                                                     \
+#define INSTANTIATE(layer, op)                                                                   \
+  template <typename TensorDataType>                                                             \
+  void fp_compute_inst(layer<TensorDataType, data_layout::MODEL_PARALLEL, El::Device::GPU>& l) { \
```
benson31:
```suggestion
  void fp_compute_impl(layer<TensorDataType, data_layout::MODEL_PARALLEL, El::Device::GPU>& l) { \
```

src/layers/activations/activations.cu
```diff
@@ -41,124 +41,134 @@ namespace {
...
+    cuda::apply_entrywise_unary_operator<TensorDataType, op<TensorDataType>>(l.get_prev_activations(), \
+                                                                             l.get_activations()); \
+  }                                                                                              \
+  template <typename TensorDataType>                                                             \
+  void bp_compute_inst(layer<TensorDataType, data_layout::MODEL_PARALLEL, El::Device::GPU>& l) { \
```
benson31:
```suggestion
  void bp_compute_impl(layer<TensorDataType, data_layout::MODEL_PARALLEL, El::Device::GPU>& l) { \
```

bvanessen:
So I disagree with part of your assessment here.  I do not think that a significant portion of the PR is related to the movement of get_activations or other related functions.  At the moment, I think that having the tensor access functions in the data_type_* classes is the most natural refactor as it doesn't require casting objects that get passed around.

> A major problem is that the bulk of this PR's line count is a consequence of changing the layer's API, mainly moving the tensor access functions (`get_activations`, `get_error_signals`, etc.) to `data_type_layer`. These functions really belong at the base layer class since we want layers with different types to interact, so we will need to undo much of the PR in the future.
> 
> In my opinion, starting from scratch will get us to mixed-precision training faster and with fewer headaches, but merging the PR does have the benefit of consolidating a lot of miscellaneous code cleanups. Per our discussion, merging is the path we are going with. I don't see any glaring errors, so we should merge it soon if the tests pass (including the weekly integration tests).
> 
> A proposed roadmap:
> 
> 1. Restore the tensor access functions in the base `Layer` class. They will be virtual and return `El::AbstractDistMatrix<DataType>`.
> 2. Restore the tensor access functions in the base `weights` class.
> 3. Undo all type-casts to `data_type_layer`.
> 4. Undo all type-casts to `data_type_weights`.
> 5. Remove `data_type_optimizer` and make optimizers independent of datatype.
> 6. Change the layer’s tensor access functions to return `El::BaseDistMatrix`. We’ll need to modify all the layers to cast their inputs to the correct datatype using proxies.
> 7. Change the weights' tensor access functions to return `El::BaseDistMatrix`.
> 8. Remove typedefs that assume a singular datatype, e.g. `DataType`, `AbsDistMat`, `CPUMat`.
> 9. Add support for layers and weights in fp64.
> 10. Add support for layers and weights in fp16.

PR #1263, opened 2019-09-23
Make jag-utils optional in the build/install

benson31:
Building the JAG utils wastes a significant chunk of time for the 99% who don't use them. This moves them out of the `all` target and into the `jag-utils` target.

The JAG utils will still be built and installed as part of the superbuild (and, by extension, by `build_lbann_lc.sh`), so most users should not be affected.

If you're in an LBANN build directory, passing the target name `jag-utils` to your build system of choice will build the executables in `model_zoo/jag_utils/` directory. These executables will only be installed as part of the `install` target if they have been built previously. For example:
```
cmake <options> ${LBANN_SRC_DIR}
cmake --build ${LBANN_BUILD_DIR} # Build the library and main executables
cmake --build ${LBANN_BUILD_DIR} --target jag-utils # Build the JAG utilities
cmake --build ${LBANN_BUILD_DIR} --target install # Install all built targets
```


benson31:
"Significant" in a relative and long-term sense. They take something like 8 seconds to build in parallel, *but* they depend on the `lbann` target. If someone is doing LBANN development, not JAG work, they pay that 8s every time they touch *any* LBANN file and recompile. Since I keep hearing more and more complaints about the long compile times, I'm trimming any proverbial baggage that can be trimmed and this seemed like low-hanging fruit.

davidHysom:
> they pay that 8s every time they touch _any_ LBANN file and recompile

Hm, I must have a different workflow. When after I edit a file I do make in someplace like this:
build/gnu.Release.pascal.llnl.gov/lbann/build/
And only the file (if its a .cpp) is re-compiled

benson31:
Ah yes, that's a good point. Rebuilding `.cpp` files only should only cause those targets to be relinked, not rebuilt. Various parts of LBANN, however, have significant header implementations (e.g., layers). That's where I've been working lately, so I'm probably feeling it more.

benson31:
> In a future step, could we move all of this out of LBANN into a JAG-specific Bitbucket?

Please yes. I would like to have a dependent project for LBANN so we can keep the `install` target in check. Admittedly, I don't test it all that often.

davidHysom:
> In a future step, could we move all of this out of LBANN into a JAG-specific Bitbucket?

The drivers make heavy use of the options class, so that would need to be replicated. Also all the calls to lbann_comm. I think that's about all that the drivers depend on from lbann. And, someone would need to set up a build system, and all that good stuph that's not really in my task set.

timmoon10:
The JAG repo could have LBANN as a dependency, so it would be free to use the options class and `lbann_comm` without needing to duplicate them. Implementing the build system would be non-trivial though.

benson31:
> Implementing the build system would be non-trivial though.
With comments and some readability-whitespace/linebreaks, it's 37 lines.
```
cmake_minimum_required(VERSION 3.14.0)

project(JAG_UTILS CXX)

# LBANN is the only dependency; find it.
find_package(LBANN NO_MODULE
  HINTS ${LBANN_DIR} $ENV{LBANN_DIR}
  PATH_SUFFIXES lib64/cmake/lbann lib/cmake/lbann
  NO_DEFAULT_PATH)
find_package(LBANN NO_MODULE REQUIRED)

# Use (i.e. don't skip) RPATH for build
set(CMAKE_SKIP_BUILD_RPATH FALSE)

# Use same RPATH for build and install
set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)

# Add the automatically determined parts of the RPATH
# which point to directories outside the build tree to the install RPATH
set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)

# Build stuff.
list(APPEND SOURCE_FILES select_samples.cpp) #etc
foreach(source_file IN LISTS SOURCE_FILES)
  get_filename_component(EXE_NAME "${source_file}" NAME_WE)
  add_executable(${EXE_NAME} ${source_file})
  target_link_libraries(${EXE_NAME} PRIVATE LBANN::lbann)
  list(APPEND TARGETS_TO_INSTALL ${EXE_NAME})
endforeach()

# Install stuff.
include(GNUInstallDirs)
install(
  TARGETS ${TARGETS_TO_INSTALL}
  EXPORT FAKETargets
  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
  )
```

bvanessen:
My first thought is no. We don’t want to pull this out into a separate repo. I will compose a more complete answer later.


Brian C. Van Essen

vanessen1@llnl.gov

(w) 925-422-9300

(c) 925-290-5470

---

Sent from my iPhone

On Sep 25, 2019, at 10:24 AM, Tom Benson <notifications@github.com> wrote:

﻿

Implementing the build system would be non-trivial though.
With comments and some readability-whitespace/linebreaks, it's 37 lines.

cmake_minimum_required(VERSION 3.14.0)

project(JAG_UTILS CXX)

# LBANN is the only dependency; find it.
find_package(LBANN NO_MODULE
  HINTS ${LBANN_DIR} $ENV{LBANN_DIR}
  PATH_SUFFIXES lib64/cmake/lbann lib/cmake/lbann
  NO_DEFAULT_PATH)
find_package(LBANN NO_MODULE REQUIRED)

# Use (i.e. don't skip) RPATH for build
set(CMAKE_SKIP_BUILD_RPATH FALSE)

# Use same RPATH for build and install
set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)

# Add the automatically determined parts of the RPATH
# which point to directories outside the build tree to the install RPATH
set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)

# Build stuff.
list(APPEND SOURCE_FILES select_samples.cpp) #etc
foreach(source_file IN LISTS SOURCE_FILES)
  get_filename_component(EXE_NAME "${source_file}" NAME_WE)
  add_executable(${EXE_NAME} ${source_file})
  target_link_libraries(${EXE_NAME} PRIVATE LBANN::lbann)
  list(APPEND TARGETS_TO_INSTALL ${EXE_NAME})
endforeach()

# Install stuff.
include(GNUInstallDirs)
install(
  TARGETS ${TARGETS_TO_INSTALL}
  EXPORT FAKETargets
  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
  )

PR #1269, opened 2019-09-26
Nvshmem

gunney1:
This is my progress toward using an NVSHMEM collective in the batch_normalization.cu code.
To configure with NVSHMEM, add these switches to the superbuild command:
-D LBANN_WITH_NVSHMEM:BOOL=ON
-D LBANN_SB_FWD_LBANN_NVSHMEM_DIR:PATH=/usr/workspace/brain/nvshmem/nvshmem_0.2/EA_0.2.5_Release-BrianG/nvshmem_0.2.5-0/build

Unfortunately, there's an error in the NVSHMEM code, complaining that identifier "__ldg" is undefined.  __ldg is a symbol in one of the NVSHMEM headers.


naoyam:
`__ldg` is a CUDA intrinsic function. It may not be defined if the device compute version is very old, but then NVSHMEM only supports the newest versions. We probably should set the version at least when using NVSHMEM.

PR #1270, opened 2019-09-26
Get Bamboo to turn green

benson31:
This changes some things and makes my button thingy on bamboo turn green. Yay.


bamboo/compiler_tests/build_script.sh
```diff
@@ -1,4 +1,69 @@
...
+
 LBANN_DIR=$(git rev-parse --show-toplevel)
-${LBANN_DIR}/scripts/build_lbann_lc.sh --with-conduit
+CLUSTER=$(hostname | sed 's/[0-9]*//g')
+USER=$(whoami)
+WORKSPACE_DIR=$(ls --color=no -d /usr/workspace/*/${USER})
+DEPENDENCY_DIR=${WORKSPACE_DIR}/stable_dependencies/${CLUSTER}/gcc-7.3.0/mvapich2-2.3
+if [ -e ${DEPENDENCY_DIR} ];
+then
+    ml gcc/7.3.0 cmake/3.14.5 cuda
```
bvanessen: Shouldn't this load a specific cuda since the following NCCL requires a particular one.

bamboo/run.sh
```diff
@@ -39,15 +39,14 @@ module load cmake/3.9.2
...
-
-cd ..
+# echo "Task: Integration Tests"
+# cd integration_tests
+# if [ ${WEEKLY} -ne 0 ]; then
+#     $PYTHON -m pytest -s -vv --durations=0 --weekly --junitxml=results.xml
+# else
+#     $PYTHON -m pytest -s -vv --durations=0 --junitxml=results.xml
+# fi
+# cd ..
```
bvanessen: do we want to fix these before merging this PR, or after.

bamboo/unit_tests/test_unit_checkpoint.py
```diff
@@ -22,7 +22,7 @@ def skeleton_checkpoint_lenet_shared(cluster, executables, dir_name,
         cluster=cluster, executable=exe, num_nodes=1, num_processes=2,
         dir_name=dir_name,
         data_filedir_default='/p/lscratchh/brainusr/datasets/MNIST',
-        data_reader_name='mnist', data_reader_percent=data_reader_percent,
+        data_reader_name='mnist', data_reader_percent=0.01,
```
bvanessen: Are we always doing this at 1% even for weekly?

bamboo/unit_tests/test_unit_mnist_softmax_classifier.py
```diff
@@ -18,7 +18,7 @@ def skeleton_mnist_softmax_classifier(cluster, executables, dir_name, compiler_n
         num_processes=1, dir_name=dir_name,
         data_filedir_default='/p/lscratchh/brainusr/datasets/MNIST',
         data_reader_name='mnist',
-        data_reader_percent=data_reader_percent,
+        data_reader_percent=0.01,
```
bvanessen: Same question as above.  Is this overridden by the normal weekly / nightly percentage.

benson31: This change will be reverted and then this test is moving to `weekly`-only. As @timmoon10 noted, it's "parameter bound", not "data bound" so changing this value actually didn't affect anything.

bamboo/compiler_tests/build_script.sh
benson31: Yes; I'll fix this when I update this script to work on other clusters (it only works on `pascal` right now), which, I assume, will be before CUDA 11 comes out.

bamboo/compiler_tests/test_compiler.py
```diff
@@ -80,7 +80,7 @@ def skeleton_clang6(cluster, dir_name, debug):
 
 
 def skeleton_gcc7(cluster, dir_name, debug):
-    if cluster not in ['catalyst', 'pascal']:
+    if cluster not in []:
```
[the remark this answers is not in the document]
benson31: I was copying the other things in the file. I have minimal interest in being pythonic. Is every object really convertible to `bool`? If this is "right", I have no issue with it. And as with all things python: there is but ONE TRUE PATH.

bamboo/allocate_and_run.sh
```diff
@@ -53,7 +53,7 @@ elif [ "${CLUSTER}" = 'catalyst' ] || [ "${CLUSTER}" = 'corona' ] || [ "${CLUSTE
             cd ..
         fi
     else
-        timeout -k 5 24h salloc -N16 --partition=pbatch -t $ALLOCATION_TIME_LIMIT ./run.sh
+        timeout -k 5 24h salloc -N2 --partition=pvis -t 120 ./run.sh
```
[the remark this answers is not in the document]
benson31: Yeah, this is part of a `[temporary]` commit; I'll revert it before merging.

bamboo/unit_tests/test_unit_check_proto_models.py
```diff
@@ -66,6 +66,7 @@ def skeleton_models(cluster, dir_name, executables, compiler_name,
                         time_limit = 3
                     if 'resnet50' in file_name:
                         node_count = 8
+                        continue
```
forsyth2: If you want to skip `resnet50` probably best to add comment of why.

bamboo/unit_tests/test_unit_checkpoint.py
forsyth2: Same question as Brian.

Possible alternatives:
-Have Nightly run at 10% of passed-in value and Weekly at 100% (requires parsing prototext to get `data_reader_percent`)
-Rework the logic in `tools.get_command` to run 100% for Weekly no matter what and then use the prototext value for Nightly.

bamboo/unit_tests/test_unit_checkpoint.py
```diff
@@ -37,7 +37,7 @@ def skeleton_checkpoint_lenet_shared(cluster, executables, dir_name,
...
```
forsyth2: Since this is repeated multiple times, I would store it in a variable, so if a change is necessary then we only have to do it in one place.

bamboo/unit_tests/test_unit_lbann2_reload.py
```diff
@@ -37,33 +42,49 @@ def skeleton_lbann2_reload(cluster, executables, dir_name, compiler_name,
...
+        executable=lbann2,
+        num_nodes=1,
+        num_processes=2,
+        ckpt_dir=ckpt_dir,
         data_filedir_default='/p/lscratchh/brainusr/datasets/MNIST',
-        data_reader_name='mnist', data_reader_percent=data_reader_percent,
-        ckpt_dir=ckpt_dir, model_folder='tests',
-        model_name='lenet_mnist_ckpt', num_epochs=2, optimizer_name='sgd',
+        data_reader_name='mnist',
+        data_reader_percent=0.005,
```
forsyth2: Since this is repeated multiple times, I would store it in a variable, so if a change is necessary then we only have to do it in one place.

bamboo/unit_tests/test_unit_mnist_softmax_classifier.py
```diff
@@ -47,6 +47,10 @@ def test_unit_mnist_softmax_classifier_intel19(cluster, exes, dirname,
 # Run with python3 -m pytest -s test_unit_softmax_classifier.py -k 'test_unit_mnist_softmax_classifier_exe' --exe=<executable>
 def test_unit_mnist_softmax_classifier_exe(cluster, dirname, exe,
                                            weekly, data_reader_percent):
+    if not weekly:
```
forsyth2: To be consistent with other tests:
```
  if not weekly:
    e = 'test_unit_mnist_softmax_classifier: Not doing weekly testing'
    print('Skip - ' + e)
    pytest.skip(e)
```

src/proto/proto_common.cpp
```diff
@@ -896,9 +896,15 @@ void print_help(std::ostream& os)
...
        "  --ckpt_dir=<string>\n"
-       "      Save to or reload from a specific checkpoint directory.\n"
+       "      Save to or restart from a specific checkpoint directory.\n"
        "      Additionally, sets the output directory for dumping weights.\n"
        "      Modifies callbacks: checkpoint, save_model, dump_weights\n"
+       "  --restart_dir=<string>\n"
+       "      Restart from a checkpoint found in the given directory.\n"
+       "      If the directory doesn't exist or doesn't contain a checkpoint,\n"
+       "      an error will be thrown.\n"
+       "  --restart_dir_is_fullpath=<bool>\n"
```
forsyth2: Add this to `allowed_flags` in `tools.get_command`.

bamboo/unit_tests/test_unit_checkpoint.py
benson31: I don't think this test in particular should ever run at any value other than "1 sample" -- this test should be data-invariant; as such, nothing new is tested by running 1 billion samples vs a single sample. As discussed with @bvanessen, I will update this to allow a value to be taken in but if `data_reader_percent` is `None`, then it will hard-code `0.01`. Does this seem reasonable?

bamboo/common_python/tools.py
```diff
@@ -494,8 +507,9 @@ def get_command(cluster,
     if lbann_errors != []:
         print('lbann_errors={lbann_errors}.'.format(lbann_errors=lbann_errors))
         raise Exception('Invalid Usage: ' + ' , '.join(lbann_errors))
-    command_lbann = '%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s' % (
-        executable, option_ckpt_dir, option_data_filedir,
+    command_lbann = '%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s%s' % (
+        executable, option_ckpt_dir, option_restart_dir, option_disable_cuda,
```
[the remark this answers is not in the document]
benson31: I've made this change, but the logic was that "checkpoint/restart" are kinda sorta intertwined, so I wanted to have them close to each other for the sake of someone reading the resulting command-line.

bamboo/common_python/tools.py
```diff
@@ -423,6 +431,10 @@ def get_command(cluster,
         option_processes_per_model = ' --procs_per_model=%d' % processes_per_model
     if ckpt_dir is not None:
         option_ckpt_dir = ' --ckpt_dir=%s' % ckpt_dir
+    if restart_dir is not None:
+        option_restart_dir = ' --restart_dir=%s' % restart_dir
+    if disable_cuda is not None:
+        option_disable_cuda = ' --disable_cuda=%d' % int(bool(disable_cuda))
```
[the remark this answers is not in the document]
benson31: Technically the parameter to `bool()` is optional, and the function is total. I don't think there's any way this value error could trigger.

bamboo/unit_tests/test_unit_checkpoint.py
bvanessen: @benson31 yes that seems reasonable.  @forsyth2 If a prototext specifies a data reader percentage, then I do not think that the test should override it.  Specifically, I can imagine many future tests that require a precise number of samples to check for weird corner cases.

bamboo/unit_tests/test_unit_lbann2_reload.py
benson31: Fixed.

bamboo/unit_tests/test_unit_checkpoint.py
benson31: Fixed.

bamboo/unit_tests/test_unit_checkpoint.py
benson31: @bvanessen @forsyth2 The caveat to that is that the data readers in `model_zoo/` might specify `data_reader_percent` for any number of reasons (e.g., the data set is huge and someone put a reasonable value for debugging and then aggressively committed). The testing infrastructure should be completely isolated from the rest of the world. I think that if a test requires a specific value, it should do one of the following, in decreasing order of desirability:
- Specify the value explicitly in PyTest.
- Copy the data reader file to a testing-specific location and use that with the `data_reader_percent='prototext'` option.
Moreover, if a test requires a specific amount of data to pass, users should not be able to command-line override that amount for those tests (because then the test will fail for a non-code reason (in particular, the contract of the test will have been broken)).

bamboo/unit_tests/test_unit_check_proto_models.py
benson31: I have updated this to just skip on non-`weekly` testing and I have added a comment to that effect.

bamboo/unit_tests/test_unit_mnist_softmax_classifier.py
benson31: Fixed.

src/proto/proto_common.cpp
benson31: Fixed.

bamboo/common_python/tools.py
```diff
@@ -19,6 +19,7 @@ def get_command(cluster,
                 time_limit=None,
                 # LBANN Parameters
                 ckpt_dir=None,
+                disable_cuda=None,
```
bvanessen: Does `disable_cuda` need to appear in the list down below?

bamboo/common_python/tools.py
```diff
@@ -60,7 +62,7 @@ def get_command(cluster,
         data_filename_test_default, data_reader_name, data_reader_path,
         data_reader_percent, exit_after_setup, metadata, mini_batch_size,
         model_folder, model_name, model_path, num_epochs, optimizer_name,
-        optimizer_path, processes_per_model,
+        optimizer_path, processes_per_model, restart_dir,
```
bvanessen: Specifically does `disable_cuda` need to appear here.

benson31: @forsyth2 I have no idea. Things are working with it the way it is, but perhaps I've done something wrong (and true to Python, it works anyway...).

bamboo/unit_tests/test_unit_reconstruction_loss.py
```diff
@@ -16,12 +16,13 @@ def skeleton_jag_reconstruction_loss(cluster, executables, dir_name, compiler_na
     command = tools.get_command(
         cluster=cluster,
         executable=executables[compiler_name],
-        num_nodes=16,
+        num_nodes=2,
         num_processes=32,
```
timmoon10:
```suggestion
         num_processes=4,
```

python/lbann/contrib/lc/systems.py
```diff
@@ -19,10 +19,12 @@ def __init__(self,
 
 # Supported LC systems
 _system_params = {'catalyst': SystemParams(24, 0, 'slurm', 'pbatch', 'brain'),
+                  'coronal':  SystemParams(24, 0, 'slurm', 'pbatch', None),
```
timmoon10:
```suggestion
                  'corona':  SystemParams(24, 0, 'slurm', 'pbatch', None),
```

Issue #1245, opened 2019-09-12
Update remaining unit tests to use Python Front-End


forsyth2:
#1105 converted one unit test to using the Python Front-End instead of `tools.get_command`. We still need to convert the remaining unit tests.

If possible, we may want to convert the integration tests to using the Python Front-End too.

Even if we remove all references to `tools.get_command` we may want to keep that function for local testing, since it can allocate nodes and determine what commands to run given the cluster it's running on.

PR #1271, opened 2019-09-27
Add layer tests to Bamboo

timmoon10:
Over the past few months, I've accumulated some tests to validate the following layers:

- argmax
- argmin
- channel-wise scale/bias
- embedding
- entry-wise batchnorm
- entry-wise scale/bias
- one-hot
- slice

These tests are implemented using the Python frontend and use the Python data reader to generate synthetic data. Most of these tests perform both metric checking and gradient checking. To compute ground-truth metric values, they perform the same operations with NumPy.

Now that our Bamboo workflow can accommodate the Python frontend (#1105), I've modified the Python data reader test (#1254) to implement these layer tests. This makes progress toward #1245. All these tests pass on Pascal and Catalyst.


bamboo/unit_tests/test_unit_layer_argmax.py
```diff
@@ -0,0 +1,232 @@
...
+current_dir = os.path.dirname(current_file)
+sys.path.insert(0, os.path.join(os.path.dirname(current_dir), 'common_python'))
+import tools
+
+# ==============================================
+# Objects for Python data reader
+# ==============================================
+# Note: The Python data reader imports this file and calls the
+# functions below to ingest data. This is the only part of the script
+# that should be executed when the script is imported, or else the
```
forsyth2: It looks like there quite a few lines in this file are run outside of functions (i.e. will run when imported).

Or do you mean only this part of some other script should be run on import here?

bamboo/unit_tests/test_unit_layer_argmax.py
```diff
@@ -0,0 +1,232 @@
...
+    data_reader.python.num_samples_function = 'num_samples'
+    data_reader.python.sample_dims_function = 'sample_dims'
+
+    return message
+
+# ==============================================
+# Setup PyTest
+# ==============================================
+
+# Generate test name based on file name
```
forsyth2: `# Set test name to be name of file excluding .py`

bamboo/unit_tests/test_unit_layer_argmax.py
```diff
@@ -0,0 +1,232 @@
...
+def _test_intel19(cluster, exes, dirname):
+    _test(cluster, exes, dirname, 'intel19')
+def _test_exe(cluster, dirname, exe):
+    if exe is None:
+        e = 'test_{}_exe: Non-local testing'.format(_test_name)
+        print('Skip - ' + e)
+        pytest.skip(e)
+    exes = {'exe': exe}
+    _test(cluster, exes, dirname, 'exe')
+globals()['{}_clang6'.format(_test_name)] = _test_clang6
```
forsyth2: Add comment above like `# Auto-generate test names by appending compiler to primary test name`

bamboo/unit_tests/test_unit_layer_argmax.py
timmoon10: Hm, looks like I need to word this better. My thinking is that the Python data reader only needs the stuff in this section when it imports this file, and it can ignore all of the model construction and PyTest stuff. In my original use-case, you would have two separate scripts to construct the model and to generate data, but we do this hacky approach so that everything conveniently fits in one script.

timmoon10:
I think consolidating the PyTest boilerplate would be a good idea. In fact, I think it could even be used for tests that don't use the Python frontend.

I'm more reticent about consolidating the data reader boilerplate, although this is mostly an gut feeling than anything logical. I feel we should sort out the problems in the data reader API before building too much infrastructure around it.

bvanessen:
> I think consolidating the PyTest boilerplate would be a good idea. In fact, I think it could even be used for tests that don't use the Python frontend.
> 
> I'm more reticent about consolidating the data reader boilerplate, although this is mostly a gut feeling than anything logical. I feel we should sort out the problems in the data reader API before building too much infrastructure around it.

I agree that we shouldn't bury it, but we are now creating ~7 copies of boilerplate code that we know that we want to replace.  Let's put it into a function, but not forget about it.  That way it is easier for us to get rid of later.

timmoon10:
I've consolidated some of the boilerplate for the PyTest infrastructure and the Python data reader. All of the tests still pass for me on Pascal and Catalyst.

bvanessen:
Thank you for the code refactoring.  Looks great.

bamboo/common_python/tools.py
```diff
@@ -657,3 +661,158 @@ def assert_failure(return_code, expected_error, error_file_name):
...
+        for t in _test_funcs:
+            globals()[t.__name__] = t
+
+    Args:
+        setup_func (function): Sets up an LBANN experiment using the
+            Python frontend. It takes in the LBANN Python module as
+            input and returns a `(lbann.Trainer, lbann.Model,
+            lbann.reader_pb2.DataReader, lbann.Optimizer)`.
+        test_name (str): Descriptive name. Should be prefixed with
+            'test_'.
```
forsyth2: Why don't we just have the function add the `test_` prefix?

bamboo/common_python/tools.py
```diff
@@ -657,3 +661,158 @@ def assert_failure(return_code, expected_error, error_file_name):
...
+            trainer=trainer,
+            model=model,
+            data_reader=data_reader,
+            optimizer=optimizer,
+            experiment_dir=experiment_dir,
+            job_name='lbann_{}'.format(test_name),
+            **kwargs)
+        assert_success(return_code, error_file_name)
+
+    # Specific test functions for different build configuration
```
forsyth2: `configurations`

bamboo/common_python/tools.py
forsyth2: Ah ok, if we extract the test name automatically as in `_test_name = os.path.splitext(os.path.basename(current_file))[0]`, then there's no reason to have `create_tests` add `test_`.

forsyth2:
Before merging though, can you squash/fixup the commits?

PR #1272, opened 2019-09-27
Replace READMEs with .gitignores in Bamboo log directories

timmoon10:
I've noticed that `bamboo/clean.sh` doesn't remove the experiment directory created by the unit test for the Python data reader. I tried fixing the script to remove everything in `bamboo/unit_tests/experiments` except `README.md`, but that gets rather hairy. It makes things much easier if we replace `README.md` with a `.gitignore`. I've gone ahead and made similar changes to the rest of the `bamboo` directory.


benson31:
Per a conversation with @bvanessen yesterday, these directories won't be used for this purpose, at least in user space, in the near future. This is good clean-up in the meantime, though.

forsyth2:
> Per a conversation with @bvanessen yesterday, these directories won't be used for this purpose, at least in user space, in the near future. This is good clean-up in the meantime, though.

Used for what purpose? Checking error/output logs from the build?

benson31:
If the directories aren't there, can't they just be made on-the-fly?

forsyth2:
> If the directories aren't there, can't they just be made on-the-fly?

Yes, I suppose, but since in theory the unit tests could run in any order, each would need to check if the directory exists and if not, then create it.

Also, it probably makes sense to have the `.gitignore` anyway, since we don't want git tracking the output/error logs from a particular build.

benson31:
> Used for what purpose? Checking error/output logs from the build?

The thought we discussed was essentially running all of these tests from the build directory. This would be a user-space (eh, "developer-space" change) that would encourage developers to run the tests as part of their build process and facilitate their doing so. I don't know that it would affect bamboo or not, which brings me to: ...

> Yes, I suppose, but since in theory the unit tests could run in any order, each would need to check if the directory exists and if not, then create it.

Why not just create the directory structure as part of `clean.sh` or something? That file already know what directories should be there (since it cleans all of them). In "bamboo-space", it won't matter that these dirty the git because bamboo doesn't commit anything (although I'd like it to -- that's another story though, which doesn't involve `git add -A` so the fact that the repo might be dirty is also moot here). It could cause problems in user-space; see the previous paragraph.

forsyth2:
@benson31 I like having a specific test directory (i.e. `bamboo`) but I suppose it could be easier for developers to have tests integrated into the build directory.

Yes, that would work. We could have `clean.sh` just make the directories we need after deleting the previous build's.

bamboo/clean.sh
```diff
@@ -6,19 +6,18 @@ LBANN_DIR=$(git rev-parse --show-toplevel)
 # Compiler Tests
```
forsyth2: Might be good to add a comment up at line 3 like `# Bamboo has an option to clean the directory after building, but that would remove log files that we may want to look at in between builds, so we need to clean the directory before building`. Otherwise, I could see someone one day thinking this `clean.sh` script isn't needed.

benson31:
> I like having a specific test directory
@forsyth2 Just to be clear, the proposal was not to eliminate the `bamboo` directory, just to make it easier for developers to run *and* get build-specific output (which lead me to believe that the right place for that is inside the relevant build directory.

For example, consider a developer building on Pascal and Lassen simultaneously. Triggering the tests locally on the respective machines simultaneously would result in a race on the output directories to which things currently are currently piped. The intention would be to create a directory structure inside the build that can be written to by the local tests and not interfere with or be interfered with by any other build.

PR #1285, opened 2019-10-09
Modified linear search through a vector to a map lookup

davidHysom:
This greatly speeds up sample_list_open_files<sample_name_t, file_handle_t>
::all_gather_packed_lists when dealing with large sample lists

### DIFF
--- a/include/lbann/data_readers/sample_list_open_files_impl.hpp
+++ b/include/lbann/data_readers/sample_list_open_files_impl.hpp
@@ -551,10 +551,11 @@ inline void sample_list_open_files<sample_name_t, file_handle_t>
         }
         mp[filename] = index;
       }else {
-        if (mp.find(filename) == mp.end()) {
+        auto search_result = mp.find(filename);
+        if (search_result == mp.end()) {
           LBANN_ERROR("mp.find(filename) == mp.end()");
         }
-        index = mp[filename];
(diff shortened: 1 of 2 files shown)


include/lbann/data_readers/sample_list_open_files_impl.hpp
```diff
@@ -548,13 +549,12 @@ ::all_gather_packed_lists(lbann_comm& comm) {
...
       }else {
-        for(size_t i = 0; i < m_file_id_stats_map.size(); i++) {
-          if(filename == get_samples_filename(i)) {
-            index = i;
-            break;
-          }
+        if (mp.find(filename) == mp.end()) {
+          LBANN_ERROR("mp.find(filename) == mp.end()");
         }
+        index = mp[filename];
```
benson31: This essentially does the lookup twice. GH finally has multiline comments, but not multiline "suggestions", so here's a codeblock instead:
```cpp
auto search_result = mp.find(filename);
if (search_result == mp.end()) {
  LBANN_ERROR("mp.find(filename) == mp.end()");
}
index = search_result->second;
```

bvanessen: As soon as @benson31 fix is applied, it looks good.

PR #1288, opened 2019-10-10
Fix for case where global archive size is > INT_MAX

davidHysom:
Modified sample_list<sample_name_t>::all_gather_archive() to deal wit…h the case where the size of the global archive is > INT_MAX.

To use the modification, include the cmd line flag: --all_gather_new

### DIFF
--- a/include/lbann/data_readers/sample_list_impl.hpp
+++ b/include/lbann/data_readers/sample_list_impl.hpp
@@ -12,6 +12,7 @@
 #include <unordered_set>
 #include <memory>
 #include <type_traits>
+#include <limits>
 
 #include <cereal/archives/binary.hpp>
 #include <unistd.h>
@@ -392,12 +393,14 @@ inline void sample_list<sample_name_t>
   // there's commented out code below to deal with the case where
   // archive.size() > INT_MAX; but for now let's assume we won't
   // encounter that (which is true for the 100M JAG set)
+  int constexpr max_int = std::numeric_limits<int>::max();
   size_t n = archive.size();
-  if (n > INT_MAX) {
-    LBANN_ERROR("(n > INT_MAX");
+  if (n > max_int
+    LBANN_ERROR("(n > max_int");
   }
 
-  // change int to size_t for case where n > INT_MAX
+  // change int to size_t for case where n > max_int (see commented out
+  // code block below)
   int size_of_my_archive= archive.size();
   std::vector<int> packed_sizes(comm.get_procs_per_trainer());
   comm.trainer_all_gather(size_of_my_archive, packed_sizes);
--- a/include/lbann/data_readers/sample_list_impl.hpp
+++ b/include/lbann/data_readers/sample_list_impl.hpp
@@ -337,6 +337,11 @@ inline void sample_list<sample_name_t>
 ::all_gather_archive(const std::string &archive,
                      std::vector<std::string>& gathered_archive,
                      lbann_comm& comm) {
+  if (options::get()->get_bool("all_gather_new")) {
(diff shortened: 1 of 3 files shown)


include/lbann/data_readers/sample_list_impl.hpp
```diff
@@ -378,6 +383,92 @@ ::all_gather_archive(const std::string &archive,
...
+inline void sample_list<sample_name_t>
+::all_gather_archive_new(const std::string &archive,
+                     std::vector<std::string>& gathered_archive,
+                     lbann_comm& comm) {
+
+  // there's commented out code below to deal with the case where
+  // archive.size() > INT_MAX; but for now let's assume we won't
+  // encounter that (which is true for the 100M JAG set)
+  size_t n = archive.size();
+  if (n > INT_MAX) {
```
benson31: I would recommend `#include <limits>` and then replace `INT_MAX` with `std::numeric_limits<int>::max()` (or define your own `constexpr int` that has value equal to `std::numeric_limits<int>::max()`. E.g.:
```cpp
int constexpr max_int = std::numeric_limits<int>::max();
```
`numeric_limits` is the C++ idiom for expressing this information, and since the result (of this particular function, anyway) is `constexpr`, it can be used anywhere the macro would be (templates, case statements, array dims, etc).

timmoon10:
We should probably discourage using `int` in favor of `size_t` and `El::Int`/`long long`. I can imagine we'll want to eventually push past 2.1B samples.

include/lbann/data_readers/sample_list_impl.hpp
benson31: In view of Tim's comment,
```
using int_type = /*whatever*/;
int_type constexpr max_value = std::numeric_limits<int_type>::max();
```
(Or template the function on the integer type and then just don't worry about it...)

benson31: Also also, comparing `size_t` to `int` will give a compiler warning from worthwhile compilers.

davidHysom:
> We should probably discourage using `int` in favor of `size_t` and `El::Int`/`long long`. I can imagine we'll want to eventually push past 2.1B samples.

Where, specifically, in this PR are you referring to? Sometimes we must use `int` because MPI demands that we do so.

timmoon10:
My comment is more of a comment on the whole of LBANN. We should prefer using 64-bit integers, and cast them down to 32-bit whenever we need to interact with MPI or other packages that expect `int`s. For the moment, I think throwing an exception when we detect large integer values is sensible.

benson31:
The only places where MPI _really_ needs `int`s to be fed into it are pointer arguments, e.g.,
```cpp
size_t rank;
MPI_Comm_rank(MPI_COMM_WORLD, &rank);
```
won't compile. But things like
```cpp
size_t size;
data_type* buf;
MPI_Bcast(buf, safe_cast<int>(size), TypeMap<data_type>(), rank, comm);
```
should be fine, where `safe_cast` is something like:
```cpp
template <typename OutT, typename InT>
OutT safe_cast(InT const& x)
{
  OutT x_cast = static_cast<OutT>(x);
  if (x != static_cast<InT>(x_cast))
    LBANN_ERROR("safe_cast failed");
  return x_cast;
}
```
I say "something like" because this snippet doesn't take into account differently signed integer types where wrap-around could cause the comparison to erroneously pass (e.g. `static_cast<int>(static_cast<size_t>(-1)) == -1`, but `size_t` cannot represent `-1`). The bigger hassle will be converting displacement arrays and the like from `size_t[]` to `int[]` before passing to MPI, though the cost should be minimal relative to the communication itself.

PR #1292, opened 2019-10-14
Support "optional" message fields in Protobuf API

timmoon10:
protobuf 3 does not support optional primitive fields. If a field is not set, it is interpreted as "zero" (`float`s are `0.0f`, `bool`s are `false`, etc.). This is inconvenient when you have a field where zero is a valid value and the default should be non-zero. For instance, I would like to add support to the embedding layer to ignore indices that are equal to a "pad index". Index 0 is a perfectly reasonable pad index (and probably a common one), so a good default would be something like -1.

Note that this only applies to primitive fields. There is still machinery to detect if a message field is set or not. Furthermore, protobuf already defines [wrapper messages for the primitive data types](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto). This means we can specify an optional field in a `.proto` file like this:
```
  message Embedding {
    int64 num_embeddings = 1;
    int64 embedding_dim = 2;
    google.protobuf.Int64Value padding_idx = 3;
  }
```
These wrapper messages are only useful for singular fields, since they will always be set inside repeated fields. I agree that this is somewhat difficult to interpret. In a more ideal world, protobuf would support typedefs and we could rename it to something like `OptionalInt64`. A corresponding prototext file looks like:
```
# With pad index
layer {
  embedding {
    num_embeddings: 1000000
    embedding_dim: 50
    padding_idx { value: 0 }

  }
}

# Without pad index
layer {
  embedding {
    num_embeddings: 1000000
    embedding_dim: 50
  }
}
```

This PR adds machinery in the Python frontend so that optional fields are handled appropriately:
```python
# With pad index
y = lbann.Embedding(x, num_embeddings=1000, embedding_dim=50, padding_idx=0)

# Without pad index
z = lbann.Embedding(x, num_embeddings=1000, embedding_dim=50)
```

### DIFF
--- a/python/lbann/util/class_generator.py
+++ b/python/lbann/util/class_generator.py
@@ -1,15 +1,20 @@
 """Utility functions to generate classes from Protobuf messages."""
 import google.protobuf.descriptor
+import google.protobuf.wrappers_pb2
 from lbann import lbann_pb2, callbacks_pb2, layers_pb2, metrics_pb2, model_pb2, objective_functions_pb2, optimizers_pb2, weights_pb2
+from lbann.util import make_iterable
 
-# Map from Protobuf label enums to strings
-_proto_label_to_str = {
+# Each field in a Protobuf message is labeled as 'optional',
+# 'required', or 'repeated'
+# Note: 'optional' is not used in Protobuf 3.
+_protobuf_field_label_names = {
     google.protobuf.descriptor.FieldDescriptor.LABEL_OPTIONAL: 'optional',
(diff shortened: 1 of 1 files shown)


ndryden:
We may want to subsequently adjust the way batchnorm statistics specifications are handled in the prototext, since that was a particularly ugly case of working around this.

timmoon10:
[Bamboo passes on Pascal and Corona (I wasn't able to get an allocation on Lassen)](https://lc.llnl.gov/bamboo/browse/LBANN-TIM219-2)

PR #1294, opened 2019-10-14
data store out-of-memory-mode

davidHysom:
First stab at modifying data_store_conduit to spill local conduit nodes to file, then read back into RAM as needed. (Actually, "local file" is misleading; you can spill to any mounted disk.)

This PR also partially answers the question as to what the question is, re, 42. But that's hidden in the codebase comments. Which you can't see, because ... it's hidden. Also encrypted. 

initial commit. Nothing to see here, please move along.


davidHysom:
While working on PR 1294: (data store out-of-memory-mode), which builds on this PR, I realized I need to refactor the code in this PR, so it seems reasonable to combine them in a single PR. (I also realized this PR hasn't been reviewed because it was still a draft. Dang it.)

PR #1295, opened 2019-10-15
Replace MNIST unit tests with individual layer tests

timmoon10:
A significant portion of the nightly Bamboo build is taken by MNIST unit tests, namely `test_unit_mnist_conv_graph`, `test_unit_mnist_ridge_regression`, and `test_unit_mnist_softmax_classifier`. The main purpose of these tests is to validate the convolution and fully-connected layers with gradient checking. However, they require dealing with MNIST data and don't perform any metric checking. These tests have missed a few bugs in the past.

This PR takes a faster, more thorough, and more fine-grained approach. The MNIST tests have been replaced with unit tests for the following layers:
- Fully-connected
- Convolution
- Softmax
- Cross entropy
- Mean squared error

The metric values are compared against those computed by NumPy, SciPy, or PyTorch. Everything passes on Pascal and Catalyst.


timmoon10:
[No suspicious failures in my Bamboo build.](https://lc.llnl.gov/bamboo/browse/LBANN-TIM223-3)

PR #1297, opened 2019-10-15
Mpi int max bug

davidHysom:
fixed bug from Tom's reversion


davidHysom:
opened under a different name; there seems to be a problem with reusing branch names

PR #1303, opened 2019-10-16
Utility functions for hashing

### LINKED ISSUE #1222: Move trainer's pair hash to utility function
>  This should probably move to "utils" or something. Perhaps near where execution_mode is declared/defined.

timmoon10:
This is a minor PR intended to close #1222. It adds a utility header `utils/hash.hpp` with the following functions/functors:

- `hash_combine`: Mix two hash values using a [magic one-liner from Boost](https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine).
- `enum_hash`: Hash function for enums.
- `pair_hash`: Hash function for `std::pair`s.

These all have Catch2 unit tests that check for hash collisions. This functionality is largely moved from `trainers/trainer.hpp`. I've also gone ahead and used the `hash_combine` function to replace how we mix RNG seeds.

### DIFF
--- a/include/lbann/trainers/trainer.hpp
+++ b/include/lbann/trainers/trainer.hpp
@@ -33,6 +33,7 @@
 #include "lbann/execution_contexts/execution_context.hpp"
 #include "lbann/io/persist.hpp"
 #include "lbann/utils/threads/thread_pool.hpp"
+#include "lbann/utils/hash.hpp"
 #include <lbann.pb.h>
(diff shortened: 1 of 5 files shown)


timmoon10:
I don't see any suspicious errors in the [Bamboo build](https://lc.llnl.gov/bamboo/browse/LBANN-TIM224-1).

PR #1307, opened 2019-10-18
Tinkering with MPI launch invocations

timmoon10:
Changes to the Python frontend's main launcher (`lbann.run`):

- Fixed the LSF launcher's `jsrun` invocation. You can now call `lbann.run` on Lassen and it will run fine.

Changes to the LC-specific launcher (`lbann.contrib.lc.launcher.run`):

- Set the environment variable `IBV_FORK_SAFE ` on Lassen and Sierra. This fixes a hang in the Python data reader (the Python data reader creates a pool of worker processes, but [InfiniBand and OpenMPI have been known to fail after forking](https://www.open-mpi.org/faq/?category=openfabrics#ofa-fork)).
- Generalized the Pascal-specific optimizations to work with 1 proc/node. Running AlexNet with 2 nodes and 1 proc/node, the mini-batch time drops from 0.59s to 0.11s.
- Generalized the Sierra/Lassen-specific optimizations and only use 16 cores/socket. Running AlexNet with 2 Lassen nodes and 4 procs/node, the mini-batch time drops from 0.050s to 0.044s. Running with 1 proc/node, the mini-batch time drops from 0.064s to 0.061s.

### DIFF
--- a/python/lbann/contrib/lc/launcher.py
+++ b/python/lbann/contrib/lc/launcher.py
@@ -112,11 +112,12 @@ def make_batch_script(script_file=None,
         launcher_args.extend(['--mpibind=off',
                               '--nvidia_compute_mode=default'])
 
-    # Deal with Pascal's hardware topology
+    # Optimized thread affinity for Pascal
     # Note: Both GPUs are on socket 0, so we only use cores on that
     # socket.
     if system == 'pascal':
-        cores_per_proc = (cores_per_node(system) // 2) // procs_per_node
+        cores_per_socket = cores_per_node(system) // 2
+        cores_per_proc = cores_per_socket // procs_per_node
         environment['AL_PROGRESS_RANKS_PER_NUMA_NODE'] = procs_per_node
         environment['OMP_NUM_THREADS'] = cores_per_proc - 1
         if scheduler == 'slurm':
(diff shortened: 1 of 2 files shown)


python/lbann/contrib/lc/launcher.py
```diff
@@ -103,45 +103,57 @@ def make_batch_script(script_file=None,
...
-        environment['OMP_NUM_THREADS'] = 4
-        # Deal with topology mis-identification on Sierra/Lassen.
-        environment['AL_PROGRESS_RANKS_PER_NUMA_NODE'] = 2
+            launcher_args.append('--bind packed:{}'.format(cores_per_proc))
+
+        # Hack to enable process forking
+        # Note: OpenMPI and InfiniBand sometimes fail if an MPI
+        # process is forked. See
+        # https://www.open-mpi.org/faq/?category=openfabrics#ofa-fork.
+        environment['IBV_FORK_SAFE'] = 1
```
naoyam: Looks like this has some performance penalty (per Adam's email). Can this be optional?

timmoon10: The [documentation](https://linux.die.net/man/3/ibv_fork_init) says:

> Calling ibv_fork_init() will reduce performance due to an extra system call for every memory registration, and the additional memory allocated to track memory regions. The precise performance impact depends on the workload and usually will not be significant.

This function is a convenience wrapper around the `lbann.run` function, so perhaps not using `IBV_FORK_SAFE` should be considered as expert usage?

timmoon10:
[Bamboo passes.](https://lc.llnl.gov/bamboo/browse/LBANN-TIM225-2)

python/lbann/contrib/lc/launcher.py
naoyam: As long as it's clearly communicated to the user about the potential performance loss and there is a way to turn it off, I think that's fine.

PR #1309, opened 2019-10-18
Redo of #1263

benson31:
The `OPTIONAL` tag was in the wrong place. See #1263 for the rest of the description.


benson31:
https://lc.llnl.gov/bamboo/browse/LBANN-BEN3-4

PR #1317, opened 2019-10-21
Fix RNG state storing

JaeseungYeom:
Address two race conditions in
- saving shared RNG states
- creating common directory

Also, changes the name of state dump files to use the rank-in-trainer instead of rank-in-world
The rebased version of PR #1320 added.


bvanessen:
@JaeseungYeom can you please rebase your branch.

JaeseungYeom:
I will create a new branch that is rebased. The way I will try is to reset to a commit before the divergence, then rebase, and add fresh commits. Rebasing as is fails. If there is a better way, let me know. I am not sure how this will affect the other unmerged existing PR of mine. Maybe that needs rebasing as well.

src/utils/random.cpp
```diff
@@ -99,42 +99,52 @@ fast_rng_gen& get_fast_io_generator() {
   return ::fast_io_generator;
 }
 
-bool save_rng_to_checkpoint(persist& p, const lbann_comm* comm) {
+bool save_rng_to_checkpoint(persist& p, lbann_comm* comm) {
```
benson31: Why is `comm` no longer pointer-to-`const`? Why does saving to checkpoint modify `comm`?

src/utils/random.cpp
```diff
@@ -99,42 +99,52 @@ fast_rng_gen& get_fast_io_generator() {
...
+    std::ofstream rng_EL(rng_name);
+    if(!rng_EL) { LBANN_ERROR("Failed to open ", rng_name); }
+    rng_EL << El::Generator();
+    rng_EL.close();
 #endif
+  }
 
-  std::string rank_in_world;
+  std::string rank_in_trainer;
   if (comm == nullptr) {
```
benson31: This code should move into the `if (comm == nullptr)` block above (line 104).

src/utils/random.cpp
JaeseungYeom: comm object internally maintains a counter of each call. I am using it here for barrier to make sure the creation of the directory is done only once by the trainer master and visible to everyone else.

benson31: 🤦‍♂.

Counters like this should be marked `mutable`... There is no logical universe in which calling collectives should alter the state of a communicator. Anyway, this isn't your fault (presumably). I'll open an issue.

JaeseungYeom:
> If you don't need to save the history of all commits in this PR, squashing them into one commit (or a small number of commits) would probably make rebasing and fixing it much easier.

Thanks for the suggestion. I will give this a try.

src/utils/random.cpp
JaeseungYeom: I see that it can save one branching.

JaeseungYeom:
rebased version got merged. closing this PR

PR #1318, opened 2019-10-21
Bugfix for softmax and log-softmax layers

timmoon10:
I've noticed some errors when I manually run the softmax and log-softmax unit tests on Lassen (although not when I run with Bamboo). Upon closer inspection, the problem is in the model-parallel GPU implementation. There is one kernel (for column-wise max reduction) that we launch even when there is no local data, causing a kernel launch error. In the process of debugging, I've made several other changes:
- Modified the layer unit tests so that the last mini-batch has size 1. This should help catch this kind of error in the future.
- Adjusted the way we do thresholding in the softmax layer. We no longer zero out the error signal if an entry is thresholded in forward prop. I don't expect this to have much of an effect, but it might help avoid getting stuck when input entries have a wide range.
- Added a utility function for GPU block-wide reductions.
- Implemented a unit test for the log-softmax layer with the Python frontend (progress toward #1245).
- Expanded documentation, especially for CUDA kernels.


timmoon10:
[This does not introduce any new Bamboo errors.](https://lc.llnl.gov/bamboo/browse/LBANN-TIM226-1)

PR #1321, opened 2019-10-22
Refactor Bamboo helper function for Python frontend

timmoon10:
One of the problems with PR #1291 is that we have a lot of boilerplate to setup and run LBANN with the Python frontend. This PR generalizes the `create_tests` function in the Bamboo utilities so that it is not specifically tied to the `unit_test` directory. It also provides options to run with more than one node.

My plan is to implement the integration tests by calling `create_tests` and writing wrapper functions that parse the log files after LBANN finishes (e.g. to check mini-batch times or accuracies). One complication is that the `_exe` tests have a different signature from the `_gcc7`, `_clang6`, and `_intel19` tests.  I've removed the `_exe` tests for simplicity. This becomes moot if we run the tests in the build directory since we no longer need to maintain different versions of the test for each build configuration.


timmoon10:
[Bamboo is green.](https://lc.llnl.gov/bamboo/browse/LBANN-TIM227-1)

PR #1322, opened 2019-10-23
Bamboo integration tests using Python frontend (attempt 2)

timmoon10:
This PR is another attempt at reworking our Bamboo integration tests (see #1291). It replaces all of the existing integration tests with the following tests:
- LeNet; 5 epochs of MNIST
- AlexNet; 5 epochs of 360K ImageNet samples
- ResNet-50; 5 epochs of 360K ImageNet samples

The LeNet test will run nightly on all systems and the AlexNet and ResNet-50 tests will run weekly on GPU systems. The LeNet test takes ~30 sec on Catalyst and the ResNet-50 test takes ~30 min on Pascal, so we have plenty of our time budget remaining.

I've attempted to reduce boilerplate by using the same utility functions as the tests in the `unit_test` directory. It's a bit messy since we need to create wrapper functions around the test functions so that we can parse the log files. This is ready for review, although I still need to get expected runtimes on Ray and Corona.

Closes #1291.


bamboo/integration_tests/test_integration_lenet.py
```diff
@@ -0,0 +1,204 @@
...
+    # No validation set
+    message.reader[0].validation_percent = 0
+
+    return message
+
+# ==============================================
+# Setup PyTest
+# ==============================================
+
+def augment_test_func(test_func):
```
forsyth2: This is identical to the alexnet `augment_test_func` except for not skipping Nightly/CPU builds (and also not including `top-5` on the accuracy). Seems like it would be better to pull this out into a common function. The late-binding closures wouldn't be affected by that, I believe.

bamboo/integration_tests/test_integration_resnet50.py
```diff
@@ -0,0 +1,223 @@
...
+    # Only evaluate on ImageNet validation set at end of training
+    message.reader[1].role = 'test'
+
+    return message
+
+# ==============================================
+# Setup PyTest
+# ==============================================
+
+def augment_test_func(test_func):
```
forsyth2: Same comment about this being duplicated code.

bamboo/integration_tests/test_integration_lenet.py
timmoon10: At a certain point, I don't think it's worthwhile reducing code duplication if it makes the code harder to understand or if it constrains us to a rigid framework. I was thinking about an alternate approach where we pass callbacks into the `create_tests` function to perform the various assertions. It's fairly general and it allows us to consolidate some code, but I suspect it's hard to understand.

forsyth2: I think reducing code duplication makes the code easier to understand - but we could argue this _ad infinitum_. I would have to see the callback approach to see how clean that is, but that seems like another good option.

In any case, if the code runs and the tests pass, then the PR is good to merge. Perhaps LBANN should at some point consider discussing code style guidelines (including things like "code duplication v factoring out common code").

PR #1323, opened 2019-10-23
Fix saving of the common RNG states for distributed checkpointing

JaeseungYeom:
Common RNG states still need to be saved by each rank for distributed checkpointing.


benson31:
@bvanessen Yes, there is a passing build.

PR #1331, opened 2019-10-31
Updating the build script

naoyam:
The build script is officially obsolete, but Distconv still relies on that.

This PR updates the CUDA, cuDNN and NCCL versions. CUDA 10.1 is the new default version with cuDNN 7.6.4 and NCCL 2.4.8. Build works on Lassen and Pascal. It also updates the default version of CMake.


naoyam:
That's true. I thought it's been addressed. I changed it back to 2.4.2.

PR #1332, opened 2019-10-31
Documentation updates

benson31:
Doxygen now installs with Sphinx for deployment on RTD.

LBANN vocabulary description is now included in the RTD.

Fixed a typo.


davidHysom:
> Doxygen now installs with Sphinx for deployment on RTD.

We should never use acronyms unless they are defined (at a minimum) the first time they appear

PR #1336, opened 2019-11-04
Initialize RNGs for trainers differently by default

JaeseungYeom:
Basically change the interface ```random_init_models_differently: true``` to ```random_init_trainers_identically: false```

Passed baboo tests https://lc.llnl.gov/bamboo/browse/LBANN-JAES35

### DIFF
--- a/src/proto/trainer.proto
+++ b/src/proto/trainer.proto
@@ -49,15 +49,13 @@ message Trainer {
   // background.
   int64 num_parallel_readers = 3;
 
-
-  // If false, trainers will have their trainer rank mixed into their random seed.
-  bool random_init_trainers_identically = 4;
-
   // -------------------------------
   // Advanced options
   // -------------------------------
 
+  // If false, trainers will have their trainer rank mixed into their random seed.
+  bool random_init_trainers_identically = 4;
+
   // Algorithmic block size for Hydrogen
   int64 hydrogen_block_size = 100;
-
 }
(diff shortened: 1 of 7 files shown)


src/proto/trainer.proto
```diff
@@ -49,6 +49,10 @@ message Trainer {
   // background.
   int64 num_parallel_readers = 3;
 
+
+  // If false, trainers will have their trainer rank mixed into their random seed.
+  bool random_init_trainers_identically = 4;
```
timmoon10: Can we put this under the "Advanced options"?

src/utils/lbann_library.cpp
```diff
@@ -182,16 +182,16 @@ std::unique_ptr<model> build_model_from_prototext(
 
   // Initialize models differently if needed.
 #ifndef LBANN_DETERMINISTIC
-  if (pb_model->random_init_models_differently()) {
+  if (!pb_trainer->random_init_trainers_identically()) {
     random_seed = random_seed + comm->get_trainer_rank();
```
timmoon10: Perhaps beyond the scope of this PR, but we have a [utility function to mix RNG seeds](https://github.com/LLNL/lbann/blob/279b4d4a5f0268519c9ce91f5940a59a355d1336/include/lbann/utils/hash.hpp#L47):
```suggestion
     random_seed = hash_combine(random_seed, comm->get_trainer_rank());
```

src/proto/trainer.proto
JaeseungYeom: sure

src/utils/lbann_library.cpp
JaeseungYeom: I will apply this.

JaeseungYeom:
Should I merge?

PR #1337, opened 2019-11-04
Add hacked bugfixes for Lassen

timmoon10:
These are intended to fix errors that show up from the latest Lassen update.

### DIFF
--- a/python/lbann/contrib/lc/launcher.py
+++ b/python/lbann/contrib/lc/launcher.py
@@ -165,8 +165,7 @@ def make_batch_script(script_file=None,
             environment['IBV_FORK_SAFE'] = 1
 
         # Hacked bugfix for hcoll (1/23/19)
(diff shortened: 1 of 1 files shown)


python/lbann/contrib/lc/launcher.py
```diff
@@ -164,6 +164,18 @@ def make_batch_script(script_file=None,
         if 'IBV_FORK_SAFE' not in environment:
             environment['IBV_FORK_SAFE'] = 1
 
+        # Hacked bugfix for hcoll (1/23/19)
+        # Note: Fixes some infrequent hangs in MPI_Allreduce and
```
ndryden: We get hangs in `MPI_Bcast`, not allreduce, as far as I know.

ndryden: (Also, the hangs happen nearly deterministically, not infrequently.)

PR #1339, opened 2019-11-05
Identify and fix bugs in data_reader_jag_conduit

davidHysom:
read_partial_node() was segfaulting due to being called before the data_store was instantiated.

### DIFF
--- a/src/data_readers/data_reader_jag_conduit.cpp
+++ b/src/data_readers/data_reader_jag_conduit.cpp
@@ -280,6 +280,8 @@ bool data_reader_jag_conduit::has_path(const hid_t& h, const std::string& path) 
 }
 
 void data_reader_jag_conduit::read_node(const hid_t& h, const std::string& path, conduit::Node& n) const {
+//XX
+if (is_master()) std::cerr << "read_node for path: " << path << std::endl;
   conduit::relay::io::hdf5_read(h, path, n);
 }
 #endif
@@ -289,6 +291,9 @@ const conduit::Node& data_reader_jag_conduit::get_conduit_node(const conduit::No
 }
 
 bool data_reader_jag_conduit::load_conduit_node(const size_t i, const std::string& key, conduit::Node& node) const {
+
+//XX
+if (is_master()) std::cerr << "load_conduit_node for key: " << key << std::endl;
 
   if (m_io_thread_pool != nullptr && m_using_random_node.count(m_io_thread_pool->get_local_thread_id())) {
     LBANN_ERROR("previously retrieved a random conduit node from data_store, so shouldn't be here");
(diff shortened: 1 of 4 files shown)


src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -925,10 +862,18 @@ if (is_master()) std::cout << "SETTING m_is_data_loaded = true\n";
   select_subset_of_data();
 }
 
+void data_reader_jag_conduit::preload_helper(const hid_t& h, const std::string &sample_name, const std::string &field_name, int data_id, conduit::Node &node) {
+  const std::string path = sample_name + field_name;
+  if (!has_path(h, path)) {
+    LBANN_ERROR("!has_path(h, path) for path: ", path);
+  }
+  conduit::Node work;
```
bvanessen: This is going to copy the value from the conduit node I think that you want something like:
` const std::string key2 = '/' + LBANN_DATA_ID_STR(data_id) + field_name;
   conduit::Node& work = node[key2];`

benson31: Or even `const std::string key2 = build_string('/', LBANN_DATA_ID_STR(data_id), field_name);` 😃

davidHysom: That block looks for the path in the conduit node on disk; 'h' is hid_t.  The node has not been read into memory at that point. We could eliminate those three lines, because if the path doesn't exist conduit will throw an exception. That would avoid the cost of the check.

davidHysom: > build_string

Looks like that instantiates extra objects: ostringstream, then returns a copy of the ss.str(). Of course, I don't know how smart the compiler might be about optimizing that.

benson31: If this is not a performance-critical section, it won't matter, and the code will be more understandable. My guess is that the cost of creating an ostringstream, piping things into it, and copying the string back out will be significantly faster than the `read_node` call... That said, this is not a compelling use-case for `build_string` since, seemingly, everything is either a string or convertible to it. (`build_string` really exists to simplify `std::string s = std::to_string(some_var) + " " + std::to_string(other_var) ... ;` use-cases in non-performance-critical code sections.)

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -523,6 +485,10 @@ bool data_reader_jag_conduit::filter(const std::set<std::string>& key_filter,
   return false;
 }
 
+/**
```
benson31:
```suggestion
/*
```

src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -915,12 +888,19 @@ void data_reader_jag_conduit::do_preload_data_store() {
...
       m_sample_list.open_samples_file_handle(index, true);
-      load_conduit_node(index, key, work);
+      auto h = m_sample_list.get_samples_file_handle(id);
       conduit::Node & node = m_data_store->get_empty_node(index);
-      const std::string padded_idx = '/' + LBANN_DATA_ID_STR(index);
-      node[padded_idx] = work;
+
+      preload_helper(h, sample_name, m_output_scalar_prefix, index, node);
+      preload_helper(h, sample_name, m_input_prefix, index, node);
+      for (auto t : m_emi_image_keys) {
```
benson31:
```suggestion
      for (const auto& t : m_emi_image_keys) {
```

benson31:
If you ask for multiple reviews, it's probably best to wait for multiple reviews. Also, reviewers don't get notified when you change "in progress" to "review requested", so it's likely to take a while for reviewers to notice (another reason to prefer draft PRs).

PR #1341, opened 2019-11-05
Move gradient scaling from learning layers to evaluation layer

timmoon10:
### Math

In forward prop, the evaluation layer computes a loss function across a mini-batch:
```
L = 1/N * sum( L_i )
```
In backward prop, learning layers compute loss gradients w.r.t. the weights:
```
dL/dw = 1/N * sum( d(L_i)/dw )
```
However, derivatives are linear and we could rearrange this:
```
dL/dw = sum( d(L_i/N)/dw )
```

### Proposal
In other words, we can scale the weights gradients in the evaluation layer rather than in the learning layers. This is nicer for several reasons:
- The eval layer's backprop matches its forward prop.
- The learning layers don't need to know about mini-batch sizes. In fact, they could have dimensions different from the mini-batch size.
- This approach can handle the case where the loss function is not a mean over a mini-batch.
- Since learning layers aren't tied to mini-batches, they no longer depend on `sgd_execution_context`s.

### Evaluation

~[Bamboo is mostly green](https://lc.llnl.gov/bamboo/browse/LBANN-TIM231-1), although Corona is hanging like the nightly build. Integration tests on Lassen all pass.~ [Bamboo is green](https://lc.llnl.gov/bamboo/browse/LBANN-TIM231-3).


include/lbann/weights/weights.hpp
```diff
@@ -33,6 +33,7 @@
 #include "lbann/optimizers/optimizer.hpp"
 #include "lbann/io/persist.hpp"
 #include "lbann/utils/description.hpp"
+#include "lbann/utils/memory.hpp"
```
ndryden: Is this needed here?

timmoon10: Nope.

timmoon10:
@ndryden The AlexNet and Resnet-50 integration tests are passing when I run on Lassen, so it doesn't seem to have a major effect on learning or performance.

ndryden:
@timmoon10 Right, but I want to confirm with both full-scale runs and fine-grained profiling.

include/lbann/weights/weights.hpp
benson31: `std::unique_ptr` is used below -- you should `#include` something.

timmoon10:
Performance results on 1 Lassen GPU with synthetic data and mini-batch 32:
- eb27044 (#1341): 0.10616 sec/mini-batch
- d2060b1 (develop): 0.10664 sec/mini-batch

ndryden:
:shipit:

include/lbann/weights/weights.hpp
timmoon10: We include `<memory>`, so I think we're good. As far as I can tell, `lbann/utils/memory.hpp` only include `make_unique`, which we don't use in this file or the associated source file.

In any case, things compile for me.

benson31: Oh, yeah, the main "conversation" snippet didn't show that line and I missed it when I skimmed the file. 👍

PR #1342, opened 2019-11-06
Fix bug in model zoo CMake

timmoon10:
PR #1335 moved `model_zoo/vision` to `applications/vision`, but forgot to update the CMake. A Bamboo build is in progress.

As an aside, it's odd that we copy the entire model zoo to the install directory.


bvanessen:
We copy in the model zoo to the install directory so that when we install LBANN, people can use the model zoo without the source code.

PR #1343, opened 2019-11-08
Add support for MPI Catch2 testing

benson31:
This is preliminary support for testing MPI code in Catch2.

In terms of Catch2, there are two blockers to using it in an MPI environment:

1. Support for parallel output.
2. Correctly handling MPI initialization/finalization.

Solving 1 is a trivial problem. This PR does nothing to change stdout/stderr. Thus, running `mpirun -np 1000 ./my_test_main`, 1000 copies of stdout/stderr will be dumped, with no ordering or synchronization. If your MPI launcher supports redirecting stdout/stderr, this can be used as usual (which is why I'm hesitant to change the behavior of stdout/stderr too much). This PR introduces a (familiar-looking) escape sequence syntax for naming files, similar to LSF or nvprof or such things. This is documented in `unit_test/utilities/ReplaceEscapes.hpp`. It enables things like: `mpirun -np 1000 ./my_test_main -o "my_output_on_rank_%r.log"`.

Solving 2 is also pretty straightforward, but, by its solution, it raises some other questions. MPI cannot be stood up in a `TEST_CASE`, because this would imply only 1 test case can exist per executable (behavior of an executable after `MPI_Finalize()` is undefined, and re-calling `MPI_Init()` is not allowed). So we need a custom `main()` function that stands up MPI before running the tests and tears it down once all tests are done. This is easy enough.

Now come the hard questions that MPI poses for testing (that would be at least somewhat ameliorated by mocking MPI, but that's a much larger undertaking that's lacking various resources).

1. What if an MPI call fails during a test? How do we tear down testing in a way where we get meaningful results about the test and about the failure?
2. Can we test cases in which MPI is expected to fail? If we switch to a mode (for testing only) in which we have MPI return error codes, is the job still in a valid state with respect to MPI after that? (This is decidedly side-stepped by the MPI spec.)

As yet, these are not addressed by this PR. I think the answer to (1) will be most easily achieved by having MPI return error codes, which trigger exceptions in our stack that will always be treated as fatal. Upon detection, remaining tests will be ignored and output up to failure will be generated prior to shutdown. This continues to side-step (2), but (2) is a much harder problem.

The remaining TODO is to implement a few test cases demonstrating the functionality. If the reviewers think this is a sufficient body of work, I'm happy to defer this to an adjacent, dependent PR.


benson31:
I have added an example that just does some broadcasts. The intent is not to be a "real" test but rather to demonstrate communicator access and provide a template for future tests. Once built, the MPI tests must be run manually (CMake integration presents a wide array of challenges that we probably shouldn't undertake). An example run command looks like:
```bash
mpirun -np 4 ${BUILD_DIR}/unit_test/mpi_catch_tests -o mpi_output_%r_%s.log
```

PR #1347, opened 2019-11-12
Develop new data reader: ras_lipid_conduit_data_reader

davidHysom:
Initial commit.


davidHysom:
> (1) Is there a way to merge this (or extend) existing NPZ data reader to avoid proliferation of data reader?
The two readers differ enough that they can't be combined into a single data reader. At best there could be base data reader with and two derived data readers.  Some differences:
 a)  the new data reader has some private variables the are not in the existing NPZ reader, so we can't have a common copy_members() function
 b) what gets stuffed into a conduit node differs
 c) what gets copied into matrices in fetch_datum and fetch_label differs
 d) because the ras-lipid dataset contains multiple samples per file, the manner in which the samples are divided amongst the the processors differ (this will change if/when we get the sample_list class working for the ras-lipid data)

> (2) We might want to consider a change of name, I have no better suggestion, but may be candle_pilot2
I admit I'm terrible about naming things. I've no objection to a name change.

include/lbann/data_store/data_store_conduit.hpp
```diff
@@ -341,7 +352,7 @@ private :
   // END: timers for profiling exchange_data
   //===========================================================
 
-  int m_cur_epoch = 0;
+  int m_cur_epoch = -1;
```
bvanessen: Does changing this have any effect on any other data store.

davidHysom: No; it's only used internally, e.g, for deciding when to print timing statistics. It had been printing timing statistics before exchange_data was called, so I set it to -1 to avoid that.

davidHysom: I've removed the m_cur_epoch variable, which was used to determine when to exchange owner maps. I think the revised logic is better.

samadejacobs:
I am still a reviewer on this? I see no button to approve. In any case, there have been back and forth between me and Dave on this, IMHO, the PR is now in a good shape merge, I approve.

bvanessen:
Great.  You are still a reviewer on it.

Brian C. Van Essen
vanessen1@llnl.gov
(w) 925-422-9300
(c) 925-290-5470

PR #1350, opened 2019-11-13
Draft implementation of RAS state classifier model

samadejacobs:
Draft implementation of CNN based classifier of RAS state in pilot2 MD simulation of RAS initiation pathway for cancer. First LBANN implementation, a work in progress. Note, python data reader in this application will be replaced  with npz reader ([PR1347](https://github.com/LLNL/lbann/pull/1347))


adammoody:
May need a transpose to move the channel data to come in a different order in memory:
https://www.programcreek.com/python/example/4564/numpy.transpose

samadejacobs:
> May need a transpose to move the channel data to come in a different order in memory:
> https://www.programcreek.com/python/example/4564/numpy.transpose

Good catch, done.

PR #1351, opened 2019-11-15
updated queue names for corona

graham63:


benson31:
It looks like the corona queue names have changed. This is aimed at fixing it. I haven't looked or tested yet. @graham63 is getting access to the machine.

PR #1352, opened 2019-11-16
Layer norm

timmoon10:
[Layer norm](https://arxiv.org/pdf/1607.06450.pdf) is similar to batchnorm, except statistics are computed over each sample rather than over the mini-batch. It's preferable to batchnorm in RNNs and is used within attention modules.


timmoon10:
[Bamboo is mostly green](https://lc.llnl.gov/bamboo/browse/LBANN-TIM236-1). Bamboo doesn't seem to be running for me on Corona.

PR #1354, opened 2019-11-19
Refactor layer modules in Python frontend

timmoon10:
This PR reorganizes the layer modules within a directory. This will help prevent all the modules being in one monstrously large `modules.py` file. My next step is to implement a multi-head attention module within this directory structure.

The external API hasn't changed, the vision and NLP application models run without incident, and [Bamboo is green](https://lc.llnl.gov/bamboo/browse/LBANN-TIM238-1).


python/lbann/modules/base.py
```diff
@@ -0,0 +1,233 @@
...
+    def forward(self, x):
+        self.instance += 1
+        name = '{0}_instance{1}'.format(self.name, self.instance)
+        y = lbann.FullyConnected(x,
+                                 weights=self.weights,
+                                 name=(name+'_fc' if self.activation else name),
+                                 data_layout=self.data_layout,
+                                 num_neurons=self.size,
+                                 has_bias=self.bias)
+        if self.activation:
```
samadejacobs: How do you deal with activations with additional parameters e.g., LeakyRelu has slope?

python/lbann/modules/base.py
```diff
@@ -0,0 +1,233 @@
...
+                              num_dims=self.num_dims,
+                              num_output_channels=self.out_channels,
+                              has_vectors=False,
+                              conv_dims_i=self.kernel_size,
+                              conv_pads_i=self.padding,
+                              conv_strides_i=self.stride,
+                              conv_dilations_i=self.dilation,
+                              num_groups=self.groups,
+                              has_bias=self.bias)
+        if self.activation:
```
samadejacobs: Same question as before

python/lbann/modules/base.py
timmoon10: I'm leaning against supporting this. [`torch.nn.modules.Linear`](https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html) doesn't support activations; [`tf.keras.layers.Dense`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dense) and [`keras.layers.Dense`](https://keras.io/layers/core/) don't allow you to pass arguments to the activation. We might be able to make this usable for some cases by supporting non-zero defaults for leaky reLU and ELU.

PR #1355, opened 2019-11-19
Refactor core components in Python frontend

timmoon10:
This PR reorganizes LBANN's core components (layers, models, trainers, etc.) within a directory, mostly to keep the package structure neat. It was originally some miscellaneous cleaning I was doing for #1354, but I figure it's better as a standalone PR.

The external API hasn't changed and [Bamboo is green](https://lc.llnl.gov/bamboo/browse/LBANN-TIM237-1).


timmoon10:
The `Trainer` class already supports `procs_per_trainer`, but we need to add support for `init_model_differently`. I think the best way is to autogenerate the `Trainer` class from the protobuf so any new options are automatically picked up. That's probably best for another PR though.

PR #1357, opened 2019-11-27
Fix bugs in Python frontend

ndryden:
Import was not updated for reorg

Broken command line parameters


applications/vision/resnet.py
```diff
@@ -51,9 +51,6 @@
     '--random-seed', action='store', default=0, type=int,
     help='random seed for LBANN RNGs', metavar='NUM')
 lbann.contrib.args.add_optimizer_arguments(parser, default_learning_rate=0.1)
-parser.add_argument(
-    '--setup_only', action='store_true',
```
timmoon10: If we want to keep this command-line argument, we can change it to:
```suggestion
    '--setup-only', action='store_true',
```

ndryden: It's already added as a command-line argument elsewhere.

PR #1360, opened 2019-12-03
Add support for Deconv (transpose conv) in python frontend

samadejacobs:


python/lbann/models/resnet.py
```diff
@@ -14,7 +14,7 @@ class ConvBNRelu(lbann.modules.Module):
 
     def __init__(self, out_channels, kernel_size, stride, padding,
                  bn_zero_init, bn_statistics_group_size,
-                 relu, name):
+                 relu, name, transpose=False):
```
timmoon10: Does ResNet work with deconvolution layers? Won't the outputs become huge?

ndryden:
I do not think `ConvBNRelu` should support transposition. It should be a separate operator.

python/lbann/models/resnet.py
samadejacobs: It is not really about ResNet but reusing the Conv-BN-activation pattern. In any case, I will take it out since my immediate need require tinkering with the activation part.

ndryden:
@samadejacobs Why was this merged without any approvals?

samadejacobs:
There was an approval by human and Slack interaction.

PR #1363, opened 2019-12-06
Support transposes in matrix multiply layer

timmoon10:
This PR adds support for transposing the inputs to the matrix multiply layer, similar to [`tf.linalg.matmul`](https://www.tensorflow.org/api_docs/python/tf/linalg/matmul). I plan to use this in multi-head attention. [Nothing suspicious so far in Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM243-1).

### DIFF
--- a/include/lbann/layers/math/matmul.hpp
+++ b/include/lbann/layers/math/matmul.hpp
@@ -50,19 +50,31 @@ class matmul_layer : public data_type_layer<TensorDataType> {
 
 public:
 
-  matmul_layer(lbann_comm *comm);
+  matmul_layer(lbann_comm *comm,
+               bool transpose_a = false,
+               bool transpose_b = false);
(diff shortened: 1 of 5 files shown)


include/lbann/layers/math/matmul.hpp
```diff
@@ -50,29 +50,43 @@ class matmul_layer : public data_type_layer<TensorDataType> {
...
+
 protected:
 
   void setup_dims() override;
   void fp_compute() override;
   void bp_compute() override;
 
+private:
+
+  /** Whether to transpose matrices from the first input tensor. */
```
samadejacobs: Prefer TF-like documentation (e.g., _transpose_a: If True, a is transposed before multiplication_), this is clear enough though

PR #1365, opened 2019-12-09
Pilot2 conduit data reader

davidHysom:
Reader for pilot2 data that was previously converted from npz to conduit format (either hdf5 or conduit-bin). This reader uses the sample_list class.


davidHysom:
No longer applicable: due to execution speed, we will not be converting npz data to conduit format on disk; will read npz files directly into memory, than convert on-the-fly to conduit for the data_store. Dave H. will modify the cnpy package to enable its use with the sample_list class.

PR #1367, opened 2019-12-10
Feature add ETI for multiple data type for learning layers

bvanessen:
Adding support for the convolutional layers to be explicitly defined for both float and double.


bvanessen:
This PR has passed bamboo.

PR #1368, opened 2019-12-10
Allow Layer classes to return a BaseDistMatrix for activations and error signals

bvanessen:
Updated the get_activations and get_error_signal functions that take a
parent or child layer as an argument have been promoted back to the
top layer abstract class.  These two functions now return a Hydrogen
BaseDistMatrix type, rather than a data type specific distributed
matrix.  Note that these updates require the use of Hydrogen v1.3.1 to
allow for the copying between BaseDistMatrix types and other abstract
distributed matrices.  Additionally, it limits the scope of the
accessor functions that return both the previous activations or error
signals.  This does require the allocation of some friend classes.

### DIFF
--- a/include/lbann/layers/data_type_layer.hpp
+++ b/include/lbann/layers/data_type_layer.hpp
@@ -310,10 +310,12 @@ private:
    */
   std::vector<std::unique_ptr<AbsDistMatrixType>> m_gradient_wrt_inputs;
 
+#ifdef LBANN_HAS_GPU
   template <typename U>
   friend class cudnn::data_parallel_layer_tensor_manager;
   template <typename U>
   friend class cudnn::entrywise_layer_tensor_manager;
+#endif // LBANN_HAS_GPU
 };
 
 #ifndef LBANN_DATA_TYPE_LAYER_INSTANTIATE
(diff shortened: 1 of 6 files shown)


include/lbann/layers/data_type_layer.hpp
```diff
@@ -297,6 +310,10 @@ class data_type_layer : public Layer {
    */
   std::vector<std::unique_ptr<AbsDistMatrixType>> m_gradient_wrt_inputs;
 
+  template <typename U>
+  friend class cudnn::data_parallel_layer_tensor_manager;
+  template <typename U>
+  friend class cudnn::entrywise_layer_tensor_manager;
```
benson31: Technically, these are only needed when `LBANN_HAS_CUDA`.

benson31:
@timmoon10 They don't need to move. They can be exposed at the `Layer` level and "covariantly returned" at the `data_type_layer` level. That could be done now with a 4-line diff.

bvanessen:
@timmoon10 I don't think that we really want or need them exposed higher at the moment.  Perhaps for access from the callbacks, but we will see when we get there.

bvanessen:
Bamboo is green.

PR #1369, opened 2019-12-11
Adding ETI with support for doubles for the math layers. 

bvanessen:
Templated the interface into cuBLAS.


bvanessen:
@timmoon10 @benson31 This is ready for review and is being checked in bamboo right now.  Please look at how I did the binary, unary, and entrywise layers.  Due to the large number of macros already there, I didn't use the PROTO pattern.

bvanessen:
Bamboo is green.

PR #1370, opened 2019-12-11
Updating the layer factories to provide ETI for float and double.

bvanessen:


bvanessen:
This passed bamboo, but now it needs updated versions of the Hydrogen.  @benson31 Please update the default installations.  Also, as committed, the PR does not explicitly set any layers to be double, so in that respect this does not change any functionality.

bvanessen:
Bamboo is green

include/lbann/layers/data_type_layer.hpp
```diff
@@ -85,6 +85,12 @@ class data_type_layer : public Layer {
   data_type_layer& operator=(const data_type_layer<TensorDataType>& other);
   virtual ~data_type_layer() = default;
 
+  /** Get a string representing the layer datatype
+   */
+  std::string get_datatype() const override {
```
benson31: Can we rename this? `get_datatype_name()` maybe?

include/lbann/layers/layer.hpp
```diff
@@ -35,9 +35,16 @@
...
 #include "lbann/utils/description.hpp"
 #include "lbann/io/persist.hpp"
+#include "lbann/macros/misc.hpp"
 #include <string>
 #include <vector>
 
+/** @brief A utility macro for easily defining default-constructed sub-class
+ *  builders.*/
+#define LBANN_DEFINE_LAYER_BUILDER(Class, FunctionName)   \
+  extern std::unique_ptr<Layer> FunctionName(lbann_comm*, \
```
benson31:
```suggestion
  std::unique_ptr<Layer> FunctionName(lbann_comm*, \
```

include/lbann/layers/layer.hpp
```diff
@@ -101,6 +108,11 @@ class Layer {
    *  human-readable, name.
    */
   inline void set_name(const std::string name) { m_name = name; }
+  /** Get a string representing the layer datatype
+   */
+  virtual std::string get_datatype() const {
```
benson31: Same comment as above -- `get_datatype_name()` perhaps?

include/lbann/macros/misc.hpp
```diff
@@ -0,0 +1,45 @@
...
+// permissions and limitations under the license.
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef LBANN_MACROS_MISC_HPP_INCLUDED
+#define LBANN_MACROS_MISC_HPP_INCLUDED
+
+namespace lbann {
+
+template <typename T>
+std::string TypeName();
```
benson31: This is not really a macro. This should probably move to its own file in `utils/`.

src/layers/data_type_layer.cpp
```diff
@@ -286,10 +286,13 @@ auto data_type_layer<TensorDataType>::get_local_error_signals(int parent_index)
...
 template <typename TensorDataType>
 auto data_type_layer<TensorDataType>::get_activations(const Layer& child) const -> const BaseDistMat& {
-  const int child_index = (std::find(m_child_layers.begin(),
-                                     m_child_layers.end(),
-                                     &child)
-                           - m_child_layers.begin());
+  if(m_child_layers.empty()) {
+    LBANN_ERROR("This layer has no children");
+  }
+  const int child_index = std::distance(m_child_layers.begin(),
```
benson31: This should change to a call to `Layer::find_layer_index`, which should be renamed to `find_child_layer_index` and it should return a `size_t` (technically the `difference_type` for whatever iterators are going to be used).

src/layers/data_type_layer.cpp
```diff
@@ -302,10 +305,10 @@ auto data_type_layer<TensorDataType>::get_activations(const Layer& child) const
 }
 template <typename TensorDataType>
 auto data_type_layer<TensorDataType>::get_error_signals(const Layer& parent) const -> const BaseDistMat& {
-  const int parent_index = (std::find(m_parent_layers.begin(),
-                                      m_parent_layers.end(),
-                                      &parent)
-                           - m_parent_layers.begin());
+  const int parent_index = std::distance(m_parent_layers.begin(),
```
benson31: If we have a method for computing the child's index, we should have the same for the parent's index.

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+  }
+  return datatype_str;
+}
+
+data_layout get_layer_data_layout_from_pbuf(const lbann_data::Layer& proto_layer) {
+  // Get parameters from prototext
+  const auto& layout_str = proto_layer.data_layout();
+  data_layout layout = data_layout::invalid;
+  if (layout_str.empty())             { layout = data_layout::DATA_PARALLEL; }
+  if (layout_str == "data_parallel")  { layout = data_layout::DATA_PARALLEL; }
```
benson31:
```suggestion
  if (layout_str == "data_parallel")  { return data_layout::DATA_PARALLEL; }
```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+    datatype_str = TypeName<DataType>();
+  }
+  return datatype_str;
+}
+
+data_layout get_layer_data_layout_from_pbuf(const lbann_data::Layer& proto_layer) {
+  // Get parameters from prototext
+  const auto& layout_str = proto_layer.data_layout();
+  data_layout layout = data_layout::invalid;
+  if (layout_str.empty())             { layout = data_layout::DATA_PARALLEL; }
```
benson31:
```suggestion
  if (layout_str.empty())             { return data_layout::DATA_PARALLEL; }
```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+  if(datatype_str.empty()) {
+    datatype_str = TypeName<DataType>();
+  }
+  return datatype_str;
+}
+
+data_layout get_layer_data_layout_from_pbuf(const lbann_data::Layer& proto_layer) {
+  // Get parameters from prototext
+  const auto& layout_str = proto_layer.data_layout();
+  data_layout layout = data_layout::invalid;
```
benson31:
```suggestion

```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+  if (layout_str.empty())             { layout = data_layout::DATA_PARALLEL; }
+  if (layout_str == "data_parallel")  { layout = data_layout::DATA_PARALLEL; }
+  if (layout_str == "model_parallel") { layout = data_layout::MODEL_PARALLEL; }
```
benson31:
```suggestion
  if (layout_str == "model_parallel") { return data_layout::MODEL_PARALLEL; }
```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+  return layout;
```
benson31:
```suggestion
  return data_layout::invalid;
```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+
+El::Device get_layer_device_from_pbuf(const lbann_data::Layer& proto_layer,
+                                      bool GPUs_disabled) {
+  // Get parameters from prototext
+  El::Device device = El::Device::CPU;
+#ifdef LBANN_HAS_GPU
+  const auto& device_str = proto_layer.device_allocation();
+  if (!GPUs_disabled) {
+    if (device_str == "gpu" || device_str.empty()) {
+      device = El::Device::GPU;
```
benson31:
```suggestion
      return El::Device::GPU;
```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+    }
+    if (device_str == "cpu") { device = El::Device::CPU; }
+    if (proto_layer.has_input()) {
+      // Input layers must be on CPU
+      device = El::Device::CPU;
```
benson31:
```suggestion
      return El::Device::CPU;
```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+    }
+  }
+#endif // LBANN_HAS_GPU
+  return device;
```
benson31:
```suggestion
  return El::Device::CPU;
```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+                                      bool GPUs_disabled) {
+  // Get parameters from prototext
+  El::Device device = El::Device::CPU;
+#ifdef LBANN_HAS_GPU
+  const auto& device_str = proto_layer.device_allocation();
+  if (!GPUs_disabled) {
+    if (device_str == "gpu" || device_str.empty()) {
+      device = El::Device::GPU;
+    }
+    if (device_str == "cpu") { device = El::Device::CPU; }
```
benson31:
```suggestion
    if (device_str == "cpu") { return El::Device::CPU; }
```

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,198 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+  if (layout_str.empty())             { layout = data_layout::DATA_PARALLEL; }
+  if (layout_str == "data_parallel")  { layout = data_layout::DATA_PARALLEL; }
+  if (layout_str == "model_parallel") { layout = data_layout::MODEL_PARALLEL; }
+  return layout;
+}
+
+El::Device get_layer_device_from_pbuf(const lbann_data::Layer& proto_layer,
+                                      bool GPUs_disabled) {
+  // Get parameters from prototext
+  El::Device device = El::Device::CPU;
```
benson31:
```suggestion

```

src/layers/data_type_layer.cpp
bvanessen: The `size_t` part is getting pushed to its own PR.

include/lbann/layers/layer.hpp
```diff
@@ -101,6 +108,11 @@ class Layer {
    *  human-readable, name.
    */
   inline void set_name(const std::string name) { m_name = name; }
+  /** Get a string representing the layer datatype
+   */
+  virtual std::string get_datatype_name() const {
```
timmoon10: Logically speaking, it doesn't make sense for the base layer class to have a single datatype. For example, what about a convolution layer that takes in `half`s and outputs `float`s? It does make sense for `data_type_layer`s to have a single type, since those explicitly have one type for their inputs and outputs.

It's not a major thing, but it does muddle the logical notion of an abstract layer.

src/layers/layer.cpp
```diff
@@ -105,7 +105,7 @@ description Layer::get_description() const {
 
   // Construct description object
   std::stringstream ss;
-  ss << get_name() << " (" << get_type() << ")";
+  ss << get_name() << " (" << get_type() << "<" << get_datatype_name() << ">" << ")";
```
timmoon10: This produces output like:
```
    layer15 (softmax<float>)
      Input dimensions: 10 (from fully connected layer "layer14")
      Output dimensions: 10 (to split layer "layer15_split")
      Data layout: data-parallel
      Device: GPU
```
I'd prefer something like:
```
    layer15 (softmax)
      Input dimensions: 10 (from fully connected layer "layer14")
      Output dimensions: 10 (to split layer "layer15_split")
      Data layout: data-parallel
      Device: GPU
      Data type: float
```
I think this has several advantages:
- The header is less cluttered.
- It allows us to move `get_datatype_name` out of the base `layer` class to `data_type_layer`, which is logically cleaner. `data_type_layer` can override `get_description` to add the datatype info.

src/proto/factories/layer_factory.cpp
```diff
@@ -110,8 +113,196 @@ std::vector<El::Int> get_slice_points_from_reader(const generic_data_reader* dr,
...
+    if (proto_layer.has_input()) {
+      // Input layers must be on CPU
+      return El::Device::CPU;
+    }
+  }
+#endif // LBANN_HAS_GPU
+  return El::Device::CPU;
+}
+
+std::unique_ptr<Layer> build_fully_connected_layer_from_pbuf(
```
timmoon10: I presume we'll eventually move this into `src/layers/learning/fully_connected.cpp`? (And similar for `build_convolution_layer_from_pbuf`?)

bvanessen: Yes, we will move these after the PR.

include/lbann/layers/layer.hpp
bvanessen: So are you thinking that this should be something like `get_weights_datatype_name` or `get_activations_datatype_name`

src/layers/layer.cpp
bvanessen: Done

include/lbann/layers/layer.hpp
bvanessen: I think that we can clarify this when we get to multiple datatypes per layer.

benson31: Just my 2 cents: I think this should wait for a future PR. As @bvanessen notes, there's only one datatype currently associated with the layer. There might be multiple input layers to a given layer, and I don't know that their activation types should be directly accessible from "this" layer. When we split the activations type and the weights type, we can change the API.

Issue #1289, opened 2019-10-10
Bamboo tests with Python frontend can't switch between builds


timmoon10:
Our CI tests work well when we run with a single build per system, but I suspect things will go poorly if we try testing multiple compilers or multiple build configurations.

### Status quo
We load in the Python frontend by appending the appropriate install directory to `sys.path` and calling `import lbann`. After the first import, Python's import machinery will not reimport LBANN since it already exists in `sys.modules`.

### Solutions
- Use `importlibs` ([for future reference](https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path)) and manipulate `sys.modules` to precisely control which Python frontend we are using. It doesn't seem that Python is designed to handle repeatedly loading and deleting modules with the same name (I shudder thinking of submodules), so I expect this approach to be extremely delicate and to produce strange errors.
- Run PyTest separately for each build. I don't think this is too bad; we just need to make the Bamboo Bash scripts loop over the builds and pass the build paths into PyTest. This is also natural if we want to run the tests from the build directories rather than the source directory.

benson31:
I have preliminary effort directed at the second bullet point; I'd like @graham63 to take a look at that when she gets up to speed with the test framework.

benson31:
Just as additional commentary: test suites are almost always directed at the build objects rather than the installed binaries (since you may want test more binaries than you install, e.g.). This is exemplified in the glorious GNU tradition of `./configure && make && make check && make install` (usually the `make check` bit is excluded from colloquial renditions of this idiom, but many packages include a warning in `make install` along the lines of "If you haven't already run `make check`, you should do that", which also implies that it is to be done (or have been done) before `make install`). I understand why ours currently does not do that, but it would be good to make the transition to that paradigm. It will also imply a shift toward using only 1 of the 4 currently defined "tests per test" and thereby getting rid of the others.

[...]

PR #1372, opened 2019-12-13
Add option to pass custom executable to Bamboo tests

timmoon10:
This is a quick hack to allow users to pass a custom LBANN build to the Bamboo tests, e.g. Spack builds. It is similar to functionality in the old prototext-based tests (e.g. [this](https://github.com/LLNL/lbann/blob/e4c698f8258f41fc4919dac485d3e883a575b9d5/bamboo/unit_tests/test_unit_layer_identity.py#L46)).

Example usage:
```bash
python -m pytest test_something_something.py --exe=/path/to/install/bin/lbann
```

This is a short-term fix and will fail if you have previously called `scripts/build_lbann_lc.sh` and have anything inside `lbann/build` (see Issue #1289). The right solution in the long term is to make the test suite build-specific.

### DIFF
--- a/bamboo/common_python/tools.py
+++ b/bamboo/common_python/tools.py
@@ -722,21 +722,14 @@ def create_tests(setup_func,
         test_name = '{}_{}'.format(test_name_base, compiler_name)
 
         # Load LBANN Python frontend
-        build_names = {
-            'clang6': 'clang.Release.{}.llnl.gov'.format(cluster),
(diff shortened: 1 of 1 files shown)


timmoon10:
[Bamboo is red](https://lc.llnl.gov/bamboo/browse/LBANN-TIM246-2). The integration tests assume that the test functions have a specific structure. A test that allows users to pass in custom executables does not fit this structure.

PR #1374, opened 2019-12-14
fix bugs in using conduit_io_handle

JaeseungYeom:
- Check the existence of `is_open()` method, and use it to check whether a file handle is valid only when the method is available. conduit_io_handle provides no method to check is file is successfully open. However, std::fstream does.
- replace the hard-coded argument type of preload_helper() with file_handle_t.

Not to merge, but to communicate.

There are still a bunch of compilation error in data_store_conduit.cpp regarding the hard-coded `conduit::relay::io::hdf5_*()`

### DIFF
--- a/include/lbann/data_readers/data_reader_jag_conduit.hpp
+++ b/include/lbann/data_readers/data_reader_jag_conduit.hpp
@@ -38,7 +38,7 @@
 #include <map>
 #include <memory>
 
-//#define _USE_IO_HANDLE_
+#define _USE_IO_HANDLE_
 #ifdef _USE_IO_HANDLE_
 #include "lbann/data_readers/sample_list_conduit_io_handle.hpp"
 #else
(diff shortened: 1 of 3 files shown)


src/data_readers/data_reader_jag_conduit.cpp
```diff
@@ -926,6 +930,9 @@ void data_reader_jag_conduit::do_preload_data_store() {
 void data_reader_jag_conduit::load_list_of_samples(const std::string sample_list_file, size_t stride, size_t offset) {
   // load the sample list
   double tm1 = get_time();
+#ifdef _USE_IO_HANDLE_
```
davidHysom: This seems backwards to me -- I may be misunderstanding. It seems to me the #ifdef should be ifndef.

include/lbann/data_readers/data_reader_jag_conduit.hpp
```diff
@@ -38,7 +38,7 @@
 #include <map>
 #include <memory>
 
-//#define _USE_IO_HANDLE_
+#define _USE_IO_HANDLE_
```
benson31: Just FYI any symbol that starts with an underscore followed by a capital letter is reserved, per the C++ standard. I would recommend something like `DR_JAG_CONDUIT_USE_IO_HANDLE` or `LBANN_USE_IO_HANDLE`.

JaeseungYeom: good point but what does DR_ stands for?

benson31: "DATA_READER" That's just _so_ many characters...

davidHysom: I'm guessing: Data Reader???

PR #1378, opened 2019-12-17
Feature supporting half datatype

bvanessen:
Adding support for both CPU and GPU half-precision floats.


src/callbacks/check_gradients.cpp
```diff
@@ -116,11 +228,11 @@ void check_gradients::do_check_gradients(model& m) const {
   // For simplicity, we assume f(chi) ~ f(x), and | f'''''(xi) | ~ 1.
   // If step size is not specified, then we choose h so that
   //   E_fl <= sqrt(epsilon)
-  const DataType epsilon = std::pow(std::numeric_limits<DataType>::epsilon(), 0.9);
-  const DataType step_size = (m_step_size > DataType{0} ?
+  const EvalType epsilon = std::pow(std::numeric_limits<EvalType>::epsilon(), 0.9);
```
timmoon10: This means that we expect the numerical accuracy corresponding to `double`, so I'd expect all of our gradient checking with `float` layers to fail. Now that I think about it, having a default error tolerance is tricky with mixed precision models. I think we should make the default correspond to `float`, and require the user to manually put in a tolerance if they're working with a model with non-`float` layers.


```suggestion
  const EvalType epsilon = std::pow(std::numeric_limits<float>::epsilon(), 0.9);
```

include/lbann/models/model.hpp
```diff
@@ -420,7 +435,7 @@ class model {
    *  is just used to create copies and is not actually used for
    *  optimization.
    */
-  optimizer* m_default_optimizer = nullptr;
+  std::unique_ptr<lbann_data::Optimizer> m_default_optimizer_msg;
```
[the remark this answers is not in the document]
benson31: Pointer makes `sizeof(model)` smaller; it also handles an optional field cleanly in a pre-C++17 world. I can't wait until we get to 2017.

include/lbann/utils/factory_error_policies.hpp
```diff
@@ -22,13 +22,13 @@ namespace lbann
...
 {
-  std::unique_ptr<ObjectT> handle_unknown_key(KeyT const&) const
+  std::unique_ptr<ObjectT> handle_unknown_key(KeyT const& key) const
   {
     // This could be expanded to print the key, but that would
     // assume that either the key can be inserted into a stream or
     // that the key can be converted to a string, which isn't
     // necessarily the case.
-    LBANN_ERROR("Unknown key detected.");
+      LBANN_ERROR("Unknown key \"", key, "\" detected.");
```
benson31:
```suggestion
    LBANN_ERROR("Unknown key \"", key, "\" detected.");
```

PR #1380, opened 2019-12-18
Layer factory revamp

benson31:
This PR offers an improved factory-based implementation of `construct_layer`. It does not replace `construct_layer_legacy`, but it shows the path forward for doing so.

I will be profoundly surprised if this doesn't break the python FE. I'm waiting for bamboo to tell me how badly its broken before trying to fix it.


benson31:
The python front end is now updated to emulate support for enums -- IMO, there's no advantage to "real enums" (which python also only emulates via the `enum` package). The support here provides the illusion of scoped enums, e.g., `lbann.DataType.FLOAT` and `lbann.DataType.DOUBLE`. Detection is automated, so `lbann.DataType.HALF` will automatically appear when it's added to the protobuf definition.

The Bamboo tests do not cover this. I have modified the first convolutional layer in `applications/vision/lenet.py` to the following:
```
x = lbann.Convolution(images,
                      name = "DoubleConv",
                      datatype = lbann.DataType.DOUBLE,
                      num_dims = 2,
                      num_output_channels = 6,
                      num_groups = 1,
                      conv_dims_i = 5,
                      conv_strides_i = 1,
                      conv_dilations_i = 1,
                      has_bias = True)
```
Running the code prints:
```
    DoubleConv (convolution)
      Input dimensions: 1x28x28 (from identity layer "layer2")
      Output dimensions: 6x24x24 (to ReLU layer "layer5")
      Weights: DoubleConv_kernel (6x1x5x5), DoubleConv_bias (6)
      Data layout: data-parallel
      Device: GPU
      Data type: double
      Convolution dimensions: 5, 5
      Strides: 1, 1
      Pads: 0, 0
      Dilations: 1, 1
      Groups: 1
      Bias: enabled
```
Note that the data type is detected correctly.

benson31:
[Bamboo is green.](https://lc.llnl.gov/bamboo/browse/LBANN-BEN17/latest) Note that my bamboo builds "weekly" tests.

benson31:
Just for documentation's sake: The current design of the layer factory requires that the factory functions be "instantiatable" for *the entire tensor product* of `Type x Layout x Device`. Note that this is ONLY the factory function, not the class itself.

I have been thinking about whether this is the right design and I have come to the conclusion that it probably *is* the right design for the following reasons. First, it greatly simplifies the factory implementation as all factory functions are only enumerated once. Second, this is really a runtime concern because we have no static control over what people will pass to the prototext front-end. Third, it allows us to provide improved error reporting. If we specialized the factories and only registered builders with valid concrete products, the best we would get is the generic factory "key error" exception. Perhaps a clever user could parse the stack trace and figure out exactly what they did wrong. But if we allow registrations of builders that will throw if they cannot create a concrete product, they can give a more detailed exception, like "Could not build XYZ Layer with Layout = BLAH and Device = WHATEVER". Certainly for non-developer users, this is more useful than "Unknown key detected. <Dump stack trace.>".

In any event, this introduces some overhead (which would be greatly reduced if we went to C++17 (*cough* it's almost 2020 *cough*)). This PR illustrates how to manage this with the convolution layer. The trick is that there can be no static reference to the invalid class type, in this case `convolution_layer<T, MODEL_PARALLEL, D>` (for any `T` and `D`). A C++11-friendly solution is shown in `src/layers/learning/convolution.cpp`.

src/proto/factories/layer_factory.cpp
```diff
@@ -335,12 +259,22 @@ std::unique_ptr<Layer> construct_layer_legacy(
...
+                                              data_layout::DATA_PARALLEL,
+                                              Device>>(
+                                                comm,
+                                                num_parallel_readers,
+                                                data_readers,
+                                                !params.data_set_per_model(),
+                                                target_mode);
+      }
+      else {
+        LBANN_ERROR("Input layers are only valide with "
```
bvanessen:
```suggestion
        LBANN_ERROR("Input layers are only valid with "
```

benson31: Fixed

PR #1381, opened 2019-12-18
Add PyTest test for sigmoid binary cross entropy layer

timmoon10:
I've also fixed some mistakes from copy-pasting.

[Bamboo is green](https://lc.llnl.gov/bamboo/browse/LBANN-TIM248-1).

### DIFF
--- a/bamboo/unit_tests/test_unit_layer_sigmoid_binary_cross_entropy.py
+++ b/bamboo/unit_tests/test_unit_layer_sigmoid_binary_cross_entropy.py
@@ -18,11 +18,11 @@ import tools
 # the functions below to ingest data.
 
 # Data
-# Note: MAE is not differentiable when the two inputs match, so we
-# make sure inputs have separated values.
-np.random.seed(201910248)
-_samples = np.random.uniform(-0.25, 0.25, size=(27,2,7)).astype(np.float32)
-_samples[:,1,:] += np.random.choice([-1.0,1.0], size=(27,7))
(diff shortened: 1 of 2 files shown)


bamboo/unit_tests/test_unit_layer_sigmoid_binary_cross_entropy.py
```diff
@@ -0,0 +1,205 @@
...
+    obj.append(z)
+    metrics.append(lbann.Metric(z, name='data-parallel layout'))
+
+    # NumPy implementation
+    vals = []
+    for i in range(num_samples()):
+        x = get_sample(i).astype(np.float64)
+        x0 = x[:slice_size]
+        x1 = x[slice_size:]
+        y = -x1 * np.log1p(np.exp(-x0)) - (1-x1) * np.log1p(np.exp(x0))
```
samadejacobs: This doesnt exactly match x1 * log(1 + exp(-x0)) + (1 - x1) * (x0 + log(1 + exp(-x0)) from [TF](https://www.tensorflow.org/api_docs/python/tf/nn/sigmoid_cross_entropy_with_logits?version=stable)? Just checking, I could be wrong.

timmoon10: It's mathematically the same. They rearranged stuff to avoid numerical errors, but this test isn't in a regime where that is a big concern.

PR #1387, opened 2019-12-20
Updated the I/O layers to properly support ETI with multiple data types

bvanessen:
Updated the I/O layers to properly support ETI with multiple data
types.  Added the TensorDataType to the generic_input_layer and
templated the io_buffers on the TensorDataType.  Also introduced an
IODataType for use within the io_buffers that is set to DataType.
This allows the data readers to operate with the compile defined
DataType.  The covnersion from IODataType to TensorDataType happens
inside of the io_buffer when the local matrices are distributed to the
input layer.


bvanessen:
Bamboo is green

PR #1390, opened 2019-12-23
Support int16 responses for the npz data reader

oyamay:
This PR updates the npz data reader to support int16 data type for responses. This does not change its behavior when other data types are used.


oyamay:
@timmoon10 I'm not sure whether this PR is relevant to you, but I would appreciate if someone reviews this PR.

PR #1392, opened 2020-01-03
bug fix: need to tell data_store that owner_maps were exchanged (actu…

davidHysom:
…ally, rebuilt by the data_reader)

emergency bug fix


samadejacobs:
I am getting a core dumped with RAS npz reader, backtrace shows something in conduit,hdf5 etc I am wondering if this fixes that?

PR #1395, opened 2020-01-06
Python frontend for JAG WAE

samadejacobs:
Here is the first in the JAG/ICF models with python frontend interface. This is for the Wasserstein autoencoder. The PR include a README of network/data descriptions and some results including predicted images and scalars.


applications/physics/data/jag_100Kdata.prototext
```diff
@@ -0,0 +1,13 @@
+data_reader {
+  reader {
+    name: "numpy"
+    role: "train"
+    shuffle: true
+    data_filename: "/p/lustre2/brainusr/datasets/jag/jag100K1vw_train.npy"
+    data_filename: "/usr/workspace/wsa/jacobs32/git.samadejacobs.lbann/model_zoo/jag_uq/jaguq_train_100K.npy"
```
bvanessen: This path shouldn't be hardcoded into your home directory and repo.  If this is necessary, lets add this npy file into the datasets directory.

samadejacobs: Oh I see, that is a typo, it should be pointing to /p/lustre2/brainusr/datasets/jag/jag100K1vw_train.npy. The file is 13G so I suggest we point to Lustre instead of user directory.

samadejacobs: datafile name now points to shared Lustre directory instead of my home directory

PR #1397, opened 2020-01-08
Instance norm layer

timmoon10:
This PR implements the instance norm layer, which normalizes each channel in each data sample to have zero mean and unit variance. It's similar to the layer norm layer, which normalizes the entire data sample. It is equivalent to [InstanceNorm1d in PyTorch](https://pytorch.org/docs/stable/nn.html#torch.nn.InstanceNorm1d) and [LayerNormalization in TensorFlow](https://www.tensorflow.org/api_docs/python/tf/keras/layers/LayerNormalization?version=stable).

This is useful for attention models, where we want to independently normalize each vector in a sequence.

[Bamboo is green](https://lc.llnl.gov/bamboo/browse/LBANN-TIM254-1).


src/layers/regularizers/instance_norm.cpp
```diff
@@ -0,0 +1,237 @@
...
+        sqsum += x * x;
+      }
+    }
+  }
+
+  // Compute statistics from sums
+  //   mean = sum(x_i) / n
+  //   var = ( sum(x_i^2)/n - mean^2 ) * n/(n-1)
+  LBANN_OMP_PARALLEL_FOR_COLLAPSE2
+  for (El::Int k = 0; k < local_mini_batch_size; ++k) {
```
ndryden: I think this loop and the next one can be fused together?

timmoon10: Good point. We can do something similar in the GPU implementation without too much effort, but it will be ugly. I'm thinking that reducing runtime by 1/3 might be worth it though...

ndryden: Yes, I think so. Can do it here or in a separate PR. 🚀

timmoon10: Done.

ndryden:
:shipit:

timmoon10:
I've updated this PR to use the builder function pattern from #1408. I've modified it a bit so that the builder functions are instantiated in `instance_norm.cpp` instead of a separate `_builder.cpp` file. I find this cleaner and more convenient when I'm churning out many layers (see #1403 and #1404). Pinging @benson31.

src/layers/regularizers/instance_norm.cpp
```diff
@@ -0,0 +1,335 @@
...
+#include "lbann/macros/instantiate.hpp"
+#undef PROTO
+
+#ifdef LBANN_HAS_GPU
+#define PROTO(T)                                        \
+  extern template class instance_norm_layer<            \
+    T, data_layout::DATA_PARALLEL, El::Device::GPU>
+#include "lbann/macros/instantiate.hpp"
+#undef PROTO
+#endif // LBANN_HAS_GPU
```
benson31: This is fine and clearly it works. However, we put a fair bit of effort to remove as much of the explicit device permutations (things like `#ifdef LBANN_HAS_THIS_OR_THAT_DEVICE` from the developer-facing ETI garbage, and this addition is explicitly adding that back in. This is why I introduced the `*_builder.cpp` files to explicitly handle cases in which the layers require separate instantiation (e.g., because they contain CUDA device code) but their builder functions are not device-specific. There is no explicit special device handling, and the `instantiate_device.hpp` macro include implies that the `PROTO_DEVICE` spec is valid across all known device types (as it does here as well -- the main differentiating point is the first clause). @bvanessen do you have an opinion?

Issue #1399, opened 2020-01-09
Absolute sample counts are broken


ndryden:
If you specify an `absolute_sample_count` for a data reader, the number of samples used will always be 0. Obviously, this messes a few things up.

The root cause is [this line](https://github.com/LLNL/lbann/blame/develop/src/data_readers/data_reader.cpp#L503) in `generic_data_reader::get_percent_to_use`:
```cpp
r = count / get_num_data();
```
Note that `r` is a `double` but `count` is a `size_t` and `get_num_data` returns an `int`, so the division will always produce 0 (except in the case that `count == get_num_data()`), which is then cast to a `double`.  So `get_percent_to_use` always returns `0.0`.

`resize_shuffled_indices` then resizes `m_shuffled_indices` to be of size 0.

A quick hack would be to change the above line to
```cpp
r = double(count) / double(get_num_data());
```
but this is inaccurate in a large number of cases (in the case of the ImageNet training dataset, 74087 cases). This is because of floating point errors: we can't assume `r * get_num_data() == count`.

This appears to have been introduced in dbc31255e2aaf7f7461cd8e6e8f6f12525e3264b. Reverting to the method used there (where the absolute sample count is used explicitly) is probably the best solution.

---

We should also add a test to check for correctness of the absolute sample count. This is not the first time we've run into bugs with it (#181). It's rarely used, but quite useful in those rare cases (esp. testing).

davidHysom:
adressed/corrected by PR #1400

PR #1400, opened 2020-01-09
data readers: absolute sample counts are broken

davidHysom:
This PR addresses Issue #1399.

An exception is now thrown if the number of train samples is zero.

### DIFF
--- a/src/proto/proto_common.cpp
+++ b/src/proto/proto_common.cpp
@@ -511,6 +511,11 @@ void init_data_readers(
         reader_validation->get_data_store_ptr()->compact_nodes();
       }
 
+      size_t ntrain = reader->get_num_data();
+      if (ntrain == 0) {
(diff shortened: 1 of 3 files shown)


src/proto/proto_common.cpp
```diff
@@ -511,6 +511,11 @@ void init_data_readers(
         reader_validation->get_data_store_ptr()->compact_nodes();
       }
 
+      size_t ntrain = reader->get_num_data();
+      if (ntrain == 0) {
+        LBANN_ERROR("num train samples = 0; something is wrong");
```
ndryden: @bvanessen I just want to double-check that we have no use cases where there are no training samples.

davidHysom: My take from today's weekly mtg: OK to throw exception for now; revisit if someone runs an experiment where there's a train reader and the reader should have zero samples.

PR #1401, opened 2020-01-13
Sample list image bcast

JaeseungYeom:
This PR includes the same commits from https://github.com/LLNL/lbann/pull/1136#issue-303149516. Additional commits are for:
- Performance improvements
   - Only the trainer-master loads the whole sample list file into a buffer, and then broadcasts it. 
      - Enabled by the command line option `--load_full_sample_list_once`.
      - Everyone constructs a sample list from the buffer in the memory.
      - broadcasted buffer content is not copied but reused by the input stream to sample list
   - checking the existence of data file is now optional, enabled by `--check_data` and turned off by default.
   - The label map `m_labels` (from sample name to label) in `data_reader_image` is now implemented as a `vector` instead of an `unordered_map`.

- Support for the original Imagenet data list file
   - Users can use the original file as is without generating a sample list file.
   - The file is internally used both for generating a sample list and for constructing a label list.

- Misc.
   - improve `load()` interface for handling istream
      mainly for making sample list to ingest bcast buffer content but also to accommodate the on-the-fly generation of sample list header
   - updated the header format of the sample list written out.
   - The prototext option `keep_sample_order` is also available as a command line option
   - clean-up: a private member variable is now protected and shared with derived classes

- Status
   - Passed  bamboo tests.


JaeseungYeom:
> ========================================
> Questions:
> 
> "data_filedir prototext variable overrides the data location written in the 3rd header line of the sample list file. As usual, the command line option --data_filedir_* overrides the former."
> 
> I think "--data_filedir_*" means: --data_filedir= ??

`*` was supposed to be replaced by either `train`, `validate` or `test`
This is a pre-existing feature before this PR. With imagenet, the directories for train, and validate are different, and thus `--data_filedir=`, which uses the same directory for all the data readers with different roles, would not have worked.
 
> The sample lists have data locations on both 3rd and fourth lines;
The third line is the location of the data as we used to indicate by `data_filedir` in the data reader prototext.
The fourth line is the path to the label file. 

> # Failures:
> On pascal, I used data_reader_imagenet_lassen.prototext and included
> on my cmd line:
> --data_filedir=/p/lustre2/brainusr/datasets/ILSVRC2012/original/train/
> I got this exception: file ./train_sample_list.txt :: data root directory '/p/gpfs1/brainusr/datasets/ILSVRC2012/original/train/' does not exist.
> Same behavior when I added:
> data_filedir: "/p/lustre2/brainusr/datasets/ILSVRC2012/original/train"
> to data_reader_imagenet_lassen.prototext

Thanks for catching this. I replicated the issue and fix is ready. 

> Question:
> "Support for the original Imagenet data list file" - I'm unsure how this is supposed to work. I modified the train reader prototext:
> sample_list: "/p/gpfs1/brainusr/datasets/ILSVRC2012/original/labels/train_c0-9.txt"
> And got error: Unknown sample list type: N01440764/N01440764_10026.JPEG

I should have been clearer. I meant the original Imagenet data list file with the original prototext you had before this PR, in which the path to the file is specified via `data_filename`. This is for the backward compatibility, and to allow users avoid preparing a sample list and modifying the prototext to use it. Once you use 'sample_list:' it expects the sample list header, and the first line of the header indicates the type of sample list. This allows us to use the current bamboo unit tests, that relies on a generated data reader prototext that does not use sample list, without modification.

davidHysom:
I'm still having problems with this (see below). I included the cmd line flag: --data_fildir_train=/p/lustre2j/brainusr/datasets/ILSVRC2012/original/train
This worked for the third line (train dir), but not for the fourth (label dir)
Looks to me that one needs to specify two directories ?? Or perhaps I'm misunderstanding.

> > ========================================
> > Questions:
> > "data_filedir prototext variable overrides the data location written in the 3rd header line of the sample list file. As usual, the command line option --data_filedir_* overrides the former."
> > I think "--data_filedir_*" means: --data_filedir= ??
> 
> `*` was supposed to be replaced by either `train`, `validate` or `test`
> This is a pre-existing feature before this PR. With imagenet, the directories for train, and validate are different, and thus `--data_filedir=`, which uses the same directory for all the data readers with different roles, would not have worked.

JaeseungYeom:
> I'm still having problems with this (see below). I included the cmd line flag: --data_fildir_train=/p/lustre2j/brainusr/datasets/ILSVRC2012/original/train
> This worked for the third line (train dir), but not for the fourth (label dir)
> Looks to me that one needs to specify two directories ?? Or perhaps I'm misunderstanding.
> 
We do not override the label file path on the 4th line of the sample list header. 
If you are running in the backward compatibility mode, it should be specified via `data_filename=` as it used to be.

JaeseungYeom:
Ideally, when a different path is required, the sample list should be updated accordingly. However, sometimes allowing it to be overridable makes it easy for automated execution. The data location overriding feature is added as a convent test plugin. If it is further desired to allow overriding the label file path, I can add such a feature. That means adding another command line option to learn. However, these are not necessary features as one can either update the sample list or override the path to the sample list by the command line option `--sample_list-*=`.

davidHysom:
> If it is further desired

I think it's OK as it is now

bvanessen (approved):

PR #1402, opened 2020-01-13
Fix lassen bamboo tests

graham63:
Bamboo is green on Lassen. (https://lc.llnl.gov/bamboo/browse/LBANN-GRAH1-5)


benson31:
Superseded by #1406.

PR #1405, opened 2020-01-16
Delete data reader jag

davidHysom:
From Jae-Seung's email (1/15/2020): "data_reader_jag was developed with the old data set that had a different format than what we use. We kept it around for a couple of reasons, but I do not see the reason to keep it any more. One of the reasons was that the Rushil and the company were using the old data set even after everyone else switched to the newer format. At this point it is not relevant any more


bvanessen:
I think that we can remove this, but I know that for a while @samadejacobs was using it too.

PR #1407, opened 2020-01-16
Cnpy extended

davidHysom:
Adds functionality to the cnpy package to open an npz file and return an std::istream; load an npz file from a given std::istream. This functionality is needed to enable data_reader_npz_ras_lipid (and possibly future npz based reader) to work with our sample_list class.


davidHysom:
no longer relevant

PR #1408, opened 2020-01-17
Layer refactorying

benson31:
Adds a bunch of layers to the factory. Once #1378 merges, I'll have to update this, I'm sure.


model_zoo/lbann.cpp
```diff
@@ -132,10 +132,13 @@ int main(int argc, char *argv[]) {
     }
     El::ReportException(e);
     return EXIT_FAILURE;
-  } catch (std::exception& e) {
+  }
+#if 0
```
bvanessen: Is this to allow normal exceptions to propagate all of the way out?

benson31: That was an accidental commit, but yes. I had some `bad_cast` problems I was debugging. I'll remove this.

src/layers/learning/entrywise_scale_bias.cpp
```diff
@@ -143,13 +143,17 @@ void entrywise_scale_bias_layer<TensorDataType, Layout, Device>::bp_compute() {
...
+#define PROTO(T)                                                        \
+  template class entrywise_scale_bias_layer<                            \
+    T, data_layout::DATA_PARALLEL, El::Device::CPU>;                    \
+  template class entrywise_scale_bias_layer<                            \
+    T, data_layout::MODEL_PARALLEL, El::Device::CPU>;                   \
+  LBANN_LAYER_BUILDER_ETI(entrywise_scale_bias, T, El::Device::CPU)
 
 #define LBANN_INSTANTIATE_CPU_HALF
 #include "lbann/macros/instantiate.hpp"
+#undef PROTO
```
bvanessen: Do we need this undef here?

src/layers/transform/concatenate.cpp
```diff
@@ -326,5 +326,6 @@ void bp_compute_impl(
 
 #define LBANN_INSTANTIATE_CPU_HALF
 #include "lbann/macros/instantiate.hpp"
+#undef PROTO
```
bvanessen: I thought that we didn't need to do this in cpp files.

src/layers/transform/concatenate.cu
```diff
@@ -470,13 +470,13 @@ void bp_compute_impl(
...
-#define PROTO(T)                                        \
-  template class concatenate_layer<                     \
-    T, data_layout::DATA_PARALLEL, El::Device::GPU>;    \
-  template class concatenate_layer<                     \
-    T, data_layout::MODEL_PARALLEL, El::Device::GPU>
+#define PROTO(T)                                                        \
+  template class concatenate_layer<                                     \
+    T, data_layout::DATA_PARALLEL, El::Device::GPU>;                    \
+  template class concatenate_layer<                                     \
+    T, data_layout::MODEL_PARALLEL, El::Device::GPU>;
```
bvanessen: pretty sure you don't need the last semi-colon

src/layers/learning/entrywise_scale_bias.cpp
benson31: No, will remove.

src/layers/transform/concatenate.cpp
benson31: We don't. Will remove. (Artifact of some temporary changes)

src/layers/transform/concatenate.cu
benson31: Correct, also an artifact of other changes.

benson31:
I'm working on the merge with develop. There are some issues, but they're small.

include/lbann/utils/cudnn.hpp
```diff
@@ -84,6 +84,23 @@ class Layer;
...
 namespace cudnn {
 
+template <typename T>
+struct ScalingParameterT
+{
+  using type = T;
+};
+
+template <typename T>
+using ParamType = typename ScalingParameterT<T>::type;
```
timmoon10: We should document and link to the [cuDNN docs](https://docs.nvidia.com/deeplearning/sdk/cudnn-archived/cudnn_701/cudnn-user-guide/index.html#scaling-parameters).

timmoon10:
As we've discussed, I'm not thrilled about the proliferation of `_builder.cpp` files. I'll update my newer, more experimental layers (#1397, #1403, #1404) to keep everything contained in the main source file.

PR #1414, opened 2020-01-23
Feature cleanup spack env

bvanessen:


spack_environments/build_lbann.sh
```diff
@@ -0,0 +1,247 @@
...
+	source ${SPACK_ENV_DIR}/${CENTER}/setup_modules.sh
+    else
+	CENTER="llnl_lc"
+    fi
+fi
+
+SPACK_ARCH=$(spack arch)
+
+SCRIPT=$(basename ${BASH_SOURCE})
+BUILD_DIR=${LBANN_HOME}/build/spack
```
timmoon10: I think it would be nicer to have a separate build directory for each system so that different builds don't clobber each other. Something like `lbann/build/gnu.Release.lassen.gov/build`.

bvanessen: Done.

spack_environments/llnl_lc/externals-linux-rhel7-broadwell.sh
```diff
@@ -8,9 +10,12 @@
...
       modules: {}
-      compiler: [gcc@7.3.0 arch=linux-rhel7-broadwell]
+EOF
+)
 
+EXTERNAL_PACKAGES=$(cat <<EOF
     cmake::
+      buildable: True
       variants: ~openssl ~ncurses
       version: [3.12.1]
```
benson31: This should be bumped to at least 3.15. DiHydrogen, e.g., will depend on a more recent CMake.

bvanessen: Grr.  LC doesn’t have anything past 3.14.

bvanessen: I think that we will bump this in another PR.

PR #1422, opened 2020-01-28
Incorporate bead euclid distances and z-values into data reader

davidHysom:


davidHysom:
no one has asked about this for over a year, so assuming it's no longer of interest, and closing.

PR #1423, opened 2020-01-29
Fix bug in LTFB with checkpoint file communication

### LINKED ISSUE #1420: missing archive files with LTFB by checkpoint
I am trying to run the LTFB example with the `checkpoint` method, and getting errors as
`Archive file not found: model0_trainer0_step4training_dr.xml`
`Archive file not found: <unknown>_dr.xml`
The `sendrecv` method still works.

I suspect that the `_dr.xml` comes from `generic_data_reader::load_from_checkpoint_shared`.
`<unknown>` comes from an unused filename in persist object.

I suspect that this has to do with the fact that
`generic_input_layer::save_to_checkpoint_shared()` contains `if(p.get_cb_type() == callback_type::execution_context_only){}`
However, `load_from_checkpoint_shared()` does not have that if-conditional.
For LTFB, we do not want to load data reader state, and therefore, we would need the if-conditional.

timmoon10:
This closes #1420 by making three changes:
- The input layer only loads a checkpoint when the execution context is required (see @JaeseungYeom's comment in #1420).
- Update the checkpoint loading in LTFB to match the checkpoint callback.
- Make sure that checkpoint file names are based on the training step, not the validation step. The validation step changes over the course of an LTFB tournament round since each trainer evaluates multiple models on the validation set, so LTFB was trying to load the wrong checkpoint files.

LTFB with checkpoint file communication works for me (LeNet; 1 Pascal node with 2 proc/node; 1 proc/trainer). [Bamboo is green](https://lc.llnl.gov/bamboo/browse/LBANN-TIM261-1).

### DIFF
--- a/include/lbann/layers/io/input/generic_input_layer.hpp
+++ b/include/lbann/layers/io/input/generic_input_layer.hpp
@@ -768,34 +768,37 @@ class generic_input_layer : public io_layer<TensorDataType> {
   bool load_from_checkpoint_shared(persist& p) override {
     // save state of data readers from input layer
     data_reader_map_t::const_iterator it;
+    if(p.get_cb_type() == callback_type::execution_context_only
+       || p.get_cb_type() == callback_type::full_checkpoint){
 
-    it = this->m_data_readers.find(execution_mode::training);
-    if ((it != this->m_data_readers.end()) && it->second) {
-      (it->second)->load_from_checkpoint_shared(p, execution_mode::training);
+      it = this->m_data_readers.find(execution_mode::training);
+      if ((it != this->m_data_readers.end()) && it->second) {
+        (it->second)->load_from_checkpoint_shared(p, execution_mode::training);
+      }
+      it = this->m_data_readers.find(execution_mode::testing);
+      if ((it != this->m_data_readers.end()) && it->second) {
+        (it->second)->load_from_checkpoint_shared(p, execution_mode::testing);
+      }
+      it = this->m_data_readers.find(execution_mode::validation);
+      if ((it != this->m_data_readers.end()) && it->second) {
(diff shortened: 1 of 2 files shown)


timmoon10:
When I run checkpoint file LTFB on Pascal, LeNet works with 16 nodes, 4 trainers, 8 procs/trainer. However, I experience problems with AlexNet:
- 16 nodes, 4 trainers, 8 procs/trainer HANGS.
- 16 nodes, 8 trainers, 4 procs/trainer WORKS.
- 8 nodes, 2 trainers, 8 procs/trainer HANGS.
- 8 nodes, 4 trainers, 4 procs/trainer WORKS.
- 8 nodes, 8 trainers, 2 procs/trainer WORKS.
- 8 nodes, 16 trainers, 1 proc/trainer WORKS.
- 4 nodes, 2 trainers, 4 procs/trainer WORKS.

Debugging with TotalView, processes seem to be stuck when loading the checkpoint files for the partner model. The stack frame looks something like:
- `MPI_Bcast`
- `El::Read(AbsDistMat&, ...)`
- `lbann::persist::read_distmat`
- `lbann::sgd::load_from_checkpoint_shared`
- `lbann::weights::load_from_checkpoint_shared` (for the last FC layer matrix weights)

Some processes seem to have gotten past this point and are already evaluating the partner model. I am not sure if this is the cause of the deadlock (it's possible some trainers get stuck while other trainers can proceed normally). It's possible that running with 8 procs/trainer just makes the `MPI_Bcast` obscenely slow, but that doesn't explain why we specifically get stuck when loading the SGD velocity for the last FC layer.

I'll merge this for now, and I'll continue debugging later.

JaeseungYeom:
Another suspicious pattern is that it hands with 8 procs per trainer.

PR #1431, opened 2020-02-07
Mpi err catching

davidHysom:
lbann currently dies with non-informative "abort" message on MPI errors. This PR is an attempt to catch and report such errors. It has the potential benefit of saving hours of debugging/head scratching.

The proposed solution is not perfect. Please improve it if you can. I understand that, after an MPI error is thrown, the state of MPI (and, presumably, the program as a whole) is undefined. However, attempting to catch and report MPI errors is superior to simply ignoring them.


davidHysom:
will reopen; this PR probably includes code changes that it shouldn't.

PR #1432, opened 2020-02-07
Mpi error catching

davidHysom:
lbann currently dies with non-informative "abort" message on MPI errors. This PR is an attempt to catch and report such errors. It has the potential benefit of saving hours of debugging/head scratching.

The proposed solution is not perfect. Please improve it if you can. I understand that, after an MPI error is thrown, the state of MPI (and, presumably, the program as a whole) is undefined. However, attempting to catch and report MPI errors is superior to simply ignoring them.


model_zoo/lbann.cpp
```diff
@@ -43,6 +43,10 @@ int main(int argc, char *argv[]) {
   world_comm_ptr comm = initialize(argc, argv, random_seed);
   const bool master = comm->am_world_master();
 
+  MPI_Errhandler err_handle;
+  MPI_Comm_create_errhandler(lbann_mpi_err_handler, &err_handle);
```
timmoon10: Do we also need to deallocate the error handler with [`MPI_Errhandler_free`](https://www.open-mpi.org/doc/current/man3/MPI_Errhandler_free.3.php)? The documentation for [`MPI_Comm_create_errhandler`](https://www.open-mpi.org/doc/current/man3/MPI_Comm_create_errhandler.3.php) doesn't mention it and quick Googling doesn't find any guidelines. If we go that route, I think it would be nicer to put this inside of `lbann::initialize`/`lbann::finalize` or to wrap the `MPI_Errhandler` inside of a RAII object.

src/utils/lbann_library.cpp
```diff
@@ -387,4 +387,11 @@ void print_lbann_configuration(lbann_comm *comm, int io_threads_per_process, int
   std::cout << std::endl;
 }
 
+void lbann_mpi_err_handler(MPI_Comm *comm, int *err_code, ... ) {
+  char err_string[1024];
```
timmoon10:
```suggestion
  char err_string[MPI_MAX_ERROR_STRING];
```

davidHysom:
The data store bug that Sam ran into was an MPI error. I tested it on that, and it worked correctly.

model_zoo/lbann.cpp
benson31: If you want to live in a shiny world of RAII purity, you should free the resources you acquire.

In this case, `MPI_Errhandler_free` will do nothing useful. It won't deallocate the error handler until `MPI_COMM_WORLD` is deallocated, which won't happen until the end of the program (anything after `MPI_Finalize()` is undefined behavior in an MPI program). All program resources are freed and reclaimed by the OS at the end of the program, which will include this error handler. Depending on the MPI implementation you're using, you _might_ see a bit of "leaked" memory in a tool like LeakSanitizer or Valgrind, but this could be (should be) ignored.

src/utils/lbann_library.cpp
```diff
@@ -387,4 +387,11 @@ void print_lbann_configuration(lbann_comm *comm, int io_threads_per_process, int
...
+  int err_string_length;
+  MPI_Error_string(*err_code, &err_string[0], &err_string_length);
+  LBANN_ERROR("MPI threw this error: ", err_string);
```
benson31: Pro: This will give you a stack trace.
Con: The `__FILE__` and `__LINE__` expansion will always point to `lbann_library.cpp:394`.

davidHysom:
> lbann::initialize()

The error handler and MPI calls to same has been transported to lbann::initialize and lbann:finalize

PR #1447, opened 2020-02-19
Evaluate Transformer on WMT 2014 dataset

timmoon10:
This PR makes three changes:
- Modified the Transformer example in the NLP application directory to perform English-German translation on the WMT 2014 dataset.
- Added an evaluation script that performs greedy decoding with PyTorch and calculates the BLEU score between predicted and ground-truth sentences.
- Moved the Transformer model into the core Python frontend (`lbann.models.Transformer`). We also expose `TransformerEncoderLayer` and `TransformerDecoderLayer` to make it easier to implement other attention models like BERT.

The per-step runtime is somewhat better than the original paper (0.3 sec/step on 4 Pascal nodes vs. 0.4 sec/step on 8 P100s). There are definitely areas for performance optimization as well, especially in the slice and concatenation layers. After 2 epochs of training, I get plausibly German-ish sentences, although not at a point to get non-zero BLEU scores.


applications/nlp/transformer/evaluate.py
```diff
@@ -0,0 +1,282 @@
...
+import dataset
+import utils
+import utils.paths
+
+# ----------------------------------------------
+# Options
+# ----------------------------------------------
+
+# Hard-coded
+mini_batch_size = 64
```
samadejacobs: This doesn't have to match training batch size right?

timmoon10: Nope. I've added a note that you don't need to match.

timmoon10:
I do notice strange behavior on 4 Pascal nodes where LBANN consistently hangs after two epochs. If I reload a checkpoint, it trains for another two epochs before hanging again. I haven't dug too much into it, but in my current run it's frozen at step 149699 (halfway through epoch 9). I suspect an allreduce is failing.

PR #1450, opened 2020-02-21
Load pretrainedmodel callback

samadejacobs:
Callback to load pretrained model(s) from given directories. This callback supports loading pretrained weights of multiple models from different (specified) directories (e.g., pretrained autoencoder model, pretrained inverse model for cycle GAN).

PR #1452, opened 2020-02-24
LBANN implementation of MaCC surrogate model

samadejacobs:
LBANN implementation of full (that is WAE, inverse and forward models) Manifold and Cyclical Consistent (MaCC) model for ICF 
Things to note:
(1) This PR depends on #1450 
(2) With PR , user can load pretrained weights from multiple models, just specify directories as space separated string, that said, only pretrained WAE is needed for the MaCC surrogate model training
(4) README/documentation will be revised to reflect addition of surrogate model and new results from training WAE+surrogate model before this PR is merged.
(3) CNN version of WAE will be added before/after the PR is merged

### DIFF
--- a/applications/physics/ICF/pre_train_jag_wae.py
+++ b/applications/physics/ICF/pre_train_jag_wae.py
@@ -1,0 +1,109 @@
+import macc_models
+from os.path import abspath, dirname, join
+import google.protobuf.text_format as txtf
+
+# ==============================================
+# Setup and launch experiment
+# ==============================================
+
+# Default data reader
+model_zoo_dir = dirname(dirname(abspath(__file__)))
+data_reader_prototext = join(model_zoo_dir,
+                             'data',
+                             'jag_100Kdata.prototext')
+
+
+def list2str(l):
+    return ' '.join(l)
+
+def construct_model():
+    """Construct LBANN model.
+
+    JAG Wasserstein autoencoder  model
+
+    """
+    import lbann
+
+    # Layer graph
+    input = lbann.Input(target_mode='N/A',name='inp_data')
+    # data is 64*64*4 images + 15 scalar + 5 param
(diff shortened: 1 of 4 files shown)


applications/physics/ICF/pre_train_jag_wae.py
```diff
@@ -0,0 +1,109 @@
...
+    model = construct_model()
+    # Setup optimizer
+    opt = lbann.Adam(learn_rate=0.0001,beta1=0.9,beta2=0.99,eps=1e-8)
+    # Load data reader from prototext
+    data_reader_proto = lbann.lbann_pb2.LbannPB()
+    with open(data_reader_prototext, 'r') as f:
+      txtf.Merge(f.read(), data_reader_proto)
+    data_reader_proto = data_reader_proto.data_reader
+
+    status = lbann.run(trainer,model, data_reader_proto, opt,
```
bvanessen: Should this be updated to use the proper contrib launcher so that it can work with slurm or lsf?

applications/physics/ICF/train_macc_surrogate.py
```diff
@@ -0,0 +1,145 @@
...
+    model = construct_model()
+    # Setup optimizer
+    opt = lbann.Adam(learn_rate=0.0001,beta1=0.9,beta2=0.99,eps=1e-8)
+    # Load data reader from prototext
+    data_reader_proto = lbann.lbann_pb2.LbannPB()
+    with open(data_reader_prototext, 'r') as f:
+      txtf.Merge(f.read(), data_reader_proto)
+    data_reader_proto = data_reader_proto.data_reader
+
+    status = lbann.run(trainer,model, data_reader_proto, opt,
```
bvanessen: Same comment about the contrib launcher.

applications/physics/ICF/pre_train_jag_wae.py
samadejacobs: Done!

PR #1454, opened 2020-02-24
Deprecate and delete old prototext files

bvanessen:
Deprecated and deleted old model prototext files for vision models
that have been migrated to the python front-end and the applications
directory.


bvanessen:
@timmoon10 It passed the nightly checks

PR #1457, opened 2020-02-26
Feature move checkpoint to trainer

bvanessen:
This PR moves the checkpoint callback to the trainer object so that the trainer can be explicitly responsible for the checkpoint and restart of the execution contexts, rather than relying on the model to save them indirectly.  The trainer then passes the checkpoint callback to all of its constituent models.


bvanessen:
This PR is passing bamboo.

src/callbacks/checkpoint.cpp
```diff
@@ -289,15 +291,15 @@ std::string checkpoint::find_latest_checkpoint(model *m, std::string& latest_fil
   // Update other ranks on where we are loading from.
   // TODO: we would want to prepend dir with the model name and model rank:
   // m->get_name() + '.' + std::to_string(comm->get_trainer_rank()) + '.'
-#if 1
   header_t<max_len_dirname> header;
+  bzero(&header, sizeof(header_t<max_len_dirname>));
```
benson31: Use `std::memset` instead. `bzero` is a legacy function, the use of which is deprecated.
```suggestion
  std::memset(&header, 0x0, sizeof(header_t<max_len_dirname>));
```

bamboo/unit_tests/test_unit_lbann2_reload.py
```diff
@@ -84,14 +85,26 @@ def skeleton_lbann2_reload(cluster, executables, dir_name, compiler_name,
...
+        restart_dir=ckpt_dir + "/trainer0/",
         output_file_name=output_file_name,
         error_file_name=error_file_name,
         weekly=weekly)
     return_code_ckpt_2 = os.system(command)
     tools.assert_success(return_code_ckpt_2, error_file_name)
-#    os.system('rm lbann2_ckpt/model0-epoch*')
-#    os.system('rm lbann2_nockpt/model0-epoch*')
+    # Delete the checkpoint from the second model it confuses the test
+    fileList = glob.glob('{no_ckpt}/trainer0/model1.*'.format(no_ckpt=no_ckpt_dir))
```
benson31: This is not the right way to walk file trees in Python. But also, why not just `rmtree('{no_ckpt}/trainer0'.format(no_ckpt=no_ckpt_dir)`?

include/lbann/callbacks/callback.hpp
```diff
@@ -83,10 +84,14 @@ class callback_base {
...
-   *         set up).
+  /** @brief Called once to set up the callback on the trainer
+   */
+  virtual void setup(trainer *t) {};
+
+  /** @brief Called once to set up the callback on the model
+   *         (after all layers are set up).
    */
-  virtual void setup(model *m) {};
+  virtual void setup(model *m, const std::string& trainer_name) {};
```
benson31: This is more a general comment than a review: This is highlighting to me the poor design of callbacks. In particular, the API suggests that you can pass in any ol' `model*` and whatever the callback does will "just work". But many callbacks keep state and some even keep model-specific state. If the intention is that callbacks are "logically stateless", we have several callbacks that defy logic but the API is fine. If the intention is that callbacks can have state, there shouldn't be a `model*` to any API call except `setup()`, which would cache its input `model*` and use that when its hook functions are called. This could be managed in the base class with no significant changes to the derived classes.

The reason this change draws this API inconsistency to mind is that there should be no need for the `trainer_name` argument here as models are not shared among trainers. That is, this is a well-defined function: `trainer& get_trainer(model const&)` and we shouldn't need to expand the callback API with redundant information.

model_zoo/lbann.cpp
```diff
@@ -89,7 +89,8 @@ int main(int argc, char *argv[]) {
     lbann_data::Model *pb_model = pb.mutable_model();
 
     auto model = build_model_from_prototext(argc, argv, pb_trainer, pb,
-                                            comm.get(), opts, io_thread_pool, true);
+                                            comm.get(), opts, trainer->get_name(), io_thread_pool,
```
benson31: IMO, this is the wrong way to associate a model with a trainer... Why the string when you actually have a trainer object?

benson31: Also, the function name is `build_model_from_prototext`. There's so much garbage in this call that is *not* related to constructing `model` objects from prototext descriptions of them... This feels like adding more of that rather than clarifying the interface.

src/callbacks/hang.cpp
```diff
@@ -31,7 +31,7 @@
 namespace lbann {
 namespace callback {
 
-void hang::setup(model* m)
+void hang::setup(model* m, const std::string& trainer_name)
```
benson31:
```suggestion
void hang::setup(model* m, const std::string&)
```

src/callbacks/imcomm.cpp
```diff
@@ -59,7 +59,7 @@ void imcomm::set_weights_comm(weights *w, comm_type ct) {
   m_weights_params[w].ct = ct;
 }
 
-void imcomm::setup(model *m) {
+void imcomm::setup(model *m, const std::string& trainer_name) {
```
benson31:
```suggestion
void imcomm::setup(model *m, const std::string&) {
```

src/callbacks/learning_rate.cpp
```diff
@@ -55,7 +55,7 @@ learning_rate::learning_rate(
   std::vector<std::string> weights_names)
   : m_weights_names(std::move(weights_names)) {}
 
-void learning_rate::setup(model *m) {
+void learning_rate::setup(model *m, const std::string& trainer_name) {
```
benson31:
```suggestion
void learning_rate::setup(model *m, const std::string&) {
```

src/callbacks/ltfb.cpp
```diff
@@ -403,7 +403,7 @@ ltfb& ltfb::operator=(const ltfb& other) {
   return *this;
 }
 
-void ltfb::setup(model *m) {
+void ltfb::setup(model *m, const std::string& trainer_name) {
```
benson31:
```suggestion
void ltfb::setup(model *m, const std::string&) {
```

src/callbacks/perturb_adam.cpp
```diff
@@ -56,7 +56,7 @@ perturb_adam::perturb_adam(DataType learning_rate_factor,
     m_perturb_during_training(perturb_during_training),
     m_weights_names(std::move(weights_names)) {}
 
-void perturb_adam::setup(model* m) {
+void perturb_adam::setup(model* m, const std::string& trainer_name) {
```
benson31:
```suggestion
void perturb_adam::setup(model* m, const std::string&) {
```

src/callbacks/perturb_dropout.cpp
```diff
@@ -39,7 +39,7 @@ perturb_dropout::perturb_dropout(EvalType keep_prob_factor,
     m_keep_prob_factor(keep_prob_factor),
     m_layer_names(std::move(layer_names)) {}
 
-void perturb_dropout::setup(model* m) {
+void perturb_dropout::setup(model* m, const std::string& trainer_name) {
```
benson31:
```suggestion
void perturb_dropout::setup(model* m, const std::string&) {
```

src/callbacks/save_model.cpp
```diff
@@ -111,7 +111,7 @@ bool save_model::do_save_model_weights(model *m) {
 
   // Shared checkpoint, logic identical to Distributed.i
   makedir(m_dir.c_str());
-  std::string epochdir = get_shared_checkpoint_dirname(m, m_dir.c_str(), c.get_execution_mode(), epoch, step);
+  std::string epochdir = get_shared_checkpoint_dirname(c.get_trainer().get_name(), m, m_dir.c_str(), c.get_execution_mode(), epoch, step);
```
benson31: See. You got the trainer's name through a model pointer...

src/proto/factories/trainer_factory.cpp
```diff
@@ -41,6 +42,14 @@ std::unique_ptr<trainer> construct_trainer(lbann_comm* comm,
   if (!name.empty()) {
     t->set_name(name);
   }
+
+  // Construct callbacks
+  std::vector<std::unique_ptr<callback_base>> callback_list;
+  for (int i=0; i<proto_trainer.callback_size(); i++) {
+    callback_list.push_back(construct_callback(proto_trainer.callback(i)));
+  }
+  for (auto&& cb  : callback_list) { t->add_callback(std::move(cb)); }
```
benson31:
```suggestion
  // Construct callbacks
  for (int i=0; i<proto_trainer.callback_size(); i++) {
    t->add_callback(construct_callback(proto_trainer.callback(i)));
  }
```

src/trainers/trainer.cpp
```diff
@@ -70,6 +70,10 @@ trainer::trainer(const trainer& other) :
   // Deep copies
   // m_io_thread_pool = (other.m_io_thread_pool ?
   //                     other.m_io_thread_pool->copy() : nullptr);
+  m_callbacks = other.m_callbacks;
+  for (auto& cb : m_callbacks) {
+    cb.reset(cb->copy());
+  }
```
benson31:
```suggestion
  m_callbacks.reserve(other.m_callbacks.size());
  for (auto const& cb : other.m_callbacks) {
    m_callbacks.emplace_back(cb->copy());
  }
```

src/callbacks/print_statistics.cpp
```diff
@@ -42,7 +42,7 @@
 namespace lbann {
 namespace callback {
 
-void print_statistics::setup(model *m) {
+void print_statistics::setup(model *m, const std::string& trainer_name) {
```
benson31:
```suggestion
void print_statistics::setup(model *m, const std::string&) {
```

src/callbacks/replace_weights.cpp
```diff
@@ -38,7 +38,7 @@
 namespace lbann {
 namespace callback {
 
-void replace_weights::setup(model *m) {
+void replace_weights::setup(model *m, const std::string& trainer_name) {
```
benson31:
```suggestion
void replace_weights::setup(model *m, const std::string&) {
```

src/trainers/trainer.cpp
```diff
@@ -81,6 +85,10 @@ trainer& trainer::operator=(const trainer& other) {
...
```
benson31:
```suggestion
  m_callbacks.reserve(other.m_callbacks);
  for (auto const& cb : other.m_callbacks) {
    m_callbacks.emplace_back(cb->copy());
  }
```

src/trainers/trainer.cpp
```diff
@@ -93,6 +101,12 @@ trainer::~trainer() {
 ////////////////////////////////////////////////////////////
 
 void trainer::set_name(std::string const& name) {
+  if (name.empty()) {
+    std::ostringstream err;
+    err << "attempted to rename trainer \"" << get_name() << "\" "
+        << "with empty string";
+    LBANN_ERROR(err.str());
+  }
```
benson31:
```suggestion
  if (name.empty()) {
    LBANN_ERROR("attempted to rename trainer \"", get_name(), "\" with empty string");
  }
```

src/trainers/trainer.cpp
```diff
@@ -115,6 +129,11 @@ void trainer::setup(std::unique_ptr<thread_pool> io_thread_pool) {
   // Setup I/O threads - set up before setting up the layers (input
   // layer depends on having a properly initialized thread pool)
   m_io_thread_pool = std::move(io_thread_pool);
+
+  // Set up callbacks
+  for (const auto& cb : m_callbacks) {
+    cb->setup(this);
```
benson31: How does this compile? `callback_base::setup(trainer*)` is not a `const` member function.

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
...
+  ss << trainer_name.c_str() << "/";
+  return ss.str();
+}
+
+inline std::string get_last_shared_checkpoint_filename(model *m, const std::string& dir) {
+  std::ostringstream ss;
+  ss << dir;
+  ss << m->get_name().c_str();
+  ss << ".last.shared.checkpoint";
   return ss.str();
```
benson31:
```suggestion
  return build_string(dir, m->get_name(), ".last.shared.checkpoint");
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
   };
 };
 
-inline std::string get_last_shared_checkpoint_filename(model *m, const std::string& dir) {
-  lbann_comm *comm = m->get_comm();
+inline std::string get_trainer_checkpoint_dirname(const std::string& trainer_name, const std::string& dir) {
   std::ostringstream ss;
   ss << dir << "/";
```
benson31:
```suggestion
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
...
-  ss << m->get_name().c_str() << ".trainer.";
-  ss << comm->get_trainer_rank() << ".last.shared.checkpoint";
+  ss << trainer_name.c_str() << "/";
+  return ss.str();
```
benson31:
```suggestion
  return build_string(dir, '/', trainer_name, '/');
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
...
-  lbann_comm *comm = m->get_comm();
   std::ostringstream ss;
-  ss << dir << "/" << m->get_name().c_str();
-  ss << ".trainer." << comm->get_trainer_rank();
+  ss << dir;
+  ss << m->get_name().c_str();
   ss << ".shared." << to_string(mode);
   ss << ".epoch." << epoch;
   ss << ".step."<< step << "/";
   return ss.str();
```
benson31:
```suggestion
  return build_string(dir, m->get_name(), ".shared.", to_string(mode), ".epoch.", epoch, ".step.", step, '/');
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
...
 }
 
+inline std::string get_last_shared_checkpoint_filename(const std::string& trainer_name, model *m, const std::string& dir) {
+  return get_last_shared_checkpoint_filename(m, get_trainer_checkpoint_dirname(trainer_name, dir));
+}
+
 inline std::string get_shared_checkpoint_dirname(model *m, const std::string& dir, execution_mode mode, size_t epoch, size_t step) {
-  lbann_comm *comm = m->get_comm();
   std::ostringstream ss;
```
benson31:
```suggestion
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
...
 inline std::string get_last_distributed_checkpoint_filename(model *m, const std::string& dir) {
-  lbann_comm *comm = m->get_comm();
   std::ostringstream ss;
-  ss << dir << "/";
-  ss << m->get_name().c_str() << ".trainer.";
-  ss << comm->get_trainer_rank() << ".last.distributed.checkpoint";
+  ss << dir;
+  ss << m->get_name().c_str();
+  ss << ".last.distributed.checkpoint";
   return ss.str();
```
benson31:
```suggestion
  return build_string(dir, m->get_name(), ".last.distributed.checkpoint");
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
...
 }
 
+inline std::string get_shared_checkpoint_dirname(const std::string& trainer_name, model *m, const std::string& dir, execution_mode mode, size_t epoch, size_t step) {
+  return get_shared_checkpoint_dirname(m, get_trainer_checkpoint_dirname(trainer_name, dir), mode, epoch, step);
+}
+
 inline std::string get_last_distributed_checkpoint_filename(model *m, const std::string& dir) {
-  lbann_comm *comm = m->get_comm();
   std::ostringstream ss;
```
benson31:
```suggestion
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
...
   return ss.str();
 }
 
+inline std::string get_last_distributed_checkpoint_filename(const std::string& trainer_name, model *m, const std::string& dir) {
+  return get_last_distributed_checkpoint_filename(m, get_trainer_checkpoint_dirname(trainer_name, dir));
+}
+
 inline std::string get_distributed_checkpoint_dirname(model *m, const std::string& dir, execution_mode mode, size_t epoch, size_t step) {
   lbann_comm *comm = m->get_comm();
   std::ostringstream ss;
```
benson31:
```suggestion
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -151,47 +170,67 @@ class checkpoint : public callback_base {
...
-  ss << dir << "/" << m->get_name().c_str();
-  ss << ".trainer." << comm->get_trainer_rank();
+  ss << dir;
+  ss << m->get_name().c_str();
   ss << ".rank." << comm->get_rank_in_trainer();
   ss << ".distributed." << to_string(mode);
   ss << ".epoch." << epoch;
   ss << ".step."<< step << "/";
   return ss.str();
```
benson31:
```suggestion
  return build_string(dir,
    m->get_name(),
    ".rank.", comm->get_rank_in_trainer(),
    ".distributed.", to_string(mode),
    ".epoch.", epoch,
    ".step.", step, '/');
```

benson31: Do you need the "/" after `dir` still? (Applies to all of these changes.)

benson31:
Overall, I think the goal of this PR -- moving callbacks to the trainer -- is worthwhile. However, I think it exposes many of the frailties in the infrastructure of the code while doing little to nothing to improve them.

The main thing bothering me is the deficiency of the construction procedure. A trainer should be constructed first, since a model cannot survive without one. Thus it should always be available to its models from the moment they are constructed. This should also be true at restore-from-checkpoint time. (Ab)using strings is generally sloppy and error-prone, and confuses the programming interface ("why do I have to give it a name when I have the actual trainer object right here?"). The logic right now just seems a bit circular and fragile to maintain.

That said, this might be an acceptable step in the right direction around which we can refactor next. I understand that this is an intermediate step in the "data coordinator" work, and perhaps it's best to push that through and clean up once the whole picture is a bit more visible. Also, we have been talking about redesigning the callback interface; perhaps it's time we move on that.

In conclusion, I don't want to hold up this PR if the "bigger picture" is more important right now, but it would also make me uncomfortable to click "Approve" with these sorts of issues unaddressed.

include/lbann/callbacks/checkpoint.hpp
bvanessen: yes we do need the trailing slash.

include/lbann/callbacks/callback.hpp
bvanessen: Okay, I have fixed the issue with passing the trainer name around.  I am not sure what to do about the model because in this new world order a trainer has a single checkpoint callback, but it could be applied to multiple models.  Thus it is function with respect the model and stateful with respect to the trainer.

benson31: > Thus it is function with respect the model and stateful with respect to the trainer.

This is the case in which it makes sense to have a `model*` parameter (technically it should be a `model&` since null models are not allowed). So I'm ok with the `model*` in this case. It's the "model callbacks" that should not have it... But that's a longer discussion.

bamboo/unit_tests/test_unit_lbann2_reload.py
bvanessen: Hmmm, why not.  The problem is that there is a file with the glob pattern and a set of directories with that name.  Also we don't want to delete everything under trainer0, just the model1 specific checkpoints.

bvanessen:
@timmoon10 I have address @benson31 comments.  Please take a look.

timmoon10:
If we want models to own some callbacks and trainers to own others, we should create two separate classes (`trainer_callback` and `model_callback`). Putting them in the same class hierarchy doesn't give us any benefit since the trainers and models use completely disjoint APIs to interact with the callbacks.

include/lbann/trainers/trainer.hpp
```diff
@@ -151,6 +177,9 @@ class trainer {
   std::unordered_map<std::pair<observer_ptr<model>, execution_mode>,
                      std::unique_ptr<execution_context>,
                      model_execution_context_hash_t> m_model_execution_context;
+
+  /** @brief Current callbacks to process. */
+  std::vector<std::shared_ptr<callback_base>> m_callbacks;
```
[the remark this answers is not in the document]
bvanessen: For the time being with the way that the callbacks are structured, the checkpoint callback is owned by the trainer but needs to be executed by every model in the trainer.  Shared ownership does achieve this in a safe way and provides the smallest delta with respect to the current approach.

bvanessen:
While it may be possible to implement future callbacks in a split environment like this, the current system is not amenable to this approach of having disjoint trainer and model callbacks.  Additionally, we don't want to force all callbacks up to the trainer because some should be specific to a given model in the trainer.

include/lbann/trainers/trainer.hpp
```diff
@@ -77,6 +77,18 @@ class trainer {
   /** Human-readable description. */
   description get_description() const;
 
+  /** @brief Get the list of callbacks for the trainer. */
+  std::vector<std::shared_ptr<callback_base>>& get_callbacks() {
+    return m_callbacks;
+  }
```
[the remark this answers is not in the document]
benson31: I both agree and disagree. I don't like this because it exposes the ownership structure, but I disagree that it "gives ownership" to the caller -- ownership is tied to the reference count, which will not go up here. The tradeoff seems expensive to me for little benefit. In some sense, though, this will hopefully be moot after refactoring the callbacks.

include/lbann/models/model.hpp
```diff
@@ -446,7 +449,7 @@ class model {
   std::vector<metric*> m_metrics;
 
   /** @brief Current callbacks to process. */
-  std::vector<callback_base*> m_callbacks;
+  std::vector<std::shared_ptr<callback_base>> m_callbacks;
```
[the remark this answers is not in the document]
benson31: The first counterexample I can think of is: a trainer holding multiple models might assign different callbacks to each. In this case, the model can track its own callbacks or the trainer needs a more complicated mapping structure to assign callbacks to models. In the end, I don't know how much it matters, but this seems a reasonable approach for now.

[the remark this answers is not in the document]
timmoon10: Good point that it returns a reference, not a copy. In any case, I don't think we'd lose much performance by copying the `shared_ptr`s into raw pointers. This isn't happening in an inner loop, so the gains in clarity and safety are worth the cost of allocating some memory on the heap.

include/lbann/trainers/trainer.hpp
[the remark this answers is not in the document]
timmoon10: Good point that it returns a reference, not a copy. In any case, I don't think we'd lose much performance by copying the shared_ptrs into raw pointers. This isn't happening in an inner loop, so the gains in clarity and safety are worth the cost of allocating some memory on the heap.

include/lbann/models/model.hpp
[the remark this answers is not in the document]
timmoon10: I think the fundamental problem is we have two distinct objects, trainer callbacks and model callbacks, which are shoehorned into the same class. The mapping structure is inherent in the callbacks, so we either need to implement it in the class structure or in the callback implementations.

Thinking about it, many of the "model callbacks" are logically layer callbacks or weights callbacks. We implement the mapping structure internally so that they are applied to the right layer or weights objects.

include/lbann/models/model.hpp
```diff
@@ -136,7 +136,7 @@ class model {
   std::vector<weights*> get_weights();
 
   /** @brief Get the list of callbacks for the model. */
-  virtual std::vector<callback_base*>& get_callbacks() {
+  virtual std::vector<std::shared_ptr<callback_base>>& get_callbacks() {
     return m_callbacks;
   }
```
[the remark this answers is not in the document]
bvanessen: There is now a way to return both observer and shared ptr vectors.

PR #1458, opened 2020-02-27
fix the compilation error in debug mode

JaeseungYeom:
- The issue with accessing a member variable in a template class
   - "dropout.hpp:72:35: error: there are no arguments to 'get_comm' that depend on a template parameter, so a declaration of 'get_comm' must be available"
- name collision issue
  - "data_store_conduit.hpp:615:8: note: in expansion of macro 'DEBUG'"
  - "data_store_conduit.cpp:337:83: error: expression cannot be used as a function"

### DIFF
--- a/include/lbann/data_store/data_store_conduit.hpp
+++ b/include/lbann/data_store/data_store_conduit.hpp
@@ -612,7 +612,7 @@ private :
     flush_profile_file();
   }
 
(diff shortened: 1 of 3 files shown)


JaeseungYeom:
It compiles and runs for me now.

timmoon10:
:shipit:

PR #1459, opened 2020-02-27
Fix makedir

JaeseungYeom:
Replace `lbann::makedir()` with `make_directory_for_trainer()` such that only the trainer master creates a directory and synchronize with the rest processes in the trainer.
There are still more occupations to be dealt with in checkpoint callback. However, I am leaving it alone as it is undergoing refactoring.


JaeseungYeom:
> I haven't compiled/run, but looks OK. At some point we need to look at this more in depth, e.g, what about the case where we're creating a directory on a local disk -- in which case all local P_0 ranks should create the directory. But for now, this PR appears to solve an existing problem, so OK.

One example I can think of, if I understand your comment correctly, is the distributed checkpoint case, in which every rank creates a directory on its local SSD, which might still be shared among the local processors of the node. In which case, you can to use `make_directory()` with the unique directory name among the local processes.

davidHysom:
> > I haven't compiled/run, but looks OK. At some point we need to look at this more in depth, e.g, what about the case where we're creating a directory on a local disk -- in which case all local P_0 ranks should create the directory. But for now, this PR appears to solve an existing problem, so OK.
> 
> One example I can think of, if I understand your comment correctly, is the distributed checkpoint case, in which every rank creates a directory on its local SSD, which might still be shared among the local processors of the node. In which case, you can to use `make_directory()` with a unique directory name among the local processes.
> 
> I created a related issue:
> [#1460 (comment)](https://github.com/LLNL/lbann/issues/1460#issue-572422870)

Actually the case I had in mind is: a SINGLE rank on each node creates a directory on its local SSD, and only that rank writes to it (although there could be a case where all local ranks write to that directory. To make this concrete, the data_store has an owner_map, that is identical on all processors in the trainer. So, for checkpointing, there only needs to be a single local copy of the map. 

However, this is getting away from the immediate purpose of this PR -- just something to think about in the future (hm, I don't even know if we're currently testing checkpointing. I suspect we're not, at least where the data store is concerned)

JaeseungYeom:
I see. So, we have the cache level orthogonal to the scope hierarchy. 
We would need a method to choose/identify the cache master rank among those which are local to the particular storage level in the storage hierarchy (in this case the node-local SSD) for the scope.
I also think it is probably better to have the checkpoint directory hierarchy match the scope hierarchy.

JaeseungYeom:
> I don't like this approach. We should try to make our utility functions modular and generalizable, so making the file utilities require extra knowledge of trainers and MPI communication is a step in the wrong direction.
> 
> In any case, there shouldn't be a race condition in `lbann::file::make_directory`. It has internal machinery to return successfully if the directory already exists. If we are still worried, it's just a few lines to do:
> 
> ```c++
> if (comm->am_trainer_master()) {
>   lbann::file::make_directory(dir_name);
> }
> comm->trainer_barrier();
> ```

I do not like to have `comm.hpp` included in `file_utils` either. So, `lbann_comm` is forward-declared in the header without including `comm.hpp`.
We can find other location to put the function. Or even manually write the whole lines whenever necessary, but the lines wrapped in a function helps people avoid error.
Checking the existence of directory does not prevent race condition unless itself is guarded by a synchronization.

davidHysom:
> just a few lines

It's simpler to make it a one liner:

make_directory(std::string dir_name, int master = 0)
...
 lbann::file::make_directory(dir_name, comm->am_master());

bvanessen:
@JaeseungYeom I added the function that you are talking about to a new file that explicitly takes the comm object so it nicely separates from the file_utils.  Please take a look at PR #1523 and see if we can make the bugfixes here work on that new function call.

PR #1463, opened 2020-03-03
Add option to build with NVSHMEM

timmoon10:
This PR adds CMake options to build LBANN with NVSHMEM (`LBANN_WITH_NVSHMEM` and `NVSHMEM_DIR`). If you are on Lassen, you can run `build_lbann_lc.sh` has the `--nvshmem` flag. If LBANN is built with NVSHMEM, it will be initialized/finalized inside `lbann::initialize`/`lbann::finalize`.

[Bamboo is green, aside from an integration test failure that is also in `develop`](https://lc.llnl.gov/bamboo/browse/LBANN-TIM270-1).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,16 @@ if (LBANN_HAS_CUDA)
   else ()
     set(LBANN_HAS_NCCL2 FALSE)
   endif ()
+
+  if (LBANN_WITH_NVSHMEM)
+    find_package(NVSHMEM REQUIRED)
+    set_property(TARGET cuda::toolkit PROPERTY
+      INTERFACE_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:-arch=sm_70>)
+    # Build LBANN as a static library to get around a bug in NVSHMEM
+    set(BUILD_SHARED_LIBS OFF)
+  endif ()
+  set(LBANN_HAS_NVSHMEM "${NVSHMEM_FOUND}")
+
 endif (LBANN_HAS_CUDA)
 
 # This shouldn't be here, but is ok for now. This will occasionally be
@@ -568,6 +578,11 @@ endif ()
 
 if (LBANN_HAS_PYTHON)
   target_link_libraries(lbann PUBLIC Python::Python)
+endif ()
+
+if (LBANN_HAS_NVSHMEM)
+  set_property(TARGET lbann PROPERTY CUDA_SEPARABLE_COMPILATION ON)
+  target_link_libraries(lbann PUBLIC NVSHMEM::NVSHMEM)
 endif ()
 
 if (TARGET LBANN_CXX_FLAGS_werror)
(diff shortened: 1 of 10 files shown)


scripts/build_lbann_lc.sh
```diff
@@ -442,6 +449,16 @@ fi
 CXX_FLAGS="${CXX_FLAGS} -ldl"
 C_FLAGS="${CXX_FLAGS}"
 
+# Hacks to build with NVSHMEM
+if [ ${WITH_NVSHMEM} -ne 0 ]; then
+    if [ "${CLUSTER}" == "lassen" ]; then
+        NVSHMEM_DIR=/usr/workspace/wsb/brain/nvshmem/nvshmem_0.3.3/cuda-10.1_ppc64le
+        CUDA_FLAGS="-gencode=arch=compute_70,code=sm_70"
```
naoyam: Do you need this in addition to the `CMAKE_CUDA_FLAGS` setting in FindNVSHMEM?

CMakeLists.txt
```diff
@@ -314,6 +314,16 @@ if (LBANN_HAS_CUDA)
   else ()
     set(LBANN_HAS_NCCL2 FALSE)
   endif ()
+
+  if (LBANN_WITH_NVSHMEM)
+    find_package(NVSHMEM REQUIRED)
+    set_property(TARGET cuda::toolkit PROPERTY
+      INTERFACE_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:-arch=sm_70>)
```
naoyam: Do you need this in addition to the `CMAKE_CUDA_FLAGS` setting in FindNVSHMEM?

scripts/build_lbann_lc.sh
timmoon10: Inside `FindNVSHMEM.cmake` we set `-gencode=arch=compute_70,code=compute_70`, which generates future-compatible PTX. By also setting `-gencode=arch=compute_70,code=sm_70`, we make sure to compile a cubin specifically for Voltas. I think this is a bit more generalizable for post-Volta GPUs, although it might also be overkill.

src/utils/nvshmem.cu
```diff
@@ -0,0 +1,51 @@
...
+
+namespace lbann {
+namespace nvshmem {
+
+void initialize(MPI_Comm comm) {
+  nvshmemx_init_attr_t attr;
+  attr.mpi_comm = &comm;
+  auto status = nvshmemx_init_attr(NVSHMEMX_INIT_WITH_MPI_COMM, &attr);
+  if (status != 0) {
+    nvshmem_finalize();
```
naoyam: This is a minor thing, but finalizing NVSHMEM when its initialization is failed doesn't seem right to me.

timmoon10: Fixed.

CMakeLists.txt
timmoon10: I put this to suppress compilation with the default architecture, which I believe is sm_30. This was also to override when Hydrogen manually sets the architecture to sm_30.

naoyam:
This looks good to me.

PR #1469, opened 2020-03-09
Add ParallelStrategy to Layer.

naoyam:
Describes how layer input and output tensors should be partitioned. It
is currently just ignored.


naoyam:
Thanks. I created an issue to track refactoring ideas.

#1470

naoyam:
Bamboo is green.

PR #1472, opened 2020-03-10
Optionally build DiHydrogen (legacy) with Superbuild

naoyam:
Just builds and links DiHydrogen and its legacy distconv. Distconv is not yet touched by LBANN.


naoyam:
@benson31 I'd like to have your help here. I'm trying to build DiHydrogen with superbuild. This seems to work when just building DiHydrogen with Distconv. However, with Distconv, the LBANN CMakeLists.txt results in errors at `add_executable` and `add_library` for those using the lbann target. 

Specifically, this works:
```
./scripts/build_lbann_lc.sh --with-dihydrogen
```

This builds DiHydrogen with superbuild. However, this fails:

```
./scripts/build_lbann_lc.sh --with-dihydrogen --with-distconv
```
Here's the first several lines from `LBANN-configure-err.log`:
```
Make Warning at CMakeLists.txt:109 (message):
  DiHydrogen support is currently expermimental.  There is no stable
  interface.  Use caution before using any features.


CMake Warning at CMakeLists.txt:116 (message):
  Distconv support is currently expermimental.  There is no stable interface.
  Use caution before using any features.


CMake Error at unit_test/CMakeLists.txt:5 (add_executable):
  Target "seq-catch-tests" links to target "cuda::CUB" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

My guess is that it has something to do with that Distconv uses CUB and cuDNN, etc. Can you please take a look at this?

benson31:
@naoyam I'm on it!

naoyam:
This works with and without distconv, thanks to @benson31.

naoyam:
Bamboo is fine.

PR #1473, opened 2020-03-11
Added DiHydrogen as a dependency for building LBANN.  Also bumped up

bvanessen:
the version of CMake and CUB.

### DIFF
--- a/spack_environments/llnl_lc/externals-linux-rhel7-power9le.sh
+++ b/spack_environments/llnl_lc/externals-linux-rhel7-power9le.sh
@@ -17,9 +17,9 @@ EXTERNAL_PACKAGES=$(cat <<EOF
     cmake::
       buildable: True
       variants: ~openssl ~ncurses
-      version: [3.12.1]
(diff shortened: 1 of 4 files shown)


spack_environments/llnl_lc/externals-linux-rhel7-power9le.sh
```diff
@@ -17,9 +17,9 @@ EXTERNAL_PACKAGES=$(cat <<EOF
     cmake::
       buildable: True
       variants: ~openssl ~ncurses
-      version: [3.12.1]
+      version: [3.13.4]
       paths:
-        cmake@3.12.1 arch=linux-rhel7-power9le:   /usr/tce/packages/cmake/cmake-3.12.1
+        cmake@3.13.4 arch=linux-rhel7-power9le:   /usr/tce/packages/cmake/cmake-3.13.4
```
benson31: Can we bump all the way to 3.14.5? That's the current *default* on LC (i.e., what you get if you `ml cmake`).

Issue #1448, opened 2020-02-20
SIGSEGV in running resnet50 with imagenet data


sheltongeosx:
Dear all,

I am running lbann using 1 GPU to test resnet50 with imagenet data on IBM power8 machine and caught signal 11 for invalid memory reference. The following are the process info and error messages:
   
       
     LBANN version: commit ca785ec, 02/24/2020
     lbann runtime info:
          protobuf_utils::verify_prototext; starting verify for 1 models
	         Num. I/O Threads: 1 (Limited to # Unused Compute Cores or 1)
          Hardware properties (for master process)
             Processes on node          : 1
             Total number of processes  : 1
             OpenMP threads per process : 160
             I/O threads per process (+offset) : 1 (+1)
             GPUs on node               : 4

          Build settings
              Type     : Release
              Aluminum : detected
              CUDA     : detected
              cuDNN    : detected
              CUB      : detected
              MV2_USE_CUDA : 

          Aluminum Features:
              NCCL : enabled

         Trainer settings
               Trainers              : 1
               Processes per trainer : 1
               Grid dimensions       : 1 x 1
         trainer0
               Background I/O: true

         writing options and prototext to file: model_8.prototext

         imagenet_reader is set
         imagenet_reader is set
         Training using 544546 samples.
         Validating using 50000 samples.
         Testing using 0 samples.

         Running with these parameters:
              General:
              datatype size:           4
              mini_batch_size:         32
              num_epochs:              10
              hydrogen_block_size:     0
              procs_per_trainer:       0
              num_parallel_readers:    0
              serialize_io:            0
              cuda:                    enabled
              cudnn:                   enabled
              random_seed:             0
              data_layout:             data_parallel
              (only used for metrics)
          Training with LLNL LBANN version 0.99.0
          --------------------------------------------------------------------------------
          [0] Epoch : stats formated [tr/v/te] iter/epoch = [17018/1563/0]
                       global MB = [  32/  32/   0] global last MB = [   2  /  16  /   0  ]
                       ocal MB = [  32/  32/   0]  local last MB = [   2+0/  16+0/   0+0]
           --------------------------------------------------------------------------------

    Error messages:
        Caught signal 11 (SIGSEGV - invalid memory reference) on rank 0
        Stack trace:
               0: lbann::stack_trace::get[abi:cxx11]()
               1: lbann::exception::exception(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool)
[...]

PR #1478, opened 2020-03-17
Check labels in image data reader

timmoon10:
Issue #1448 was caused because the ImageNet data reader expected samples with 1000 labels, but got a label outside the valid range. This PR adds a check so that this error results in an informative error message rather than a segfault.

[Bamboo build](https://lc.llnl.gov/bamboo/browse/LBANN-TIM274-1).

### DIFF
--- a/src/data_readers/data_reader_image.cpp
+++ b/src/data_readers/data_reader_image.cpp
@@ -119,15 +119,13 @@ void image_data_reader::set_input_params(const int width, const int height, cons
 
 bool image_data_reader::fetch_label(CPUMat& Y, int data_id, int mb_idx) {
   const label_t label = m_image_list[data_id].second;
-  if (label >= 0 && label < m_num_labels) {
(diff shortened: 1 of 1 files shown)


src/data_readers/data_reader_image.cpp
```diff
@@ -119,7 +119,15 @@ void image_data_reader::set_input_params(const int width, const int height, cons
 
 bool image_data_reader::fetch_label(CPUMat& Y, int data_id, int mb_idx) {
   const label_t label = m_image_list[data_id].second;
-  Y.Set(label, mb_idx, 1);
+  if (label >= 0 && label < m_num_labels) {
```
JaeseungYeom: I would use static_cast<label_t>(0) and static_cast<label_T>(m_num_labels) just in case.

timmoon10: Done

PR #1479, opened 2020-03-18
Add a warning to the documentation for IBV_FORK_SAFE

benson31:
Since I'm _sure_ all our users meticulously comb through our documentation before punching more-or-less arbitrary strings into their terminal, I figured I'd at least give us some future protection, however theoretical, from encountering issues related to `IBV_FORK_SAFE` and the Python data reader.

### DIFF
--- a/docs/running_lbann.rst
+++ b/docs/running_lbann.rst
@@ -209,7 +209,7 @@ for a list of available preprocessing transforms.
              that doesn't interact with Infiniband all that well by
              default. Users may encounter hangs on clusters that use
              Infiniband. To avoid this, ensure that
-             :code:`IBV_FORK_SAFE=1` is exported into the environment
+             :bash:`IBV_FORK_SAFE=1` is exported into the environment
              when running LBANN.
 
 ------------------------------------------------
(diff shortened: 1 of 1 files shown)


docs/running_lbann.rst
```diff
@@ -205,6 +205,13 @@ consistent and some preprocessing is typically required. See
 <https://github.com/LLNL/lbann/blob/develop/src/proto/transforms.proto>`_
 for a list of available preprocessing transforms.
 
+.. warning:: The python data reader will trigger some process forking
+             that doesn't interact with Infiniband all that well by
+             default. Users may encounter hangs on clusters that use
+             Infiniband. To avoid this, ensure that
+             :bash:`IBV_FORK_SAFE=1` is exported into the environment
+             when running LBANN.
```
timmoon10:
```suggestion
.. warning:: The Python data reader will trigger some process forking
             that doesn't interact with InfiniBand all that well by
             default. Users may encounter hangs on clusters that use
             InfiniBand. To avoid this, ensure that
             :bash:`IBV_FORK_SAFE=1` is exported into the environment
             when running LBANN.
```

PR #1480, opened 2020-03-19
Bugfix checkpoint trainer

bvanessen:
this PR started off by working on hoisting the RNG state and training algorithm state out of the model's portion of the checkpoint and bringing it up to a training algorithm level of the checkpoint hierarchy.  This PR now structures the checkpoint as <trainer>/<execution_context_and_epoch_step>/<model>  This allows the common state to be separated from each model.

Additionally, it deprecates the lbann2 front end and creates two new tests for checkpoints and the load/save model callbacks.


include/lbann/callbacks/callback.hpp
```diff
@@ -32,7 +32,6 @@
 #include "lbann/trainers/trainer.hpp"
 #include "lbann/layers/layer.hpp"
 #include "lbann/models/model.hpp"
-#include "lbann/trainers/trainer.hpp"
```
benson31: Don't remove this line. `trainer` is used in the API. Include what you use. Or forward-declare it -- not a fight I'm really interested in picking. But the fight I am willing to pick is that you shouldn't use a symbol without somehow explicitly declaring it via either by forward declaration or by `#include`-ing the interface.

include/lbann/callbacks/load_model.hpp
```diff
@@ -63,12 +63,18 @@ class load_model : public callback_base {
...
+    m_dirs.emplace_back(dir);
+  }
+
   void on_train_begin(model *m) override;
   /* ckptdir_is_fullpath flag if true
  * allow user to specify full path to model weights to load
  * and allow system to ignore appending trainer id, num of epochs/steps
  * to default ckpt_dir*/
   static bool load_model_weights(std::string ckpt_dir,
+                                 std::string alg_name,
```
benson31:
```suggestion
                                 const std::string& alg_name,
```
Is there a reason `const string&` isn't sufficient?

include/lbann/execution_contexts/execution_context.hpp
```diff
@@ -105,6 +107,21 @@ class execution_context {
     return const_cast<trainer&>(static_cast<const execution_context&>(*this).get_trainer());
   }
 
+  inline void set_training_algorithm(training_algorithm* t){
```
benson31: Should you be able to explicitly set the training algorithm to `nullptr`? If not, this should take a reference.

include/lbann/io/persist.hpp
```diff
@@ -183,6 +183,8 @@ class persist {
   template <typename TensorDataType>
   bool read_datatype (persist_type type, const char *name, TensorDataType *val);
 
+  const std::string get_checkpoint_dir() const { return m_checkpoint_dir; }
```
benson31:
```suggestion
  const std::string& get_checkpoint_dir() const { return m_checkpoint_dir; }
```

include/lbann/models/model.hpp
```diff
@@ -428,6 +428,11 @@ class model {
    */
   std::string m_name;
 
+  /** @brief Is the model setup
+   *  @details Flag to indicate if the setup function has been called
+   */
+  bool m_model_setup;
```
benson31: Put this somewhere else. Order members by increasing or decreasing size for optimal alignment potential.

include/lbann/trainers/trainer.hpp
```diff
@@ -109,7 +109,7 @@ class trainer {
                                     execution_mode mode);
 
   execution_context_key_pair_t
-  check_and_build_execution_context(const execution_context& c,
+  check_and_build_execution_context(/*const */execution_context& c,
```
benson31:
```suggestion
  check_and_build_execution_context(execution_context& c,
```

src/trainers/trainer.cpp
```diff
@@ -144,26 +144,27 @@ trainer::execution_context_key_pair_t trainer::check_and_build_execution_context
...
                                                                                  model& model,
                                                                                  execution_mode mode) {
   auto key = std::make_pair(&model, mode);
   if(m_model_execution_context.count(key) == 0) {
     std::unique_ptr<execution_context> context;
-    if(dynamic_cast<observer_ptr<const sgd_execution_context>>(&c) != nullptr) {
-      context = make_unique<sgd_execution_context>(this, m_comm, mode, model.get_max_mini_batch_size());
+    //    observer_ptr<training_algorithm> alg = const_cast
+    if(dynamic_cast<observer_ptr</*const */sgd_execution_context>>(&c) != nullptr) {
+      context = make_unique<sgd_execution_context>(this, &(c.get_training_algorithm()), m_comm, mode, model.get_max_mini_batch_size());
```
benson31: This shouldn't work with pointers -- prefer references unless you can meaningfully pass `nullptr`. The address-of operator here is clunky at best.

src/training_algorithms/training_algorithm.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "lbann/training_algorithms/training_algorithm.hpp"
+#include "lbann/models/model.hpp"
+#include "lbann/callbacks/callback.hpp"
+#include "lbann/callbacks/checkpoint.hpp"
+#include "lbann/callbacks/save_model.hpp"
+#include "lbann/callbacks/load_model.hpp"
+
+namespace lbann {
+
+void training_algorithm::setup_models(model& model) {
```
timmoon10: The code path here is convoluted and brittle. What happens if you call `model::setup` outside this function? If `model::setup` really needs a training algorithm, then it should require it as an input argument.

include/lbann/callbacks/callback.hpp
bvanessen: it is a duplicate of line 32.

include/lbann/training_algorithms/training_algorithm.hpp
```diff
@@ -55,10 +55,15 @@ class training_algorithm {
...
   //  virtual training_algorithm* copy() const = default;
 
+  virtual std::string get_name() const = 0;
+
   virtual void apply(execution_context& context,
                      model& model,
                      execution_mode mode,
                      termination_criteria const& term_criteria) = 0;
+
+  void setup_models(model& model);
```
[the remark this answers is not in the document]
bvanessen: Well, it really should be able to setup all of the models in a trainer.  I will update the code to take a vector.

src/training_algorithms/training_algorithm.cpp
bvanessen: @benson31 Is this a concern for you?

bvanessen:
@timmoon10 @benson31 @samadejacobs The PR is now really ready for review. Please take a look at it.

bvanessen:
Bamboo is green on catalyst and pascal.  Rerunning corona and ray.

Issue #1474, opened 2020-03-11
weight freezing might have been broken


JaeseungYeom:
I am trying to run `lbann/model_zoo/models/jag/wae_cycle_gan/cycle_gan_only.prototext`, and getting error
`layer and weights of them are inconsistently frozen`

JaeseungYeom:
PR #1482 fixes the issue.

PR #1482, opened 2020-03-22
Fix layer freezing

JaeseungYeom:
- This addresses the issue of inconsistently frozen layer and its weight (reported in the issue #1474).
This occurs during `setup_data()` of the recently refactored fully connected layers.
Basically, the issue is the misuse of `is_frozen()` interface which checks if the layer and its weights are consistently frozen/unfrozen. This has been used during the setup to freeze newly created weights of a frozen layer (i.e., before they are frozen). 
- Another issue that this PR addresses is checking weight pointer for _nullptr_ in `replace_weights` in `data_type_layer`.
- This passed the bamboo tests on _pascal_. However, it fails with `test_compiler_build_script` on _ray_ although I can build/run it myself on _ray_ and _lassen_.

### DIFF
--- a/src/layers/data_type_layer.cpp
+++ b/src/layers/data_type_layer.cpp
@@ -492,6 +492,9 @@ void data_type_layer<TensorDataType>::replace_weights(Layer* other_layer) {
   const std::vector<WeightsType*>& other_layer_weights =
     dynamic_cast<data_type_layer<TensorDataType>*>(other_layer)->get_data_type_weights();
   for (size_t i = 0; i < m_weights.size(); ++i) {
+    if (m_weights[i] == nullptr) {
+      continue;
+    }
     m_weights[i]->set_values(other_layer_weights[i]->get_values());
   }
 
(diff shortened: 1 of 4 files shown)


src/layers/data_type_layer.cpp
```diff
@@ -492,6 +492,9 @@ void data_type_layer<TensorDataType>::replace_weights(Layer* other_layer) {
   const std::vector<WeightsType*>& other_layer_weights =
     dynamic_cast<data_type_layer<TensorDataType>*>(other_layer)->get_data_type_weights();
   for (size_t i = 0; i < m_weights.size(); ++i) {
+    if (m_weights[i] == nullptr) {
+      continue;
+    }
     m_weights[i]->set_values(other_layer_weights[i]->get_values());
```
benson31:
```suggestion
    if (m_weights[i]) {
      m_weights[i]->set_values(other_layer_weights[i]->get_values());
    }
```

PR #1483, opened 2020-03-24
Python frontend launcher will launch interactive job by default

timmoon10:
The current launcher attempts to detect whether it is being run from within an allocation. If it is, it submits an interactive job (`srun`/`jsrun`), and otherwise submits a batch job (`sbatch`/`bsub`). This is convenient, but it causes errors when LBANN is integrated within a larger workflow, e.g. with Merlin. We need to give the user control over how their jobs are run.

This PR changes the default behavior to always launch an interactive job. If a batch job is required, the user can pass `batch_job=True` to `lbann.run` or `lbann.contrib.launcher.run`. Note that `jsrun` must be called from a compute node, so calling `lbann.run` on an LSF system without `batch_job=True` will fail. These changes supersede #1477 (close #1477).

Since we are prioritizing the case where LBANN is run within an allocation, I've taken the opportunity to remove the LC-specific defaults for scheduler partitions and accounts. LC systems already have default partitions and it's a bad idea for all our users to bill their CPU-hours to our accounts.

[Bamboo build](https://lc.llnl.gov/bamboo/browse/LBANN-TIM275-1).

### DIFF
--- a/python/lbann/launcher/lsf.py
+++ b/python/lbann/launcher/lsf.py
@@ -53,47 +53,36 @@ class LSFBatchScript(BatchScript):
                          interpreter=interpreter)
         self.nodes = nodes
         self.procs_per_node = procs_per_node
+        self.reservation = reservation
         self.launcher = launcher
         self.launcher_args = launcher_args
 
         # Configure header with LSF job options
-        self._construct_header(job_name=job_name,
-                               nodes=self.nodes,
-                               time_limit=time_limit,
-                               partition=partition,
-                               account=account,
-                               reservation=reservation)
-
-    def _construct_header(self,
-                          job_name=None,
-                          nodes=1,
-                          time_limit=None,
-                          partition=None,
-                          account=None,
-                          reservation=None):
-        """Construct script header with options for bsub."""
-        if job_name:
-            self.add_header_line('#BSUB -J {}'.format(job_name))
-        if partition:
-            self.add_header_line('#BSUB -q {}'.format(partition))
-        self.add_header_line('#BSUB -nnodes {}'.format(nodes))
+        self.add_header_line(f'#BSUB -cwd {self.work_dir}')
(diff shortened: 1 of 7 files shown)


python/lbann/launcher/lsf.py
```diff
@@ -53,47 +53,36 @@ def __init__(self,
...
-
-    def _construct_header(self,
-                          job_name=None,
-                          nodes=1,
-                          time_limit=None,
-                          partition=None,
-                          account=None,
-                          reservation=None):
-        """Construct script header with options for bsub."""
+        self.add_header_line(f'#BSUB -cwd {self.work_dir}')
```
samadejacobs: Header lines are only needed if submitting a batch job not in interactive (default) mode right? I don't see where add_header  is optional for both LSF and SLURM

python/lbann/launcher/slurm.py
```diff
@@ -50,47 +58,39 @@ def __init__(self,
...
-            self.add_header_line('#SBATCH --time={}-{:02d}:{:02d}:{:02d}'
-                                 .format(days, hours, minutes, seconds))
-        self.add_header_line('#SBATCH --chdir={}'.format(self.work_dir))
-        self.add_header_line('#SBATCH --output={}'.format(self.out_log_file))
-        self.add_header_line('#SBATCH --error={}'.format(self.err_log_file))
-        if partition:
-            self.add_header_line('#SBATCH --partition={}'.format(partition))
-        if account:
-            self.add_header_line('#SBATCH --account={}'.format(account))
+        self.add_header_line(f'#SBATCH --chdir={self.work_dir}')
```
samadejacobs: Same as previous comment.

python/lbann/launcher/lsf.py
samadejacobs: Actually, I see this is only when batch scripts are constructed, so ignore.

PR #1485, opened 2020-03-29
Pytest data store redux

davidHysom:
Unit tests for all remaining data_reader/data_store combinations 
1) lbann/core/model.py - modified to return the protobuf object so users can set fields that are not explicitly paramatized in the constructor; this is supported by the added method: get_protobuf_model()
2) contrib/lc/paths.py - added method: jag_dir(), to support unit tests with JAG data
3) core/objective_function.py - added 'else : raise RuntimeError' clause to export_proto() method; without this, users can pass erroneous objects to the ctor, which will be silently ignored, possibly resulting in an empty objective_function
4) bamboo/unit_tests/test_unit_datastore_jag.py - working version; passes bamboo on pascal


davidHysom:
overtaken by history; no longer relevant.

PR #1486, opened 2020-03-30
callback to freeze/unfreeze layers

JaeseungYeom:
callback to freeze/unfreeze layers
`freezer` callback that has the following four parameters: 
- freeze_by_epoch
- unfreeze_by_epoch
- freeze_by_step
- unfreeze_by_step
Each of these is a list of the pair `[epoch layer_name]` or `[step layer_name]`


timmoon10 (changes requested):
I think this has gotten stale. We can dig it up if we need this functionality in the future.

PR #1489, opened 2020-04-02
Enable Distconv in convolutions

naoyam:
This is a follow-up to a prior PR #1488 and adds support for convolution and deconvolution. Deconvolution only works for very limited cases at this point, but it should correctly fall back to the base LBANN implementation when necessary.

Bamboo test: https://lc.llnl.gov/bamboo/browse/LBANN-NAOYAM5-2

### DIFF
--- a/include/lbann/layers/learning/convolution.hpp
+++ b/include/lbann/layers/learning/convolution.hpp
@@ -159,8 +159,8 @@ protected:
     if(this->using_gpus()) {
 #ifdef LBANN_HAS_DISTCONV
       if (this->distconv_enabled()) {
-        this->dc().fp_compute_convolution();
-        this->dc().fp_apply_bias();
+        this->get_distconv_adapter().fp_compute_convolution();
+        this->get_distconv_adapter().fp_apply_bias();
         return;
       }
 #endif // LBANN_HAS_DISTCONV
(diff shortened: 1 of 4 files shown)


include/lbann/layers/learning/convolution.hpp
```diff
@@ -159,8 +193,134 @@ class convolution_layer : public base_convolution_layer<TensorDataType, Device>
...
+#ifdef LBANN_HAS_DISTCONV
+template <typename TensorDataType, data_layout T_layout, El::Device Dev>
+void convolution_distconv_adapter<TensorDataType, T_layout, Dev>::
+setup_distributions(tensor_overlap_constraints &constraints) {
+  base_convolution_adapter<TensorDataType, Dev>::setup_distributions(
+      constraints);
+  auto &l = dynamic_cast<convolution_layer<
+    TensorDataType, T_layout, Dev>&>(this->layer());
+  auto kernel_dims = l.get_kernel_dims();
+  std::reverse(kernel_dims.begin(), kernel_dims.end());
```
timmoon10, line 233: For clarification, are distconv tensor dimensions in small-stride-to-large-stride order? So data tensors are WHCN and conv kernels are WHCK?

include/lbann/layers/learning/convolution.hpp
```diff
@@ -151,6 +174,17 @@ class convolution_layer : public base_convolution_layer<TensorDataType, Device>
 
   void bp_compute() override {
     if(this->using_gpus()) {
+#ifdef LBANN_HAS_DISTCONV
+      if (this->distconv_enabled()) {
+        if (this->get_distconv_adapter().m_conv->is_overlap_bwd_halo_exchange_enabled()) {
+          this->get_distconv_adapter().m_conv->backward_data_exchange_halo(
+              this->get_distconv_adapter().get_prev_error_signals());
```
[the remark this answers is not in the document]
naoyam: That's for overlapping halo exchange. Backprop data needs to do halo exchange, which can be run in parallel with backprop filter, so we start the halo exchange, and then call backprop filter and data. It's not very obvious from the code, but the concurrent operations are managed in the distcovn conv implementation. In forward prop, overlapping is done within `m_conv->forward`, so nothing is needed at the LBANN side. 

The reason of the implementation difference between the fp and bp overlapped halo exchange is that the bp overlapping method is preferable as it's simpler but not applicable to the fp case.

include/lbann/layers/learning/convolution.hpp
naoyam: That is correct. It's column major.

PR #1491, opened 2020-04-04
Revising python front end (PFE) for improved API and documentation

davidHysom:
optimizer.py is close to final form, but not yet tested.


python/lbann/core/optimizer.py
```diff
@@ -2,18 +2,158 @@
...
+    self._momentum = momentum
+    self._nesterov = nesterov
+
+  def export_proto(self) :
+    '''Construct and return a protobuf message.'''
+    proto = metrics_pb2.Optimizer.SGD()
+    proto.learn_rate = self._learn_rate
+    proto.momentum = self._momentum
+    proto.nesterov = self._nesterov
+    return proto
```
timmoon10, line 157: To match the behavior of the current implementation, how about:
```suggestion
    base_message = optimizer_pb2.Optimizer()
    derived_message = base_message.sgd
    derived_message.learn_rate = self._learn_rate
    derived_message.momentum = self._momentum
    derived_message.nesterov = self._nesterov
    derived_message.SetInParent()
    return base_message
```

Issue #1495, opened 2020-04-06
Cleaning up legacy prototext models


bvanessen:
autoencoder_candle_pilot1
- [ ] model_autoencoder_chem_ecfp.prototext
- [x] model_autoencoder_chem_ecfp_200x150x100x100x100.prototext
- [x] model_autoencoder_chem_ecfp_500x250x100.prototext
- [x] model_autoencoder_chem_sigmoid.prototext
- [ ] model_dnn_chem_ecfp.prototext

autoencoder_cifar10
- [x] model_autoencoder_cifar10.prototext
- [x] model_conv_autoencoder_cifar10.prototext

autoencoder_imagenet
- [x] model_conv_autoencoder_imagenet.prototext

autoencoder_mnist
- [x] model_autoencoder_mnist.prototext
- [x] model_conv_autoencoder_mnist.prototext
- [ ] vae_mnist.prototext

candle/pilot1
- [ ] ae_nodeselect_gdc.prototext
- [ ] combo.prototext

cosmoflow
- [x] ae_nodeselect_gdc.prototext

gan
- [x] jags/cycle_gan
- [ ] mnist

jag
- [x] ae_cycle_gan
- [x] gan
- [x] wae_cycle_gan
- [x] vae_fcn.prototext
- [x] wae.prototext

molecular_autoencoder_candle_pilot2
- [ ] model_conv_molecular_autoencoder_pilot2.prototext
- [ ] model_conv_molecular_bead_autoencoder_pilot2.prototext
- [ ] model_molecular_autoencoder_pilot2.prototext

siamese
- [x] finetune-cub
- [x] triplet

simple_mnist
- [x] model_mnist_simple_1.prototext
- [x] model_mnist_simple_2.prototext

davidHysom:
Simple_minst models should go away; Katie previously notified

bvanessen:
Closed by #1526 .

PR #1497, opened 2020-04-06
Add self-supervised learning scripts to application directory

timmoon10:
This PR moves some scripts for self-supervised learning from [Bitbucket](https://mybitbucket.llnl.gov/users/moon13/repos/lbann_siamese) to the application directory. I've also removed deprecated prototext models (see #1495) and data readers.

Pretraining runs fine, both with Siamese networks and supervised learning. However, finetuning fails because the images in the CUB dataset have been removed from the Lustre brainusr directory.

[Bamboo build](https://lc.llnl.gov/bamboo/browse/LBANN-TIM276-1).


applications/selfsupervised/main.py
```diff
@@ -0,0 +1,140 @@
...
+# Initialize LBANN executable and command-line arguments
+lbann_exe = os.path.realpath(lbann.lbann_exe())
+lbann_exe = os.path.join(os.path.dirname(lbann_exe), 'lbann2')
+lbann_command = [lbann_exe]
+
+# Construct experiment directory
+experiment_dir = util.make_experiment_dir(args.job_name)
+
+# Export model prototext files
+# Note: lbann2 driver doesn't have a command-line argument to get
```
bvanessen, line 101: Does this model require lbann2 as a front end?

timmoon10: Yep. It first trains with the Siamese network on ImageNet, and then transfers the weights to a standard AlexNet and trains on CUB.

timmoon10:
- This implementation doesn't freeze weights during fine-tuning.
- I tinkered with some of the hyperparameters (especially to increase the mini-batch size), so it's not quite the same as Nathan's experiments.
- In practice, I don't think we're going to need the CUB dataset anymore, so that's fine. At this point these scripts are more important as guides for future projects.
- I copied my Bitbucket repo to the application directory, with some minor clean up to remove hard-coded paths.

PR #1498, opened 2020-04-07
Enable Distconv in transform layers

naoyam:
Supported layers include `split`, `sum`, `pooling` and `concatenate`. 

Depends on #1488.

### DIFF
--- a/include/lbann/layers/transform/concatenate.hpp
+++ b/include/lbann/layers/transform/concatenate.hpp
@@ -29,11 +29,27 @@
 
 #include "lbann/layers/data_type_layer.hpp"
 #include "lbann/utils/exception.hpp"
+#include "lbann/utils/distconv.hpp"
 
 #include <lbann/proto/proto_common.hpp>
 #include <layers.pb.h>
 
 namespace lbann {
+
+#ifdef LBANN_HAS_DISTCONV
+template <typename TensorDataType, data_layout Layout,
+          El::Device Device>
+class concatenate_distconv_adapter : public data_type_distconv_adapter<TensorDataType> {
+ public:
+  using TensorDevType = typename data_type_distconv_adapter<TensorDataType>::TensorDevType;
+  concatenate_distconv_adapter(Layer& layer):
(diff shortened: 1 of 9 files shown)


include/lbann/layers/transform/split.hpp
```diff
@@ -80,8 +101,64 @@ class split_layer : public transform_layer<TensorDataType> {
...
+setup_distributions(tensor_overlap_constraints &constraints) {
+  data_type_distconv_adapter<TensorDataType>::setup_distributions(
+      constraints);
+
+  auto &x = this->get_prev_activations_dist();
+  auto &y = this->get_activations_dist();
+  auto &dx = this->get_error_signals_dist();
+  auto &dy = this->get_prev_error_signals_dist();
+
+  constraints.mark_equivalent(x, y);
```
timmoon10, line 144: For clarification, do all outputs share the same distribution? Is that why we only mark the constraint once?

include/lbann/layers/transform/concatenate.hpp
```diff
@@ -289,6 +344,50 @@ void concatenate_layer<TensorDataType,Layout,Device>::bp_compute() {
...
+  return dynamic_cast<const concatenate_distconv_adapter<TensorDataType, T_layout, Dev>&>(
+      data_type_layer<TensorDataType>::get_distconv_adapter());
+}
+
+template <typename TensorDataType, data_layout Layout, El::Device Device>
+dc::Shape concatenate_distconv_adapter<TensorDataType, Layout, Device>::
+get_activations_local_shape(int index) const {
+  assert_eq(index, 0);
+  auto shape = this->get_prev_activations().get_local_shape();
+  shape[-2] = this->get_activations_shape()[-2];
```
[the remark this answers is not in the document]
naoyam: Addressed 71eba661958ee3af464f099017920ad9ba9e0346.

include/lbann/layers/transform/split.hpp
naoyam: That is correct and is one of the current limitations.

Issue #1470, opened 2020-03-09
Distconv refactoring notes


naoyam:
* Move `ParallelStrategy` into its own header file. (#1469, @ndryden)
* Have `ParallelStrategy` convert between groups and splits as needed (#1469, @ndryden)

naoyam:
https://github.com/LLNL/lbann/pull/1488#pullrequestreview-387626109

PR #1499, opened 2020-04-07
Enable Distconv in activations layers

naoyam:
ReLU, leaky ReLU and softmax are supported.

Builds on #1488.


naoyam:
This PR also just adds a call to the corresponding distconv implementation, except for the ReLU layer, which needed to be expanded like the leaky ReLU layer because distconv doesn't have a CPU implementation of ReLU.

bamboo/unit_tests/test_unit_layer_leaky_relu_distconv.py
```diff
@@ -0,0 +1,206 @@
...
+        vals.append(z)
+    val = np.mean(vals)
+    tol = 8 * val * np.finfo(np.float32).eps
+    callbacks.append(lbann.CallbackCheckMetric(
+        metric=metrics[-1].name,
+        lower_bound=val-tol,
+        upper_bound=val+tol,
+        error_on_failure=True,
+        execution_modes='test'))
+
```
timmoon10, line 144: This part of the test doesn't involve distconv and can be deleted.
```suggestion
```

bamboo/unit_tests/test_unit_layer_relu_distconv.py
```diff
@@ -0,0 +1,204 @@
...
+        vals.append(z)
+    val = np.mean(vals)
+    tol = 8 * val * np.finfo(np.float32).eps
+    callbacks.append(lbann.CallbackCheckMetric(
+        metric=metrics[-1].name,
+        lower_bound=val-tol,
+        upper_bound=val+tol,
+        error_on_failure=True,
+        execution_modes='test'))
+
```
timmoon10, line 142:
```suggestion
```

include/lbann/layers/activations/relu.hpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+#ifndef LBANN_RELU_LAYER_INSTANTIATE
+#define PROTO_DEVICE(T, Device) \
+  extern template class relu_layer<T, data_layout::DATA_PARALLEL, Device>; \
+  extern template class relu_layer<T, data_layout::MODEL_PARALLEL, Device>
+
+#include "lbann/macros/instantiate_device.hpp"
+#undef PROTO_DEVICE
+#endif // LBANN_SOFTMAX_LAYER_INSTANTIATE
```
timmoon10, line 127:
```suggestion
#endif // LBANN_RELU_LAYER_INSTANTIATE
```

naoyam:
> * ReLU and leaky ReLU are just entry-wise operations, so why not leave the compute kernels inside LBANN?

Then, input and output tensors would need to be copied between LBANN and Distconv. If LBANN can represent spatially partitioned tensors, we could share underlying memory buffers between them, but that's not the case yet.

bamboo/unit_tests/test_unit_layer_leaky_relu_distconv.py
naoyam: It does not because `model_parallel` is not supported, but it makes sure having a parallel strategy does not induce any error whether the layout is `data_parallel` or `model_parallel`.

bamboo/unit_tests/test_unit_layer_relu_distconv.py
naoyam: Same as Leaky ReLU.

bamboo/unit_tests/test_unit_layer_leaky_relu_distconv.py
timmoon10: I would say it's an invalid configuration, so we should throw an exception.

naoyam: While I agree that's a reasonable behavior, the overall policy so far is to disable distconv and fall back to LBANN, rather than throwing exceptions. I added a note to #1470. 

In any case, the test should still be here and make sure the corresponding exception is thrown.

PR #1508, opened 2020-04-15
Clean up back-prop memory usage

benson31:
The goal of this PR is to dynamically allocate and free memory during back propagation, primarily to ease memory strain on the GPUs.

The current state of this is work-in-progress, but it should be done today. I'm cleaning stuff up. I'll post bamboo results when I have them.


benson31:
I'm going to add some more documentation, but I think the code is ok.

benson31:
[Bamboo is green](https://lc.llnl.gov/bamboo/browse/LBANN-BEN24-13).

PR #1515, opened 2020-04-21
Enable distconv in input

naoyam:
This PR extends `input_layer` for distconv. It copies a mini batch in Hydrogen matrices into distconv GPU tensors partitioned as described in its parallel strategy. When the Hydrogen matrices are already partitioned with the slab-based data reader, it just copies them to distconv tensors. Shuffling is involved, otherwise.

Major caveats:
- Data are copied into GPU memory, even though the `Device` template value is `CPU`. This was just to simplify the integration of distconv into LBANN, where all distconv tensors are assumed to exist in GPU memory. In other words, distconv tensors for each layer, which are held in `data_type_distconv_adatper` class, are all GPU tensors, even though the corresponding LBANN layer has `CPU` device. This can be fixed.
- Currently, only the HDF data reader, which still exists only in the distconv branch, implements slab-based reading, so the input layer always does shuffling (i.e., `m_shuffle_required` is always `true`). Fundamentally, I think a data reader should also be annotated with a parallel strategy, and the input layer uses that information to see whether shuffling is required.

As always, nothing should be impacted unless a parallel strategy is given to the layer.

### DIFF
--- a/src/layers/io/input/input_layer.cpp
+++ b/src/layers/io/input/input_layer.cpp
@@ -26,8 +26,335 @@
 
 #define LBANN_INPUT_LAYER_INSTANTIATE
 #include "lbann/layers/io/input/input_layer.hpp"
+#include "lbann/utils/profiling.hpp"
 
 namespace lbann {
+
+#ifdef LBANN_HAS_DISTCONV
+template <typename TensorDataType, typename T_io_buffer,
+          data_layout T_layout, El::Device Dev>
+input_distconv_adapter<TensorDataType, T_io_buffer, T_layout, Dev>::
+input_distconv_adapter(Layer& layer): data_type_distconv_adapter<TensorDataType>(layer),
+                                      m_shuffle_required(true) {
+  // Input data is only processed when its consumer layer is also
+  // enabled for distconv
+  for (int i = 0; i < layer.get_num_children(); ++i) {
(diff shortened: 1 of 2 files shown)


src/layers/io/input/input_layer.cpp
```diff
@@ -26,9 +26,336 @@
...
+  }
+}
+
+template <typename TensorDataType,
+          typename T_io_buffer,
+          data_layout T_layout,
+          El::Device Dev>
+void input_layer<TensorDataType, T_io_buffer, T_layout, Dev>::
+fp_compute() {
+  generic_input_layer<TensorDataType>::fp_compute();
```
timmoon10, line 352: What happens if `keep_original_outputs` returns false? If we always call `fp_compute`, don't we need to maintain the original buffer for the activations?

naoyam: That's a good question. The original tensor must be always kept, but `keep_original_outputs()` will return `false`. This means `data_type_layer::fp_setup_outputs` is not completed, but as far as I can tell, it still works. I think that that's because the activation matrix is properly resized when `Copy` is called from `partitioned_io_buffer::distribute_from_local_matrix`. 

I'll fix `keep_original_outputs` anway.

naoyam: Fixed with bf1d57492ee5daa4e8295a63d06979a967fc1c76

PR #1517, opened 2020-04-23
Upgrade JAG/MACC scripts to streamline training at scale

samadejacobs:
I have upgraded ICF scripts to streamline training at scale. In particular, both training and evaluation scripts now use parser arguments so you can choose and pick whatever you want. Default is 1M training data, 1K test data with single view and all channels but everything is now controllable from PFE.
Lastly, I added my small python script to check for zero-valued scalar samples. Enjoy!


applications/physics/ICF/eval_macc_surrogate.py
```diff
@@ -0,0 +1,222 @@
...
+                                           execution_modes='test',
+                                           directory=args.dump_outputs,
+                                           batch_interval=1,
+                                           format='npy'),
+                 lbann.CallbackTimer()]
+                                            
+    # Construct model
+    num_epochs =1
+    #serialize io if using single trainer
+    serialize_io = (True if args.procs_per_trainer==0
```
bvanessen, line 180: I am not sure about this flag here.  serialize_io needs to be set for all JAG data fetch, not only when there is a single trainer.

applications/physics/ICF/pre_train_jag_wae.py
```diff
@@ -90,7 +149,7 @@ def construct_model():
 if __name__ == '__main__':
     import lbann
     
-    trainer = lbann.Trainer()
+    trainer = lbann.Trainer(procs_per_trainer=args.procs_per_trainer)
```
bvanessen, line 152: I believe that this model also needs the seriealize_io flag.  The only time that we don't need it is if we are preloading the data store, but even then we should just set it.

applications/physics/ICF/train_macc_surrogate.py
```diff
@@ -95,28 +160,30 @@ def construct_model():
...
+                 lbann.CallbackLoadModel(dirs=str(args.pretrained_dir)),
                  lbann.CallbackTimer()]
                                             
     # Construct model
-    mini_batch_size = 128
-    num_epochs = 100
-    return lbann.Model(mini_batch_size,
-                       num_epochs,
+    #serialize io if using single trainer
+    serialize_io = (True if args.procs_per_trainer==0
```
bvanessen, line 180: Same comment as above.

applications/physics/ICF/pre_train_jag_wae.py
samadejacobs: This model (and the next one) has preload_data_store, so it is the reason I didnt set as am not sure how preload_data_store/use_data_store interacts with serialize_io. In any case, I will set it for all models.

samadejacobs: I have made _serialize_io=True_ default.

PR #1521, opened 2020-04-23
Add data_reader_cosmoflow from distconv

naoyam:
This was originally written by @oyamay. 

@oyamay, please lead merging of this PR. Let me know If you don't think this is no longer necessary as we use the HDF5 data reader.

### DIFF
--- a/include/lbann/data_readers/CMakeLists.txt
+++ b/include/lbann/data_readers/CMakeLists.txt
@@ -3,6 +3,7 @@ set_full_path(THIS_DIR_HEADERS
   compound_data_reader.hpp
   data_reader.hpp
   data_reader_cifar10.hpp
+  data_reader_cosmoflow.hpp
   data_reader_csv.hpp
   data_reader_image.hpp
   data_reader_imagenet.hpp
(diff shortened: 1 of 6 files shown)


naoyam:
One note that the data reader does not do any type conversion. Since the Cosmoflow samples are `int16`, we actually need to cast them to `IODataType` of `input_layer` if `IODataType` is not `int16`. The code used to have casting, but I believe we should not do that here, but rather we should use `int16` in `input_layer`'s `TensorDataType` and `IODataType` so that all I/O is done in `int16`. Casting to `TensorDataType` of subsequent layers is done at layer boundaries automatically.

Since LBANN does not instantiate `int16` type for the layers, I don't think the above approach works at this point, but we should really add `int16` to the type instantiation list. (Pinging @benson31 and @timmoon10).

oyamay:
@naoyam This data reader still works with the latest Distconv-version of LBANN (d30e053b6), but the CosmoFlow dataset is also supported by the HDF5 data reader, which is more generic than this data reader. So I think we don't have to merge this to the develop branch except for research purposes specific to CosmoFlow.

naoyam:
OK, let's close this for now.

PR #1522, opened 2020-04-24
Merging data_reader_hdf5 and slab extensions

naoyam:
This PR merges distconv extensions for HDF5 data reader and slab-based reading.

This PR needs much more robust design but is intended to show what's been working for Cosmoflow. There are a lot of ad-hoc changes scattered around data store, data reader and io buffer. The fundamental reason of these changes is that in the baseline LBANN a data reader is assumed to return full samples, which is no longer the case with slab reading.

Another thing needed here is to determine how each sample should be partitioned. Currently, it is assumed to be partitioned in the depth dimension, and the number of partitions is provided by an environment variable, which can be queried with `dc::get_number_of_io_partitions()`. I think that the parallel strategy of `input_layer` should be propagated here, and the use of the environment variable should be replaced.

### DIFF
--- a/include/lbann/data_readers/CMakeLists.txt
+++ b/include/lbann/data_readers/CMakeLists.txt
@@ -18,5 +18,10 @@ set_full_path(THIS_DIR_HEADERS
   data_reader_synthetic.hpp
   )
 
+if (LBANN_HAS_DISTCONV)
+  list(APPEND THIS_DIR_HEADERS
+    "${CMAKE_CURRENT_SOURCE_DIR}/data_reader_hdf5.hpp")
+endif ()
+
 # Propagate the files up the tree
 set(HEADERS "${HEADERS}" "${THIS_DIR_HEADERS}" PARENT_SCOPE)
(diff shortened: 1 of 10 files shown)


naoyam:
Note that the HDF5 data reader is intended to be used for the Cosmoflow datasets. Although it should work with any other datasets with the same format, it can be more generalized since HDF5 is self descriptive.

naoyam:
Another thing to note is that the HDF5 reader does not currently work. It assumes `TensorDataType` is `int16_t`, but the type is not instantiated. Previously, in the distconv branch, we worked around by using float buffers to store int16_t data, but that should not be a permanent solution.

oyamay:
The change itself looks good to me, but this branch needs to be updated to merge to the develop branch as the way of model initialization has been changed. I will update and create a new PR to do it (and another PR will follow to fix the data type issue of the data reader ([oyamay/hdf5-data-reader-enhancement](https://github.com/oyamay/lbann/tree/hdf5-data-reader-enhancement))).

oyamay:
This PR will be merged by #1572.

bvanessen:
This PR has been superseded by #1572 and is now redundant.

PR #1523, opened 2020-04-24
Bugfix trainer specific callback outputs

bvanessen:
Ensure that dump outputs places output files in a multi-trainer aware directory hierarchy.  Also added support for a trainer master aware make directory function that is inspired by PR #1459 but places the code in a separate utility file that requires access to the comm object.  Additionally, added helper functions to the callback and execution context classes to get a multi-trainer aware directory hierarchy.

### DIFF
--- a/include/lbann/callbacks/callback.hpp
+++ b/include/lbann/callbacks/callback.hpp
@@ -187,6 +187,38 @@ public:
 
   ///@}
 
+  /** @brief Build a standard directory hierachy including trainer,
+   * execution context, and model information (in that order).
+   */
+  inline std::string get_multi_trainer_ec_model_path(const model& m,
+                                                     const std::string& root_dir) {
+    std::string dir = root_dir;
+    if (dir.empty()) { dir = "./"; }
+    if (dir.back() != '/') { dir += "/"; }
+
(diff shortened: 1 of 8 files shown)


include/lbann/callbacks/callback.hpp
```diff
@@ -187,6 +187,38 @@ class callback_base {
...
+  }
+
+  /** @brief Build a standard directory hierachy including trainer,
+   * model information in that order.
+   */
+  inline std::string get_multi_trainer_model_path(const model& m,
+                                                  const std::string& root_dir) {
+    std::string dir = root_dir;
+    if (dir.empty()) { dir = "./"; }
+    if (dir.back() != '/') { dir += "/"; }
```
samadejacobs, line 213: Can we also handle the situation where root directory is given (that is _dir is not empty_) but doesn't exist yet? That is make non-existing but requested _root_dir_  as well.

bvanessen: Yes the make_directory command that is eventually called will make all non-existent directories in the path.

bvanessen:
> Looks good. I will close #1459 without merging.

Okay, but lets create a new PR that uses this function to fix the bugs that you identified.

PR #1526, opened 2020-04-28
Remove deprecated model prototext files

bvanessen:
Removed models that are deprecated or migrated into the Python front end.


bvanessen:
@timmoon10 @benson31 Can you please take a look at the changes that I made to the tests.  I am not sure why it is unable to find the lbann python packages.

bamboo/integration_tests/test_integration_lenet.py
```diff
@@ -14,6 +14,15 @@
...
 import tools
 
+# Applications directories
+bamboo_dir = os.path.dirname(current_dir)
+lbann_dir = os.path.dirname(bamboo_dir)
+sys.path.insert(0, os.path.join(lbann_dir, 'applications', 'vision'))
+import data.mnist
+
+# Python directory
+sys.path.insert(0, os.path.join(lbann_dir, 'python'))
```
timmoon10, line 24: Importing the Python package in the source directory will not work since it doesn't have access to the LBANN executable or the Protobuf messages. You need to import the Python package from the install directory.

bamboo/integration_tests/test_integration_lenet.py
```diff
@@ -14,6 +14,15 @@
 sys.path.insert(0, os.path.join(os.path.dirname(current_dir), 'common_python'))
 import tools
 
+# Applications directories
+bamboo_dir = os.path.dirname(current_dir)
+lbann_dir = os.path.dirname(bamboo_dir)
+sys.path.insert(0, os.path.join(lbann_dir, 'applications', 'vision'))
+import data.mnist
```
timmoon10, line 21: I'd suggest copy-pasting the MNIST and ImageNet scripts into `bamboo/common_python`. The Bamboo framework is going to be horribly messy and complicated (until we have build-based testing) and it would be a bad idea for that messiness to contaminate the application directory. After all, the LeNet model is supposed to be simple and useful for learning how to use LBANN.

bamboo/integration_tests/test_integration_lenet.py
bvanessen: Okay.  Fixed this.

bamboo/integration_tests/test_integration_lenet.py
bvanessen: Done.

bvanessen:
@timmoon10 @benson31 It is now passing on Pascal and Corona.  Ray of course is still failing and Catalyst is still running.

bamboo/common_python/data/mnist/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.gz
+train-images-idx3-ubyte
+train-labels-idx1-ubyte
+t10k-images-idx3-ubyte
+t10k-labels-idx1-ubyte
```
timmoon10, line 5: This file isn't needed since we aren't going to download MNIST.

PR #1501, opened 2020-04-08
Smiles data reader

davidHysom:
Data reader for SMILES data

PR #1528, opened 2020-05-02
smiles_data_reader rev 2

davidHysom:
This is a follow-on to PR #1501, and should not be merged until that PR is merged.

Primary features of this PR:

Major revision of the test_encode() self-test to make it more comprehensive; this entailed addition of the decode_smiles() method.

The self-test is activated via the cmd line flag: --test_encode
In order to test the test, you can additionally add: --make_test_fail

Revises encode_smiles() and associated code, to use "const char*" instead of "string" as input. This should improve efficiency (reduce run time), though the improvement will likely be lost in the noise.


davidHysom:
Closing this PR, since there are many merge CONFLICTS, and getting them reconciled correctly is more work than starting over.

PR #1529, opened 2020-05-05
Set weight value callback

samadejacobs:
Bamboo (pascal gpu) is green.


src/callbacks/set_weights_value.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include <callbacks.pb.h>
+
+
+namespace lbann {
+namespace callback {
+
+void set_weights_value::on_epoch_begin(model *m) {
+  const auto& c = static_cast<const sgd_execution_context&>(m->get_execution_context());
+  if(El::Int(c.get_epoch()) !=  m_epoch_interval)  return;
```
timmoon10, line 40:
```suggestion
  const auto& c = m->get_execution_context();
  if (c.get_step() != m_step) { return; }
```

src/proto/callbacks.proto
```diff
@@ -334,4 +335,10 @@ message Callback {
   message CallbackPrintModelDescription {
   }
 
+  message CallbackSetWeightsValue {
+    string weight_name = 1;
+    float  weight_value = 2;
+    int64  epoch_interval = 3;
+  }
```
timmoon10, line 342:
```suggestion
  /** Set value of a weights object at a given training step. */
  message CallbackSetWeightsValue {
    string weights = 1;
    double value = 2;
    int64 step = 3;
  }
```

include/lbann/callbacks/set_weights_value.hpp
```diff
@@ -0,0 +1,79 @@
...
+  set_weights_value* copy() const override {
+    return new set_weights_value(*this);
+  }
+  void on_epoch_begin(model *m) override;
+  std::string name() const override { return "set weights value"; }
+ private:
+  /** weight name. */
+  std::string m_weight_name;
+  /** new weight value */
+  float m_weight_value;
```
timmoon10, line 66:
```suggestion
  double m_weight_value;
```
This is slightly more general.

timmoon10:
I've changed the callback to be based on mini-batch steps rather than epochs. I've also added a Bamboo unit test, which passes when I run in manually. I'll run Bamboo after it comes back online.

timmoon10:
[Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM283-1) doesn't show any new failures.

PR #1532, opened 2020-05-07
Add access to cuDNN tensor core functionality, controlled by cmake

benson31:
This is a fairly broad hammer for applying this functionality, and it will eventually be controlled on a layer-by-layer basis. For now, a CMake flag controls access and it's applied to all convolution layers.


benson31:
Closing in favor of #1534.

PR #1534, opened 2020-05-07
Add per-convolutional-layer support for cuDNN tensor operations

benson31:
This is a redo of #1532 that controls access dynamically. It's ugly and hacky, but it's cuDNN-related so it's all getting ripped apart in other work anyway; don't read too much into it...

I've only updated `lbann.cpp`. The others will obviously default to no tensor cores.

`applications/vision/resnet.py` shows the enum being used.

### DIFF
--- a/applications/vision/lenet.py
+++ b/applications/vision/lenet.py
@@ -26,6 +26,7 @@ images = lbann.Identity(input_)
 labels = lbann.Identity(input_)
 
 # LeNet
+tensor_ops_mode = lbann.ConvTensorOpsMode.DEFAULT_TENSOR_OPS
 x = lbann.Convolution(images,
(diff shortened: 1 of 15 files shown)


benson31:
My thought is either this or #1532, but not both. Thoughts?

benson31:
@timmoon10 I chose `lenet.py` because it explicitly lists the model's layers instead of hiding them deep in some opaque pre-defined network file. The purpose was to illustrate accessing the enum values from python. Nonetheless, I've moved the example code to `resnet.py` but I don't think it's as illustrative.

benson31:
Am I correct that we prefer this PR over #1532? If so, I'll close that one. @oyamay do you have any thoughts on this?

PR #1535, opened 2020-05-07
Fail hard if an exception is not caught by the time it reaches main()

benson31:
The current mechanism will hang indefinitely if a proper subset of processes throws an error (e.g. in gradient checking in the python unit tests). This is bad. We abort instead.

### DIFF
--- a/model_zoo/lbann.cpp
+++ b/model_zoo/lbann.cpp
@@ -139,10 +139,14 @@ int main(int argc, char *argv[]) {
       e.print_report(fs);
     }
     El::ReportException(e);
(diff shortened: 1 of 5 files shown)


benson31:
I don't care about the other front-ends because they aren't used in testing. That said, I'm happy to copy-paste into all them if there's actually a need.

benson31:
Updated.

davidHysom:
> the other front ends.

The only other (c++) front end I see is lbann_help, and nothing there throws an exception

benson31:
Because I fixed the other ones already.

PR #1538, opened 2020-05-08
Move I/O buffers and control to data coordinator

bvanessen:
Created a buffered data coordinator child class to hold data type
specific I/O buffers in the data coordinator.  This moves the
functionality for holding the data out of the input layer.


bvanessen:
@timmoon10 @benson31 I believe that this PR is ready for review.  It is passing all bamboo tests locally except for tessellate.   When reviewing this please remember that this is supposed to be as close to a "straight" refactor as possible.  There are a number of future optimizations that we will want to do, but I wanted to get this pushed out first before adding more changes.

src/io/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_subdirectory(data_buffers)
+#add_subdirectory(data_buffers)
```
timmoon10, line 1:
```suggestion
```

src/training_algorithms/sgd_training_algorithm.cpp
```diff
@@ -138,7 +144,7 @@ bool sgd_training_algorithm::train_mini_batch(sgd_execution_context& c,
 
   // Update step
   model.update_weights();
-  finished = model.update_layers();
+  /*finished = */model.update_layers();
```
timmoon10, line 147:
```suggestion
  model.update_layers();
```

timmoon10 (approved):
LGTM

PR #1539, opened 2020-05-08
Fix PFE where trainer requires mini_batch_size

bvanessen:
Fixed PFE trainer invocations that didn't get properly updated with the mini-batch size argument.

### DIFF
--- a/applications/physics/ICF/pre_train_jag_wae.py
+++ b/applications/physics/ICF/pre_train_jag_wae.py
@@ -96,8 +96,8 @@ def construct_model():
 
     z = lbann.Gaussian(mean=0.0,stdev=1.0, neuron_dims="20")
     model = macc_models.MACCWAE(args.zdim,args.ydim,use_CNN=args.useCNN)
-    d1_real, d1_fake, d_adv, pred_y  = model(z,gt_y) 
-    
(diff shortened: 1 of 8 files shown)


applications/physics/ICF/pre_train_jag_wae.py
```diff
@@ -82,7 +82,7 @@ def construct_model():
     import lbann
 
     # Layer graph
-    input = lbann.Input(target_mode='N/A',data_set_per_model=True, name='inp_data')
```
samadejacobs, line 85: Why remove _data_set_per_model_ from here and other places?

bvanessen: As a part of the design, this went away.  A data reader is global to a trainer and all models will share it.  We also deprecated the use case for the data set per model.

bvanessen:
> In some places (all places?) mini-batch-size is passed to both Trainer and Model. This seems like a point of failure; can't it be passed to one, but not both?
> 
> This does solve the error I was getting with train_atom_char_rnn.py

So for some of the models, the python setup code still needed the mini-batch size to configure something else in the model.  Can you please point out to where this is going on can double check.  Also what error are you getting with your run.  I cannot execute that file due to a lack of torch.

PR #1542, opened 2020-05-12
Fix a bug related to some recent changes in Hydrogen

benson31:
Recent changes in Hydrogen introduced a new stream on which Aluminum communications would be run. This was supposed to be an implementation detail, but it does not elegantly handle the nonblocking case.

I've discussed this with @ndryden and we've hypothesized that the NCCL backend should have its own stream as an implementation detail of Aluminum, which would hide this better. This patch should hold LBANN over until we get that in.

### DIFF
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -230,11 +230,13 @@ void nb_allreduce_aluminum(El::Matrix<T, El::Device::GPU>& m,
                            typename BackendT::allreduce_algo_type algo
                            = BackendT::allreduce_algo_type::automatic) {
   const auto local_size = m.Height() * m.Width();
+  const auto& syncinfo = El::SyncInfoFromMatrix(m);
+  req.req_syncinfo_ = syncinfo;
   ::Al::NonblockingAllreduce<BackendT>(
     m.Buffer(),
     local_size,
     mpi_op_to_al_op(op),
-    c.template GetComm<BackendT>(El::SyncInfoFromMatrix(m)),
+    c.template GetComm<BackendT>(syncinfo),
     GetRequest(req, tag),
     algo);
 }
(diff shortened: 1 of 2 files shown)


src/comm.cpp
```diff
@@ -230,13 +246,16 @@ void nb_allreduce_aluminum(El::Matrix<T, El::Device::GPU>& m,
...
   ::Al::NonblockingAllreduce<BackendT>(
     m.Buffer(),
     local_size,
     mpi_op_to_al_op(op),
-    c.template GetComm<BackendT>(El::SyncInfoFromMatrix(m)),
-    GetRequest(req, tag),
+    c.template GetComm<BackendT>(syncinfo),
+    request,
     algo);
+  UpdateRequest(request, syncinfo);
```
timmoon10, line 258: Would it be better to put this before the allreduce, just in case Aluminum does something funky when the `req_type`'s stream is wrong?

benson31: No; the request gets created as part of the call to Aluminum. (For Allreduce, e.g., [this line](https://github.com/LLNL/Aluminum/blob/3f4170fce069eef773f25142665afc16fb5d3ee5/src/nccl_impl.hpp#L206) calls the common `setup_completion_event()` function, which [creates the request](https://github.com/LLNL/Aluminum/blob/3f4170fce069eef773f25142665afc16fb5d3ee5/src/nccl_impl.hpp#L354). The `UpdateRequest()` line just resets the stream to the right thing.

benson31:
@timmoon10 Aluminum does the right thing here; the issue is a change I made in hydrogen to try to hide the fact that I wanted all NCCL communications on their own stream. So there are 3 streams in play here:

1. The compute stream that the communication should block (for the blocking case) or that will eventually use the result (for the non-blocking case).
2. The stream that Hydrogen shims in to force all blocking NCCL communications to serialize.
3. The internal stream that Hydrogen uses for the nonblocking communication.

There is no problem for blocking communication. Streams 1 and 2 synchronize around the communication and stream 3 is never used.

For nonblocking communication, however, streams 1, 2, and 3 synchronize around the launch, but only streams 2 and 3 synchronize around the completion. Since stream 2 is a "fake stream" (impl detail of hydrogen not exposed to users), this doesn't help when LBANN takes a hydrogen communicator (the source of the fake stream) and makes a raw Aluminum call with it.

A few solutions would be:

1. Update Aluminum so that it only ever uses 1 stream for NCCL communications so they serialize automatically at the Aluminum level.
2. Update Hydrogen to wrap the nonblocking Aluminum calls and just decree that if you're using a Hydrogen communicator object, you must use Hydrogen's MPI interface for all communications. If you do not, this should "at your risk".

I like 2 better, but it requires us to expose the rest of the nonblocking interface of Aluminum via Hydrogen. It has nice symmetry, and then my multistream Gemm code can do whatever it wants as long as the Hydrogen interface is consistent.

We should still do 1, though, per vendor guidance that except in very specific situations (basically one communication is strictly intranode and another is strictly internode), NCCL will not actually overlap the communciation.

PR #1546, opened 2020-05-21
Real support for mixed-precision

benson31:
Eventually this will wind its way to real support for mixed precision. In the meantime, please enjoy this implementation of covariant smart-pointer returns, based on [this blog post](https://www.fluentcpp.com/2017/09/12/how-to-return-a-smart-pointer-and-use-covariance/).

Well, it's wound its way there. One major shortcoming is that the gradients are still accumulated to FP32 before accumulating into the weights. This is unavoidable until we have optimization kernels (Axpy, mostly) that can manage mixed-precision inputs.


include/lbann/weights/weights_proxy.hpp
```diff
@@ -0,0 +1,170 @@
...
+  {
+    if (!values_->Viewing())
+      El::Copy(master_weights_->get_values(), *values_);
+  }
+
+  /** @brief Access the values. */
+  ValuesType const& values() const noexcept { return *values_; }
+
+  /** @brief Access the master weights object directly. */
+  weights const& master_weights() const noexcept { return *master_weights_; }
```
timmoon10, line 130:
```suggestion
  /** @brief Access the values. */
  ValuesType const& values() const { return *values_; }

  /** @brief Access the master weights object directly. */
  weights const& master_weights() const { return *master_weights_; }
```
Won't we hit an exception if the user requests the values before setup?

timmoon10, line 130:
```suggestion
  /** @brief Access the values. */
  ValuesType const& values() const { return *values_; }

  /** @brief Access the master weights object directly. */
  weights const& master_weights() const { return *master_weights_; }
```
Won't we hit a segfault if the user calls these functions before setup?

include/lbann/weights/weights_proxy.hpp
```diff
@@ -0,0 +1,170 @@
...
+    ValuesPtrType ret(vals.Construct(vals.Grid(), vals.Root()));
+    El::LockedView(*ret, vals);
+    return ret;
+  }
+
+  /** @brief Establish the target matrix storage. */
+  ValuesPtrType setup_values_(weights const& w) const
+  {
+    if (auto dtw = dynamic_cast<data_type_weights<TensorDataType> const*>(&w))
+      return setup_values_(*dtw);
```
timmoon10, line 147: Do we handle the case where weights have different devices or data layouts? A better way is to instantiate and align `values_` first, and then decide whether to view or align by comparing the `DistData` s. It also doesn't require any dynamic casting.

include/lbann/weights/weights_proxy.hpp
timmoon10, line 130:
```suggestion
  /** @brief Access the values. */
  ValuesType const& values() const { return *values_; }

  /** @brief Access the master weights object directly. */
  weights const& master_weights() const { return *master_weights_; }
```
Won't we hit a segfault if the user calls these before setup?

include/lbann/weights/weights_proxy.hpp
```diff
@@ -0,0 +1,170 @@
...
+ *  throughout the iteration. At the end of the iteration, the proxy
+ *  values are considered invalid once gradient updates have begun.
+ *
+ *  @tparam TensorDataType The type to which the weights are proxied.
+ */
+template <typename TensorDataType>
+class WeightsProxy
+{
+  using DataTypeWeights = data_type_weights<TensorDataType>;
+  using ValuesType = El::AbstractDistMatrix<TensorDataType>;
```
timmoon10, line 52: Outside the scope of this PR: It would be nice to add template arguments so `ValuesType` is a concrete `DistMatrix`. This could handle cases where CPU and GPU layers can share the same weights, or two layers with different data layouts. We would need to update all the layers to manually specify the data layout they need (unlike our current approach, where they just hope it already is the way they need it).

include/lbann/weights/weights_proxy.hpp
benson31: Dereferencing pointers is a `noexcept` operation. Dereferencing a `nullptr` is a segfault, which is still not an exception. The contract of this class is that these will not be accessed if `this->empty()`. I will add debug-mode checks to enforce this contract, which will suitably adjust the `noexcept` tag.

include/lbann/weights/weights_proxy.hpp
benson31: Given the current design of the larger LBANN, I'm not sure this would be valuable without a lot of extra effort. Even though matrix distributions are statically determinable (they appear to be functions of `data_layout`), we treat them dynamically. So unless we make this polymorphic with the construction similarly (dynamically) branched, I think this would be hard. A temporary alternative might be to take a (dynamically-determined) `DistData` object that could be factored into the "view or copy" decision.

benson31:
This should be ok to review now.

One glaring shortcoming could be batchnorm. Right now, it implicitly assumes the weights are all of its `TensorDataType`, so using it in a mixed setting should throw. It should be fixed; it's not clear that the fix is simple enough to do it here. I'd rather do a separate PR once I understand a bit better how BN should work in a mixed setting.

To mitigate the potential issues here, I haven't added the batchnorm layers to the "real" layer factory, so they're only constructible at `DataType`. You would have to manually setup weights of a different data type to induce a failure now.

src/layers/regularizers/batch_norm_helpers.hpp
```diff
@@ -0,0 +1,71 @@
...
+/** @class SafeWeightsAccessor
+ *  @brief Ensure safe access to weights objects' data.
+ */
+template <typename TensorDataType>
+struct SafeWeightsAccessor
+{
+  using ValuesType = El::AbstractDistMatrix<TensorDataType>;
+  using DataTypeWeights = data_type_weights<TensorDataType>;
+
+  static ValuesType& mutable_values(weights& w)
```
timmoon10, line 58: I wonder if this functionality could be moved into the `WeightsProxy` class. It could be templated on the desired data type or it could take in a matrix that would be returned as a view. This would be useful for other hacky layers, like the dist embedding layer.

include/lbann/optimizers/optimizer.hpp
```diff
@@ -422,21 +422,69 @@ void optimizer::accumulate_all_gradient_contributions(
...
+                  to_string(grad_mgr.get_status()));
+    }
+    El::Copy(grad_mgr.gradient(), gradient);
+  }
+  else if (this->gradients_.size() > 1UL) {
+    // Need a temporary matrix for the type-casted copy.
+    auto tmp = std::unique_ptr<AbsDistMatType>{
+      gradient.Construct(gradient.Grid(), gradient.Root())};
+
+    for (auto const& grad_mgr_v : this->gradients_) {
```
timmoon10, line 481: Don't you need to skip if the datatype is `TensorDataType`, or else you'll double count?

benson31: Whoops. In my head, I had done that. Clearly need more coffee.

benson31: Fixed.

include/lbann/weights/weights_proxy.hpp
```diff
@@ -0,0 +1,398 @@
...
+    this->internal_setup_(w);
+  }
+
+  /** @brief Synchronize the held values with the master set.
+   *
+   *  If empty(), this function takes the view that there is no master
+   *  with which to synchronize, so no action is required -- it is a
+   *  no-op.
+   */
+  void synchronize_with_master()
```
bvanessen, line 263: If multiple layers have a proxy to a weight, this synchronize would clobber each other, right?  Is there a synch collective allreduce operations?

oyamay:
There are some compilation issues with Distconv, but since the Distconv feature is currently not working due to other problems, so I will work on it in another PR.

include/lbann/weights/weights_proxy.hpp
```diff
@@ -0,0 +1,398 @@
...
+
+#include "lbann_config.hpp"
+
+#include "lbann/base.hpp"
+#include "lbann/weights/data_type_weights.hpp"
+#include "lbann/weights/weights.hpp"
+
+#if defined LBANN_DEBUG
+#define LBANN_DEBUG_ASSERT_POINTER(ptr) \
+  do { if (ptr) LBANN_ERROR("Pointer \"" #ptr "\" is null."); } while (0)
```
benson31, line 37:
```suggestion
  do { if (!ptr) LBANN_ERROR("Pointer \"" #ptr "\" is null."); } while (0)
```
Oops

benson31:
[Bamboo finally matches develop](https://lc.llnl.gov/bamboo/browse/LBANN-BEN30-32). Good luck, everybody else!

PR #1549, opened 2020-05-26
Fix LBANN inference

bvanessen:
Updated the load_model callback so that it works on inference only
runs.  Updated the lbann_inf front end so that it uses the
lbann_library support for the load_model callback, rather than an
explicit call to a static function.  Changed how the evaluate function
sets up the model and data coordinator so that it has the execution
context and mode at the start of the function. This fixes a bug where
a model is called in inference only.

### DIFF
--- a/applications/physics/ICF/eval_macc_surrogate.py
+++ b/applications/physics/ICF/eval_macc_surrogate.py
@@ -1,7 +1,7 @@
-import macc_models 
+import macc_models
 import argparse
 import os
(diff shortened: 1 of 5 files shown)


bvanessen:
I have fixed the eval script.  Yes we need a unit test for lbann_inf and so many other things.  Right now things like the load weights callback are done as a callback, but should really be first class properties of the base code.  Since it is a callback, that can be implicitly added that is why you get the warning.

PR #1553, opened 2020-05-28
Optimize GPU performance of slice and concat layers

timmoon10:
This PR makes two changes to the slice and concatenation layers:

- Use Hydrogen `simple_buffer`s for GPU workspaces rather than Thrust `device_vector`s. Thrust synchronizes the CUDA stream whenever it creates a `device_vector`, hurting kernel pipelining.
- CUDA kernels use nested loops to iterate through tensors. Previously we had a single loop and computed tensor positions with divides and moduli, but the kernel runtime was dominated by integer arithmetic. The new approach is optimized for the slice layers in multi-head attention, although it may be susceptible to bad performance with certain tensor shapes (if there are many children/parents, if an NCHW tensor is dominated by the N dimension, if the children/parent tensors have a wide range of sizes).

Timings with a toy model with repeated slices and concats (1 proc on Pascal):

| Commit | Mini-batch time |
|----|----|
| 9a9e31c (develop) | 0.0284 s |
| 14e00c4 | 0.0281 s |
| df70e84 | 0.0128 s |

Timings with Transformer (1 Pascal node):

| Commit | Mini-batch time |
|----|----|
| 9a9e31c (develop) | 0.276 s |
| 14e00c4 | 0.286 s |
| df70e84 | 0.259 s |

[Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM284-1) doesn't show any new failures.

Notes for future optimizations:
- We map the CUDA kernel grid to the C, H, and W dimensions of an NCHW tensor. Is it worthwhile coming up with a separate implementation if N is large?
- We don't get full occupancy on the SMs because the kernel requires too many registers per thread (48 on Pascal). I tried staging some of the values in shared memory, but that ended up increasing the number of required registers.
- The kernel runtime is still dominated by integer arithmetic.
- In a toy model with repeated slices and concats, kernel pipelining is perfect.

### DIFF
--- a/src/layers/transform/concatenate.cu
+++ b/src/layers/transform/concatenate.cu
@@ -199,6 +199,13 @@ void fp_compute_impl(
                 "but only 3-D tensors are currently supported");
   }
 
+  // Get synchronization info from output tensor
+  using LocalMatrix = El::Matrix<TensorDataType, El::Device::GPU>;
+  auto& output = l.get_activations();
+  auto& local_output = dynamic_cast<LocalMatrix&>(output.Matrix());
+  auto&& sync_info = El::SyncInfoFromMatrix(local_output);
+  auto&& stream = sync_info.stream_;
+
   // Get dimensions and strides for each input tensor
   const size_t num_inputs = l.get_num_parents();
   std::vector<const TensorDataType*> input_buffer_list;
@@ -232,7 +239,6 @@ void fp_compute_impl(
 
   // Get strides for output tensor
   dim4 output_strides;
-  auto& output = l.get_activations();
   {
     const auto& output_dims = l.get_output_dims();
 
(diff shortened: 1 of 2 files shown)


timmoon10:
Timings with the ATOM RNN model:
| Commit | Mini-batch time |
| --- | --- |
| 7e3d1bf (develop) | 0.0826 s |
| df70e84 | 0.0758 s |

timmoon10:
My original toy model (repeated slices and concats) doesn't scale, likely because it is small enough for the runtime to be dominated by communication. I bumped up the number of layers and the mini-batch size and ran on Pascal:

| Nodes | Mini-batch time |
|---|---|
|1 | 0.260 sec |
| 2 | 0.166 sec |
| 4 | 0.134 sec |
| 8 | 0.110 sec |

PR #1559, opened 2020-06-09
Added an additional guard to when we do in-training validation work.

bvanessen:
If there is no validation data reader then the call to evaluate will
partially set the state of the data coordinator before checking to
bail out if there is no valid data reader.  This is a temporary
inconsistency while we are cleaning up the data coordinator.  Added an
extra guard to prevent evaluation when there is no valid data reader,
before calling the evaluate function.


bvanessen:
This should fix the bamboo issues on entrywise batchnorm testing.

PR #1560, opened 2020-06-11
Cleanup and consolidate initialization of the RNGs

bvanessen:
Cleaned up the initialization of the random seed in LBANN library and
associated front ends.  Since the seed can be set by the user for each
trainer, the random seed should be initialized by the construction of
the trainer.  Remove the multiple RNG initializations that occur in
the base.cpp initialize function and multiple times in the
lbann_library construct trainer function.  Now the random seeds are
set by the user (possibly) and then have the trainer and global IDs
mixed in appropriately.  Also fixed a bug where the trainer rank was
never actually properly mixed in with the hash_combine function.  This
also adds the reporting of the user specified parameters to the
contruct_trainer function and properly reports the random seed that is
used.

### DIFF
--- a/include/lbann/trainers/trainer.hpp
+++ b/include/lbann/trainers/trainer.hpp
@@ -66,7 +66,9 @@ public:
   /** Archive for checkpoint and restart */
   template <class Archive> void serialize(Archive & ar) {
     ar(CEREAL_NVP(m_persist),
-       CEREAL_NVP(m_max_mini_batch_size));
+       CEREAL_NVP(m_max_mini_batch_size),
+       CEREAL_NVP(m_random_seed),
+       CEREAL_NVP(m_data_seq_random_seed));
   }
 
   /** Set the trainer's name; this is an arbitrary string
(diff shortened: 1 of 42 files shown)


include/lbann/trainers/trainer.hpp
```diff
@@ -66,7 +65,9 @@ class trainer {
   /** Archive for checkpoint and restart */
   template <class Archive> void serialize(Archive & ar) {
     ar(CEREAL_NVP(m_persist),
-       CEREAL_NVP(m_max_mini_batch_size));
+       CEREAL_NVP(m_max_mini_batch_size),
+       CEREAL_NVP(m_random_seed),
+       CEREAL_NVP(m_data_seq_random_seed));
```
timmoon10, line 70: Does the checkpoint need access to the original seeds? Aren't the current seed values the important thing for reproducibility?

src/utils/lbann_library.cpp
```diff
@@ -192,6 +195,9 @@ std::unique_ptr<trainer> construct_trainer(lbann_comm *comm,
       std::cout << "\n"
                 << trainer->get_description()
                 << std::endl;
+
+      // User feedback
+      print_parameters(*comm, pb, random_seed, data_seq_random_seed);
```
timmoon10, line 200: The reported seeds will be different from the user-provided seeds since they've gone through several `hash_combine`. This may confuse users who manually set the seed.

bvanessen: I have improved the reporting to include the base and the actual used parameters.

include/lbann/trainers/trainer.hpp
bvanessen: Some of the seeds are set by the command line arguments so if a new execution doesn't have all of the right flags, they would be different.

timmoon10: But once the model starts training, the RNG seeds are going to change as the generator is used. At that point, the original values and their origins don't matter anymore for restarting.

PR #1563, opened 2020-06-15
Quick refactor of weights layer

benson31:
We have more static type knowledge that is easy to exploit in this layer.

I doubt that it will be significant in terms of performance, but I did this whilst debugging something else. It didn't fit with the rest of that PR, but I didn't want it to go to waste. In the interest of smaller and easier PRs, I've split it off into this very little PR.


benson31:
[Bamboo is actually green](https://lc.llnl.gov/bamboo/browse/LBANN-BEN31-2), which is weird considering Nightly...

PR #1564, opened 2020-06-16
Initialize thread-local RNGs used for I/O in a deterministic way

JaeseungYeom:
Instead of using the thread id, which may change run-to-run, for seeding thread specific RNGs used for I/O, use the local index number assigned to the thread.


bvanessen:
@JaeseungYeom Please rebase onto develop.  I just merged PR #1560 which rearranged things somewhat.  It didn't solve the instability, but is a better foundation.

JaeseungYeom:
I am closing this because PR #1575 addresses the bug

PR #1565, opened 2020-06-17
fixed empty data_config default

ben-bay:
Intended for @samadejacobs 

Along with some formatting changes and a `requirements.txt` file, this fixes the first issue I ran into when the `data_config` would not load by default.


ben-bay:
@samadejacobs these changes fix the script for me (it begins training, at least).

PR #1566, opened 2020-06-18
Bugfix for lbann inf

bvanessen:
Check to make sure that there is a valid data set.


model_zoo/lbann_inf.cpp
```diff
@@ -81,9 +81,11 @@ int main(int argc, char *argv[]) {
 
     thread_pool& io_thread_pool = trainer->get_io_thread_pool();
     int training_dr_linearized_data_size = -1;
-    auto *dr = trainer->get_data_coordinator().get_data_reader(execution_mode::training);
+    auto *dr = trainer->get_data_coordinator().get_data_reader(execution_mode::testing);
     if(dr != nullptr) {
       training_dr_linearized_data_size = dr->get_linearized_data_size();
+    }else {
```
benson31, line 87:
```suggestion
    } else {
```
or
```suggestion
    }
    else {
```
I favor the latter. Cuddled `else`s are the worst.

PR #1569, opened 2020-06-22
3D Convolutional Auto Encoder Example for MOF Data 

szaman19:
- 3D convolutional autoencoder for Metal-Organic Frameworks
- Data ingested using Python DataReader from custom dataset extending PyTorch Dataset 
    - Each data point is a 11x32x32x32 tensor
- Dataset unit test in `test/dataset_test.py`
- Integration test in `test/test_integration_mof.py'
- Currently, the data is expected to be a .npy in `data/mofs.npy'


applications/MOF/MOFae.py
```diff
@@ -0,0 +1,287 @@
+import argparse 
+import lbann 
+#import data.mnist
```
samadejacobs, line 3: Please delete

applications/MOF/MOFae.py
```diff
@@ -0,0 +1,287 @@
...
+# Command-line arguments
+# ----------------------------------
+
+## TO DO. Add more CLI arguemnts for Model storage and Generated Data
+
+desc = ("Training 3D-CAE on 4D MOF Data using LBANN")
+
+parser = argparse.ArgumentParser(description = desc)
+
+'''
```
samadejacobs, line 18: Please uncomment or delete

applications/MOF/MOFae.py
```diff
@@ -0,0 +1,287 @@
...
+    help='job name', metavar='NAME')
+parser.add_argument(
+    '--mini-batch-size', action='store', default=128, type=int,
+    help='mini-batch size (default: 128)', metavar='NUM')
+parser.add_argument(
+    '--num-epochs', action='store', default=100, type=int,
+    help='number of epochs (default: 100)', metavar='NUM')
+
+lbann.contrib.args.add_scheduler_arguments(parser)
+'''
```
samadejacobs, line 43: Please uncomment or delete

applications/MOF/README.md
```diff
@@ -0,0 +1,52 @@
...
+	author = {Kim, Baekjun and Lee, Sangwon and Kim, Jihan},
+	title = {Inverse design of porous materials using artificial neural networks},
+	volume = {6},
+	number = {1},
+	elocation-id = {eaax9324},
+	year = {2020},
+	doi = {10.1126/sciadv.aax9324},
+	publisher = {American Association for the Advancement of Science},
+	}
+	eprint = {https://advances.sciencemag.org/content/6/1/eaax9324.full.pdf},
```
samadejacobs, line 49: This link is broken

applications/MOF/MOFae.py
```diff
@@ -0,0 +1,287 @@
...
+input_ = lbann.Input( target_mode = "reconstruction")
+
+
+tensors = lbann.Identity(input_)
+
+tensors = lbann.Reshape(tensors, dims="11 32 32 32")
+# Input tensor shape is  32x32x32x(number_of_atoms)
+
+# Encoder 
+x = lbann.Convolution(tensors,
```
samadejacobs, line 79: Could you use a for loop here to make it more compact since Conv/BN/Activation are repeated a couple times? You could also simplify it as Encoder/Decoder blocks of Conv/BN/Activation.

applications/MOF/MOFae.py
szaman19: deleted in the next commit.

applications/MOF/MOFae.py
szaman19: deleted in the next commit

applications/MOF/MOFae.py
szaman19: Yeah definitely. I was following the LeNet convention to explicitly write out layers but that was mostly so I could explicitly see every layer of the model. I can refactor the code.

applications/MOF/README.md
szaman19: This one? https://advances.sciencemag.org/content/6/1/eaax9324.full.pdf
It seems to work for me.

szaman19:
> LGTM. Please see my request for change in the code. In addition, here are three broad comments:
> (1) I am thinking we could create a material science top directory in applications and this will be one of sub applications, see applications/physics example, @timmoon10 and @bvanessen what do you think?
> (2) While I like leveraging Pytorch a lot, I don't think that dependency is needed for this application. If the data is already in numpy then a numpy data reader (with benefit of parallel ingestion) or python data reader will do. Getting Pytorch on IBM systems (and may be Cori/Corona) may be an issue.
> (3) data is not provided

Thanks so much Sam! 

Fixing up the code changes. About the broad comments,
(2) Yes, the Torch dependency of the initial set up of the data I had but I eventually converted everything to NumPy to get rid of some strange behaviors of pickle and objects. I left the torch dataset in there as an example of extending torch datasets and plugging it into python DataReader. 

(3) We had a little discussion yesterday on the best way/location to make the data available.  I'm not sure about how I can make it available for all the systems so we can run the tests in all the different machines.

applications/MOF/dataset.py
```diff
@@ -0,0 +1,50 @@
...
+##
+## For an example look at: 
+## https://github.com/LLNL/lbann/blob/develop/applications/nlp/transformer/dataset.py
+##
+
+data_dir = os.path.dirname(os.path.realpath(__file__))
+
+## Add CLI arguments for training file location and error handling 
+train_file_path = os.path.join(data_dir, 'data/mofs.npy')  
+test_file_path = os.path.join(data_dir, 'data/mofs.npy')
```
timmoon10, line 16: Is it intentional for the training set and test set to be the same?

samadejacobs:
> This looks good to me. I say we should merge sooner rather than later.
> 
> Re @samadejacobs:
> 
> 1. If there are obvious categories, I don't mind grouping the applications together into things like "physics" or "material science". We shouldn't try too hard to make the categorization scheme consistent though. We don't need to follow the pattern if an application is the only one in its category, if it spans multiple categories, or if the application is less important than other factors (like the data or the algorithms).
> 2. I don't mind the PyTorch dependency since this application is more of a "work in progress". If @szaman19 continues to use this model for his research, he should use whatever hacks and conveniences he can to get good results.
> 3. I also don't mind that the data is only available on Ray. For this kind of app, research convenience is more important than portability.

Well, on 1, the naming as it is now is not descriptive. 
Generally, I agreed this is self contained and should be merge as soon as possible but we are trading off research convenience for standard and maintenance.

applications/MOF/MOFae.py
szaman19: Updated in the last commit to make code more concise

szaman19:
(1) I didn't change the naming as I wasn't sure but that can be moved later after the merge so I'll let ya'll take care of it haha. 
(2) Took out the PyTorch dependence as it really was not that necessary

PR #1573, opened 2020-06-29
Enhance the HDF5 data reader with Distconv

oyamay:
This PR should follow #1572.

This PR applies the changes of [oyamay/lbann/distconv-hdf5-key-names](https://github.com/oyamay/lbann/tree/distconv-hdf5-key-names) to the main branch as well as some catching up commits, mainly to support the 3D U-Net with Distconv. 

This PR
* enhances the HDF5 data reader to use arbitrary data/labels/responses key names
* adds the "label_reconstruction" mode for input layers, which sets the same tensor shape for both data and labels, and calls `fetch_label()` to retrieve the second input tensor
* supports integer labels for cross-entropy layers (https://github.com/LLNL/DiHydrogen/pull/39)


oyamay:
I replace this PR with #1574 as I noticed that I specified a wrong branch.

PR #1575, opened 2020-06-30
Pair RNG with I/O thread work

bvanessen:
Fixed the data reader so that it will dispatch blocks of work to the
I/O thread pool.  The local executing thread no longer does a local
block of work.  The fetch_data_block function now properly renames the
parameter thread_id to block_offset to indicate that it marks which
block of work is being executed.  The fetch_data_block now sets a
thread local variable to index into the new vector of I/O random
number generators.  The I/O random number generators are no long
thread local variables, but there is a global vector of them that is
paired with each block of I/O and transformation work.  The number of
I/O RNGs is setup during initialization to have one per thread in the
I/O pool.

### DIFF
--- a/include/lbann/utils/random_number_generators.hpp
+++ b/include/lbann/utils/random_number_generators.hpp
@@ -56,6 +56,11 @@ fast_rng_gen& get_fast_generator();
 rng_gen& get_data_seq_generator();
 
 /**
+ * Sets the local index for a thread to access the correct I/O RNGs.
+ */
+void set_io_generators_local_index(size_t idx);
+
+/**
  * Return a reference to the global LBANN random number generator used
  * for shuffling the data samples within each mini-batch
  * @note This is stored in a thread_local variable.
@@ -76,7 +81,7 @@ fast_rng_gen& get_fast_io_generator();
  *              into the seed; if not, uses the MPI world rank.
  *
  */
-void init_random(int seed = -1, lbann_comm *comm = nullptr);
+void init_random(int seed = -1, int num_io_RNGs = 1, lbann_comm *comm = nullptr);
 
 /**
  * Initialize a random number generator (with optional seed) that is
@@ -91,10 +96,11 @@ void init_data_seq_random(int seed = -1);
  * Initialize a random number generator (with optional seed) that is
  * specifically used by the I/O threads for tasks such as data
  * preprocessing, etc.
+ * Includes the number of I/O RNGs required.
  *
  * Called from init_random
  */
-void init_io_random(int seed = -1);
+void init_io_random(int seed = -1, int num_io_RNGs = 1);
 
 } // namespace lbann
 
--- a/include/lbann/utils/random_number_generators.hpp
+++ b/include/lbann/utils/random_number_generators.hpp
@@ -55,9 +55,7 @@ fast_rng_gen& get_fast_generator();
  */
 rng_gen& get_data_seq_generator();
 
-/**
- * Sets the local index for a thread to access the correct I/O RNGs.
- */
+/** @brief Sets the local index for a thread to access the correct I/O RNGs. */
 void set_io_generators_local_index(size_t idx);
 
 /**
--- a/include/lbann/utils/random_number_generators.hpp
+++ b/include/lbann/utils/random_number_generators.hpp
@@ -29,11 +29,20 @@
 
 #include "lbann/comm.hpp"
 #include <random>
+#include <mutex>
+#include <thread>
 
 namespace lbann {
 
 using rng_gen = std::mt19937;  // Mersenne Twister
 using fast_rng_gen = std::minstd_rand;  // Minimum standard, LC
+
+struct io_rng_t {
+  lbann::rng_gen io_generator;
+  lbann::fast_rng_gen fast_io_generator;
+  std::unique_ptr<std::mutex> io_mutex;
+  std::thread::id active_thread_id;
+};
 
 /**
  * Return a reference to the global LBANN random number generator.
@@ -56,7 +65,7 @@ fast_rng_gen& get_fast_generator();
 rng_gen& get_data_seq_generator();
 
 /** @brief Sets the local index for a thread to access the correct I/O RNGs. */
-void set_io_generators_local_index(size_t idx);
+io_rng_t& set_io_generators_local_index(size_t idx);
 
 /**
  * Return a reference to the global LBANN random number generator used
--- a/include/lbann/utils/random_number_generators.hpp
+++ b/include/lbann/utils/random_number_generators.hpp
@@ -46,6 +46,22 @@ struct io_rng_t {
   std::thread::id active_thread_id;
 };
 
+struct locked_io_rng_ref {
+  io_rng_t* rng_;
+  std::unique_lock<std::mutex> lock_;
+  locked_io_rng_ref(io_rng_t& rng)
+    : rng_(&rng),
+      lock_(*(rng.io_mutex.get()))
+  {
+    rng_->active_thread_id = std::this_thread::get_id();
+  }
+  explicit operator io_rng_t&() { return *rng_; }
+  ~locked_io_rng_ref() {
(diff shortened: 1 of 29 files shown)


src/utils/random_number_generators.cpp
```diff
@@ -90,26 +92,29 @@ rng_gen& get_data_seq_generator() {
   return ::data_seq_generator;
 }
 
+void set_io_generators_local_index(size_t idx) { ::local_io_generators_index = idx; }
+
 rng_gen& get_io_generator() {
-  if (!::io_generator_inited) {
+  size_t idx = ::local_io_generators_index;
```
benson31, line 98:
```suggestion
  const size_t idx = ::local_io_generators_index;
```

src/utils/random_number_generators.cpp
```diff
@@ -90,26 +92,29 @@ rng_gen& get_data_seq_generator() {
...
+    ::io_generator[idx].seed(hash_combine(::io_generator_seed_base, idx));
+    ::io_generator_inited[idx] = true;
   }
-  return ::io_generator;
+  return ::io_generator[idx];
 }
 
 fast_rng_gen& get_fast_io_generator() {
-  if (!::fast_io_generator_inited) {
+  size_t idx = ::local_io_generators_index;
```
benson31, line 108:
```suggestion
  const size_t idx = ::local_io_generators_index;
```

include/lbann/utils/random_number_generators.hpp
```diff
@@ -55,6 +55,11 @@ fast_rng_gen& get_fast_generator();
  */
 rng_gen& get_data_seq_generator();
 
+/**
+ * Sets the local index for a thread to access the correct I/O RNGs.
+ */
```
benson31, line 60:
```suggestion
/** @brief Sets the local index for a thread to access the correct I/O RNGs. */
```

include/lbann/utils/random_number_generators.hpp
```diff
@@ -29,12 +29,21 @@
...
 
 namespace lbann {
 
 using rng_gen = std::mt19937;  // Mersenne Twister
 using fast_rng_gen = std::minstd_rand;  // Minimum standard, LC
 
+struct io_rng_t {
+  lbann::rng_gen io_generator;
+  lbann::fast_rng_gen fast_io_generator;
+  std::unique_ptr<std::mutex> io_mutex;
```
timmoon10, line 43: Why are we keeping this inside a `unique_ptr`? Do we need the mutex in the heap rather than the stack?

timmoon10, line 43: Pedantic style comment: the `io` in the variable names is redundant with the name of the struct `io_rng_t`.

src/utils/random_number_generators.cpp
```diff
@@ -90,26 +86,35 @@ rng_gen& get_data_seq_generator() {
...
 }
 
+int get_num_io_generators() {
+  return ::io_generators.size();
+}
+
+io_rng_t& set_io_generators_local_index(size_t idx) {
+  ::local_io_generators_index = idx;
+  if (!::io_generators_inited) { LBANN_ERROR("I/O RNG seed not set"); }
+  return ::io_generators[idx];
```
JaeseungYeom, line 96: perhaps use ::io_generators.at(idx) to detect out-of-range error.

JaeseungYeom, line 96: perhaps use ::io_generators.at(idx) to detect out-of-range error.

src/utils/random_number_generators.cpp
```diff
@@ -90,26 +86,35 @@ rng_gen& get_data_seq_generator() {
...
+}
+
 rng_gen& get_io_generator() {
-  if (!::io_generator_inited) {
-    if (!::io_generator_seed_inited) { LBANN_ERROR("I/O RNG seed not set"); }
-    ::io_generator.seed(hash_combine(::io_generator_seed_base,
-                                     std::this_thread::get_id()));
-    ::io_generator_inited = true;
+  const size_t idx = ::local_io_generators_index;
+  io_rng_t& io_rng = ::io_generators[idx];
```
JaeseungYeom, line 101: Same. use ::io_generators.at(idx) to detect out-of-range error.

src/utils/random_number_generators.cpp
```diff
@@ -90,26 +86,35 @@ rng_gen& get_data_seq_generator() {
...
 }
 
 fast_rng_gen& get_fast_io_generator() {
-  if (!::fast_io_generator_inited) {
-    if (!::fast_io_generator_seed_inited) { LBANN_ERROR("Fast I/O RNG seed not set"); }
-    ::fast_io_generator.seed(hash_combine(::fast_io_generator_seed_base,
-                                          std::this_thread::get_id()));
-    ::fast_io_generator_inited = true;
+  const size_t idx = ::local_io_generators_index;
+  io_rng_t& io_rng = ::io_generators[idx];
```
JaeseungYeom, line 110: Same. use ::io_generators.at(idx) to detect out-of-range error.

src/utils/random_number_generators.cpp
```diff
@@ -178,22 +183,23 @@ void init_data_seq_random(int seed) {
...
-  ::io_generator_seed_base = seed;
-  ::io_generator_seed_inited = true;
-  /// Reset the init flag so that generator will reinitialize
-  ::io_generator_inited = false;
-
-  ::fast_io_generator_seed_base = seed;
-  ::fast_io_generator_seed_inited = true;
-  /// Reset the init flag so that generator will reinitialize
-  ::fast_io_generator_inited = false;
+  ::io_generators_inited = true;
```
JaeseungYeom, line 194: move this after the for loop below (line 202)

src/utils/threads/thread_pool.cpp
```diff
@@ -63,20 +63,30 @@ void thread_pool::launch_pinned_threads(
 
   auto error = pthread_getaffinity_np(pthread_self(),
                                       sizeof(cpu_set_t), &cpuset);
+
```
JaeseungYeom, line 66: use strerror(error) instead of error 
for reporting errors with both thread_getaffinity_np() and thread_setaffinity_np()

JaeseungYeom:
It does not hang anymore with multiple threads on catalyst, and the number of threads used looks correct. However, It hangs with a single thread.
Ie accuracy result is still not perfectly reproducible from run to run with IO threads.

JaeseungYeom:
Looks like all the issues are fixed.
It not only generates identical accuracy numbers for the same number of IO threads, but also it does so regardless of the number of IO threads.
It also generates identical results from run to run even using GPUs with DETERMINISTIC=ON set.

src/data_readers/data_reader.cpp
```diff
@@ -78,16 +78,18 @@ void generic_data_reader::setup(int num_io_threads, observer_ptr<thread_pool> io
 }
 
 
-bool lbann::generic_data_reader::fetch_data_block(CPUMat& X, El::Int thread_id, El::Int mb_size, El::Matrix<El::Int>& indices_fetched) {
-  std::string error_message;
-  for (int s = thread_id; s < mb_size; s+=m_io_thread_pool->get_num_threads()) {
+bool lbann::generic_data_reader::fetch_data_block(CPUMat& X, El::Int block_offset, El::Int block_stride, El::Int mb_size, El::Matrix<El::Int>& indices_fetched) {
+  io_rng_t& io_rng = set_io_generators_local_index(block_offset);
+  const std::lock_guard<std::mutex> lock(*(io_rng.io_mutex.get()));
+  io_rng.active_thread_id = std::this_thread::get_id();
```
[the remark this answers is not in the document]
bvanessen: A locked RNG class has been added to encapsulate the locking.

include/lbann/utils/random_number_generators.hpp
bvanessen: Because it cannot be initialized on the stack.

PR #1580, opened 2020-07-02
Enhance the HDF5 data reader with Distconv

oyamay:
This PR should follow #1572.

This PR applies the changes of [oyamay/lbann/distconv-hdf5-key-names](https://github.com/oyamay/lbann/tree/distconv-hdf5-key-names) to the main branch as well as some catching up commits, mainly to support the 3D U-Net with Distconv. 

This PR
* enhances the HDF5 data reader to use arbitrary data/labels/responses key names
* adds the "label_reconstruction" mode for input layers, which sets the same tensor shape for both data and labels, and calls `fetch_label()` to retrieve the second input tensor
* supports integer labels for cross-entropy layers (https://github.com/LLNL/DiHydrogen/pull/39)


bvanessen:
@oyamay I merged the #1572 into develop and then merged develop into this to address merge conflicts.  Can you please re-test this PR and see if it is ready to go from your point of view.

src/layers/io/input/input_layer.cpp
```diff
@@ -42,6 +45,18 @@ input_distconv_adapter(Layer& layer, const bool shuffle_required)
...
+  // TODO: hdf5_reader is assumed to return a sub-sample partitioned
+  // in the same way as specified by the parallel strategy of this input
+  // layer. Other data readers are assumed to return a complete
+  // sample, thus shuffling is required (unless sample-parallel
+  // strategy is given). Conceptually, it seems to make sense if a
+  // data reader is annotated with a parallel strategy. Note that,
+  // when the HDF5 data reader is used, it is assumed that it is used
+  // in all execution modes.
+  auto training_dr = l.get_data_reader(execution_mode::training);
+  m_shuffle_required = dynamic_cast<hdf5_reader<TensorDataType>*>(training_dr) == nullptr;
```
oyamay, line 59: @bvanessen Do we actually need this change? This raises an exception because it tries to get the training execution context, which is set after the setup pass.
I have fixed this issue in 632941e1a754a0445ee9b7265661ca0fbabefcbb (and removed in ed5a8ab98af66e39dafbb620b7bd0b9f3a7ca411) by passing DataReaderMetaData to the adapter which contains the information whether shuffle is required.

bvanessen: No I don't think that we need that.  I reverted that deletion because I thought that it was due to a bad merge.  If it shouldn't be there, please remove it again.

bvanessen (approved):
LGTM

PR #1583, opened 2020-07-07
randomize samples in smiles reader

davidHysom:
Modified so that, when --num_samples is requested, the indices are randomly pulled from all possible indices.

### DIFF
--- a/src/data_readers/data_reader_smiles.cpp
+++ b/src/data_readers/data_reader_smiles.cpp
@@ -135,7 +135,9 @@ void smiles_data_reader::load() {
   //  std::iota(m_shuffled_indices.begin(), m_shuffled_indices.end(), 0);
   //
   if (! opts->get_bool("smiles_random_off")) {
-    LBANN_WARNING("starting to generate some random numbers ... theoretically this could run forever, esp. if you set --num_samples too high.");
+    if (is_master()) {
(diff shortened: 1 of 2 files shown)


src/data_readers/data_reader_smiles.cpp
```diff
@@ -123,13 +125,34 @@ void smiles_data_reader::load() {
...
+  //
+  // Randomly select N out of a possible M indices, where N << M.
+  // Unf, if a user sets N to be close or identical to M, the simple
+  // algorithm could run for a long time.
+  //
+  // This block replaces:
+  //  std::iota(m_shuffled_indices.begin(), m_shuffled_indices.end(), 0);
+  //
+  if (! opts->get_bool("smiles_random_off")) {
+    LBANN_WARNING("starting to generate some random numbers ... theoretically this could run forever, esp. if you set --num_samples too high.");
```
samadejacobs, line 138: Can we let _world master_ spit out this warning.

samadejacobs:
This doesn't work with _num_samples_ and _smiles_random_off_ flags are set. For the example I tested, num_samples=1000000 from enamine 680M dataset. Here is the error:
 src/data_readers/data_reader_smiles.cpp:571 sample_offsets.size()/3: 1 should be equal to m_shuffled_indices.size which is 1000000

davidHysom:
> This doesn't work with ...

I'll check on this. What I know now is that you're not using the data-store. It looks like some recent changes I made broke that mode, so I'll fix and test it.

PR #1588, opened 2020-07-22
Update all files to use Hydrogen 1.4 API

benson31:
This assumes CUDA/cuBLAS.

Testing won't run on this because Hydrogen v1.4.0 is not back-compatible, so it would break everything for everyone else. We should be careful about merging this for this reason.

That said, to force the issue a bit, I'm doing all the Cholesky work in Hydrogen based on v1.4.0 so I don't have to duplicate effort in the cuBLAS/rocBLAS implementation.


include/lbann/layers/transform/concatenate.hpp
```diff
@@ -377,7 +377,7 @@ fp_compute() {
   dc::tensor::Concatenate(this->get_activations(0),
                           this->get_prev_activations(0),
                           this->get_prev_activations(1),
-                          El::GPUManager::Stream());
+                          hydrogen::cuda::GetDefaultStream());
```
bvanessen, line 380: @benson31 Here is the spot where cuda is still used.

include/lbann/layers/transform/concatenate.hpp
```diff
@@ -386,7 +386,7 @@ bp_compute() {
   dc::tensor::Slice(this->get_error_signals(0),
                     this->get_error_signals(1),
                     this->get_prev_error_signals(0),
-                    El::GPUManager::Stream());
+                    hydrogen::cuda::GetDefaultStream());
```
bvanessen, line 389: And here.

include/lbann/layers/transform/concatenate.hpp
benson31: ...what else would we use?

PR #1589, opened 2020-07-27
Set LD_LIBRARY_PATH in Python frontend contrib launchers

timmoon10:
When I build LBANN with Spack and launch a job with the Python frontend, I encounter an error where it can't find shared libraries. The problem is that Spack installs all of LBANN's dependencies in arbitrary places and then manipulates `LD_LIBRARY_PATH` to make them discoverable. The quick fix is to just forward the user's `LD_LIBRARY_PATH` to the LBANN executable.

While we're thinking about this, should we forward any other environment variables? First one off the top of my head is `OMP_NUM_THREADS`.

### DIFF
--- a/python/lbann/contrib/lc/launcher.py
+++ b/python/lbann/contrib/lc/launcher.py
@@ -46,6 +46,9 @@ def make_batch_script(
         if key not in environment:
             environment[key] = os.getenv(key, default)
 
+    # Set environment for shared libraries
+    set_environment('LD_LIBRARY_PATH', os.environ['LD_LIBRARY_PATH'])
+
     # Setup GPU bindings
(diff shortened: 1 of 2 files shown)


timmoon10:
It turns out this PR is not needed since a subprocess inherits its environment from its parent. My bug went away because of a change I made to the environment, not because I explicitly set the environment in `batch.sh`. Even if you submit jobs as a batch script, SLURM will forward the environment (I'm not sure what LSF does).

Also, @benson31 get bad vibes from manipulating `LD_LIBRARY_PATH`.

benson31:
Very bad vibes.

LSF should forward the environment as well.

PR #1592, opened 2020-07-31
Graph Convolution Layers in Python Front End Modules

szaman19:
This adds a few often used Graph Convolution Kernels to the Python Modules 

- Graph Layers: GCN, GIN, Graph, GatedGraph (associated papers are cited in the readme)
- Adds a helper data object, GraphVertexData, in the graph module
    - This streamlines some tensor like operations like row-level access, cloning, and updates
- GCN and Graph layers also have a faster Dense variant which only utilizes Matrix Multiplication
- The layers are tested on two graph datasets (MNIST, and PROTEINS)


python/lbann/modules/graph/sparse/GATConv.py
```diff
@@ -0,0 +1,98 @@
...
+        self.weights = [] 
+
+        for i in range(num_layers):
+            weight_init = lbann.Weights(initializer = lbann.NormalInitializer(mean=0, 
+                                                          standard_deviation = 1/((i+1) * output_channels)))
+            weight_layer = lbann.WeightsLayer(dims = str_list([output_channels, output_channels]),
+                                              weights = weight_init, 
+                                              name = self.name+'_'+str(i)+'_weight',
+                                              data_layout = self.data_layout)
+            self.weights.append(weight_layer)
```
timmoon10, line 55: Don't we only need one `Weights` for this module? We want to apply the same weights at each RNN step.

python/lbann/modules/graph/sparse/GATConv.py
```diff
@@ -0,0 +1,98 @@
...
+        
+
+    def forward(self, X, A):
+        """Call the GatedGraphConv
+        Args:
+            X (GraphVertexData): LBANN Data object, which is a collection of Layers. Each Layer is of
+                                 the shape (1,input_channels) 
+            A (Layer): Adjacency matrix input with shape (num_nodes, num_nodes)
+                                applied. (default: lbann.Relu)
```
timmoon10, line 64:
```suggestion
```

python/lbann/modules/graph/sparse/GATConv.py
```diff
@@ -0,0 +1,98 @@
...
+
+    https://github.com/rusty1s/pytorch_geometric/blob/\
+    master/torch_geometric/nn/conv/gated_graph_conv.py 
+    """
+    global_count = 0
+    def __init__(self, 
+                 output_channels,
+                 num_layers = 1,
+                 name = None,
+                 data_layout = 'data_parallel'):
```
timmoon10, line 22: I'd remove this argument. We only support the data-parallel case for now.

python/lbann/modules/graph/sparse/GATConv.py
szaman19: I was a little confused about whether the same weights were applied at each RNN  step or not, and it seemed like the paper was a bit ambiguous. 

Pytorch Geometric seems to use separate weights for each RNN step. https://github.com/rusty1s/pytorch_geometric/blob/df0612bca76b14464b13c86759eec0c56ebc8154/torch_geometric/nn/conv/gated_graph_conv.py#L47-L72

python/lbann/modules/graph/sparse/GATConv.py
```diff
@@ -0,0 +1,98 @@
+import lbann
```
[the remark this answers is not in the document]
szaman19: No reason, I was working on a different graph layer and forgot to change the file name. Fixed in the next commit.

python/lbann/modules/graph/dense/DenseGCNConv.py
```diff
@@ -0,0 +1,27 @@
...
+        
+        
+        bounds = math.sqrt(6.0 / (input_channels + output_channels))
+        self.weights = lbann.Weights(initializer = lbann.UniformInitializer(min=-bounds,max=bounds),
+                                    name=self.name+'_Weights')
+
+        self.W = lbann.WeightsLayer(dims = str_list([input_channels, output_channels]),
+                               name = self.name+'_layer',
+                               weights =self.weights)
+    def forward(self,X,A):
```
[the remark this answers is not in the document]
szaman19: Added in a few comments saying we assume adjacency matrix is normalized. I tried to think of a way to normalize the matrix on the fly and add a flag, but this seemed a better solution. 

It Could be good to add that capability if/when we write the custom layers.

python/lbann/modules/graph/sparse/GATConv.py
timmoon10: Hm, that's a good point. Following PyTorch Geometric's lead is probably the best call then.

In any case, let's also use the initialization in PyTorch, which is uniform in `[-1/sqrt(output_channels),1/sqrt(output_channels)]`.

szaman19: Fixed. Also, that initialization upped the accuracy on test as a bonus.

PR #1594, opened 2020-07-31
Update superbuild to use Hydrogen 1.4.0 and Aluminum 0.4.0

timmoon10:
This makes changes so `build_lbann_lc.sh` works with the recent version bumps (#1588 and https://github.com/LLNL/Elemental/pull/102).

I'm trying to get LBANN built with Spack, but I'm running into trouble since the new versions are not available in the built-in Spack packages for Hydrogen and Aluminum. Would it be feasible to maintain our own Spack packages for LBANN, Hydrogen, and Aluminum inside of LBANN and copy them to the Spack repo whenever we make changes? That way we can rapidly modify the Spack packages without waiting for a PR to get approved by the Spack team.

### DIFF
--- a/scripts/build_lbann_lc.sh
+++ b/scripts/build_lbann_lc.sh
@@ -330,7 +330,7 @@ fi
 # Load packages
 if [ ${USE_MODULES} -ne 0 ]; then
     module load git
-    module load cmake/3.14.5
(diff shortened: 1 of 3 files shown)


benson31:
Regarding maintaining our own packages, I believe we can do this without much problem.

Issue #1596, opened 2020-08-07
LTFB produces garbage after first two LTFB rounds


timmoon10:
Bug report from @samadejacobs:

> I write to inform you that PR1546 (mixed precision support) broke training with LTFB. I am putting together a unit test to catch this early. In the meantime, I have given you, Brian and Tim a reduced version of my experiment’s scripts (10K JAG dataset, which should in less than 5 minutes with 12 or 16 Lassen nodes). The problem as you will find out is that metrics and objective functions are “nan”.

When I run LTFB with AlexNet (1 Lassen node, 2 or 4 trainers, 1 proc/trainer), the first LTFB round proceeds correctly. During the second LTFB round, some of the trainers have garbage validation accuracy. I can't find a pattern in what trainers fail, although it appears deterministic. By the third LTFB round, all trainers produce garbage.

I can't think of an obvious reason why #1546 would cause this failure. The LTFB callback `dynamic_cast`s all of the weights to `data_type_weights<DataType>`:

https://github.com/LLNL/lbann/blob/2cb48bab4fceeb7fb4ec2b5054cf5fe52755dbaf/src/callbacks/ltfb.cpp#L471

If we're not doing any mixed precision stuff in the model, this should be well-behaved.

timmoon10:
I think I've narrowed things down. Inside the LTFB callback, we call `data_type_weights`'s copy-assignment operator:

https://github.com/LLNL/lbann/blob/9502b2d6a050d4d9834ac638e87a639ee233a2c1/src/callbacks/ltfb.cpp#L523

The weight values are still sensible at this point. However, I see that `weights`'s copy-assignment operator does not copy `std::unordered_map<std::type_index, gradient_manager_ptr> gradients_`:

https://github.com/LLNL/lbann/blob/9502b2d6a050d4d9834ac638e87a639ee233a2c1/src/optimizers/optimizer.cpp#L61

During the next mini-batch step the values turn in to NaNs inside `data_type_optimizer::step()`:

https://github.com/LLNL/lbann/blob/9502b2d6a050d4d9834ac638e87a639ee233a2c1/src/optimizers/data_type_optimizer.cpp#L140

This is even when I set the learning rate to zero. I suspect we are not properly setting the gradient status, so the optimizer receives uninitialized gradients and the optimizer step becomes `0 * NaN -> NaN`. This also explains why some trainers survive an LTFB round: if the uninitialized gradients don't have any NaNs, then the model might still be a state where it can learn.

PR #1599, opened 2020-08-14
Reset data coordinator after each LTFB round

### LINKED ISSUE #1597: Input layer doesn't properly switch between data readers during LTFB
While debugging #1596, I encountered errors with the message:

```
Number of samples: 244 does not match the number of responses: 200
```

The error comes from:

https://github.com/LLNL/lbann/blob/95474f70114ccef94ca6194febf9e566bfaecf43/include/lbann/io/data_buffers/generic_io_buffer.hpp#L62

I suspect the error happens because LTFB switches between execution modes without reseting background fetching in the IO buffers. If I understand correctly, the input layer works like:

```c++
void fp_compute() {
  ++m_buffer_idx;
  m_io_buffers[m_buffer_idx % m_io_buffers.size()].get_data();
  m_io_buffers[(m_buffer_idx+1) % m_io_buffers.size()].fetch_data_in_background();
}
```

If we don't reset the IO buffers when switching execution modes, then the input layer receives data that was fetched asynchronously from the wrong data reader. If an LTFB round happens at step 10 and there are 11 steps/epoch, the 11th training step will expect a partial mini-batch from the training set but receive a full mini-batch from the validation set. This mismatch triggers our error.

Do we already have machinery to reset the IO buffers? During the LTFB round, it seems that the evaluation epochs are correctly using the validation set. It's only when we finish the evaluation and go back to the middle of the training set that we appear to load the wrong data.

I'm not sure if this will help #1596. I wouldn't expect us to get divergence from a few mini-batches with incorrect data, unless it was filled with garbage.

timmoon10:
As discussed in #1597, the data coordinator doesn't properly switch from the validation set to the training set after an LTFB round finishes. I've adjusted the LTFB callback to match how we reset the execution mode in `sgd_training_algorithm`, e.g.:

https://github.com/LLNL/lbann/blob/9fcc5c315a079cf8a49781c22a859d915ea3d5fc/src/training_algorithms/sgd_training_algorithm.cpp#L109

This fixes #1597, but does not affect #1596.

### DIFF
--- a/src/callbacks/ltfb.cpp
+++ b/src/callbacks/ltfb.cpp
@@ -339,7 +339,8 @@ EvalType evaluate(model& m, const std::string& metric_name) {
   m.make_data_store_preloaded(execution_mode::validation);
 
   // Clean up and return metric value
-  c.set_execution_mode(original_mode);
+  m.reset_mode(c, original_mode);
+  c.get_trainer().get_data_coordinator().reset_mode(c);
   return metric_value;
 
 }
--- a/include/lbann/io/data_buffers/generic_io_buffer.hpp
+++ b/include/lbann/io/data_buffers/generic_io_buffer.hpp
@@ -59,9 +59,9 @@ class fetch_data_functor {
       num_responses_fetched = data_reader->fetch_labels(responses);
     }
     if(num_samples_fetched != num_responses_fetched) {
-      std::string err = std::string("Number of samples: ") + std::to_string(num_samples_fetched)
-        + std::string(" does not match the number of responses: ") + std::to_string(num_responses_fetched);
-      throw lbann_exception(err);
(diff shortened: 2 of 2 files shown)


src/callbacks/ltfb.cpp
```diff
@@ -339,7 +339,8 @@ EvalType evaluate(model& m, const std::string& metric_name) {
   m.make_data_store_preloaded(execution_mode::validation);
 
   // Clean up and return metric value
-  c.set_execution_mode(original_mode);
+  m.reset_mode(c, original_mode);
+  c.get_trainer().get_data_coordinator().reset_mode(c);
```
bvanessen, line 343: This should be correct and we do not want to change the execution mode of the execution context as was being done in the previous code.

PR #1600, opened 2020-08-17
Added model_autoencoder_chem_ecfp.py and model_dnn_chem_ecfp.py

graham63:


samadejacobs:
I am tagged to review this, is it ready or still in draft mode? Two things for a start (1) MNIST VAE should be standalone and not part of this PR, implementation should probably go with _vision/mnist_ (2) a unit test would be helpful

graham63:
I’m sure there will be other things I need to fix- I created the PR before I knew I had to create a branch for each one to make separate PRs for them. I’ll move MNIST VAE.

Also, I’ve come across a couple things to ask you about- num_neurons_from_data_reader=True isn’t a valid option anymore. I left it in the python file and commented it out because I didn’t know what to do with it.

I’m also not sure what to do with super_steps in adversarial_model.

What data reader shoud I use for adversarial_model?

Thanks,

-Katie

samadejacobs:
1. If _num_neurons_from_data_reader_ is deprecated, then it is ok to remove it from PFE too.
2. I wouldn't worry about _adversarial_model_, I think it should be deprecated. In its place (in a separate follow up PR) we should have a conventional [DCGAN](https://pytorch.org/tutorials/beginner/dcgan_faces_tutorial.html) with simple MNIST/CIFAR10 dataset or CelebA dataset.

samadejacobs:
I just reviewed PR 1603, how is this different?

PR #1617, opened 2020-09-05
Added a barrier after the conduit node exchange in the data store.

bvanessen:
This eliminates the race condition that is occuring in the VAE ATOM
model with the smiles data reader.  In principle this should not be
required, but there is some odd race between the non-blocking messages
on the CPU and the GPU code.


bvanessen:
@samadejacobs @davidHysom  This should fix the bug with the smiles data reader and the VAE.

bvanessen:
@ndryden Can you think of any reason that the non-blocking calls in the data store would entangle with other code in the code base.  Specifically, if we run the normal GPU code and the pre-patched data store we see occasional hangs.

PR #1620, opened 2020-09-07
Adding explicit data loading.

davidHysom:
Currently, all sample are loaded onto all processors during preloading. This is left over from something I forgot to re-look, however, should not have affected data_exchange. (1) This PR corrects this defficiency. Additionally, it (2) removes the local-cache capability that was unique to the smiles reader (because it is unscalable to larger data sets), and (3) replace it with explicit loading, wherein samples are read from file during fetch_datum.


davidHysom:
no longer relevant; the smiles reader now uses sample lists, which addresses concerns this PR was designed around.

PR #1621, opened 2020-09-07
Updated distconv adaptors to use new optimizer methods

bvanessen:
also fixed bug in how weights were being accessed.


benson31 (approved):
LGTM

PR #1625, opened 2020-09-10
Updated docs, added quick start guide

mrwyattii:
Added a quick start guide for installing, testing, and using LBANN.  This includes a short guide on transitioning from PyTorch to LBANN.  Also made minor fix to build instructions.

### DIFF
--- a/docs/building_lbann.rst
+++ b/docs/building_lbann.rst
@@ -137,12 +137,12 @@ CMake flags known to LBANN's "Superbuild" build system.
     :code:`-e` flag. A full list of options can be viewed with the
     :code:`-h` flag.
 
-2.  Setup the LBANN CMake environment using the Spack environment for
-    the dependencies.
+2.  Setup the LBANN CMake environment using the Spack environment for the
+    dependencies.  If you used a custom Spack environment name in the step
+    above, be sure to specify that with the :code:`-e` option:
 
     .. code-block:: bash
 
-        spack env activate -p lbann
         <path to lbann repo>/scripts/build_lbann_from_source.sh
 
 
--- a/docs/building_lbann.rst
+++ b/docs/building_lbann.rst
@@ -142,6 +142,7 @@ CMake flags known to LBANN's "Superbuild" build system.
 
     .. code-block:: bash
 
+        spack env activate -p lbann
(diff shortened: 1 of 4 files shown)


mrwyattii:
I've added 2 commits based on feedback from @bvanessen and @benson31

docs/building_lbann.rst
```diff
@@ -137,12 +137,12 @@ CMake flags known to LBANN's "Superbuild" build system.
     :code:`-e` flag. A full list of options can be viewed with the
```
graham63, line 137: Shouldn't it say that the -e option allows you to specify the name of your custom environment?

benson31: It is. The diff doesn't show line 136, but that's what that is.

graham63: Ok, I see that, but when you actually try looking at the full list of options using the script it doesn't explain -e. I actually discovered that when I was going through quick_start.rst

Script help menu:
  --help               Display this help message and exit.
  --verbose            Verbose output.
  -d | -deps-only)     Only install the lbann dependencies
  -e | --env           Build and install LBANN in the spack environment provided.
  --half               Enable support for HALF precision data types in Hydrogen and DiHydrogen
  --disable-gpus       Disable GPUS
  -s | --superbuild    Superbuild LBANN with dihydrogen, hydrogen, and aluminum

docs/quick_start.rst
```diff
@@ -0,0 +1,362 @@
...
+    or the :ref:`CMake install <build-with-cmake>`.* To use this installation
+    script, clone the repository and run the script:
+
+    .. code-block:: bash
+
+        git clone https://github.com/llnl/lbann
+        cd ./lbann
+        ./scripts/install_lbann.sh -e lbann
+
+    View other options available by passing the :code:`-h` option to the
```
graham63, line 64: Should the -d option be used here as well? And the -e option should probably be explained so that it's more explicit, and/or the help menu in the script should explain that -e can be used to customize the name of the environment.

bvanessen: Actually, it is correct as written here.  If you don't provide the `-d` flag it will build lbann using that environment.  This is what users would want. The `-d` flag is really for developers.

bvanessen (approved):

PR #1626, opened 2020-09-10
GRU layer

timmoon10:
Our current implementation of GRU is a layer module in the Python frontend. It operates on each time step sequentially, so it requires that input sequences are sliced into time steps and its GPU performance tends to be dominated by the kernel launch overheads from many small kernels.

To improve performance, I've implemented a dedicated GRU layer that calls into cuDNN on GPU. Its API is different from the Python frontend module since it operates on the entire sequence at a time. This helps avoid fine-grained slicing and concatenating. Currently it only supports single-layer, unidirectional GRUs and it has no CPU implementation.  

In addition to the layer, I've added some wrapper classes to the cuDNN header. I found that they made dealing with cuDNN objects much nicer. @mrwyattii might be interested in looking in `include/lbann/utils/cudnn.hpp`/`src/utils/cudnn.cpp` when he starts working on a cuDNN/MIOpen abstraction layer. I've also updated the ATOM VAE model to use the new layer.

The GRU layer passes metric checking and gradient checking on Pascal and Lassen. In some rough experiments with the ATOM VAE model on Pascal, the mini-batch time dropped from 0.26 sec to 0.09 sec. It's not ready to be merged into `develop` yet, but I think we can use this work for some experiments.

### DIFF
--- a/applications/ATOM/models/vae.py
+++ b/applications/ATOM/models/vae.py
@@ -57,6 +57,8 @@ class MolVAE(lbann.modules.Module):
         :return: float, recon component of loss
         """
 
+        x = lbann.Slice(x, slice_points=str_list([0, self.input_feature_dims]))
+        x = lbann.Identity(x)
         x_emb = lbann.Embedding(
             x,
             num_embeddings=self.dictionary_size,
@@ -86,7 +88,10 @@ class MolVAE(lbann.modules.Module):
         :return: float, kl term component of loss
         """
 
-        h = lbann.Constant(value=0.0, num_neurons='256')
+        h = lbann.Constant(
+            value=0.0,
+            num_neurons='256',
+        )
         h = lbann.GRU(
             x_emb, h,
             hidden_size=256,
--- a/applications/ATOM/models/vae.py
+++ b/applications/ATOM/models/vae.py
@@ -122,11 +122,11 @@ class MolVAE(lbann.modules.Module):
         fc = lbann.modules.FullyConnectedModule
         gru = GRUModule
         #Encoder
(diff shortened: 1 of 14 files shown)


timmoon10:
This layer still doesn't have CPU support, but it doesn't break the build on non-CPU systems. I think it's ok to merge.

[Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM301-1).

bvanessen:
@timmoon10 I run into a compiler error: 
```
../../../src/utils/cudnn.cpp: In copy constructor 'lbann::cudnn::RNNDescriptor::RNNDescriptor(const lbann::cudnn::RNNDescriptor&)':
../../../src/utils/cudnn.cpp:581:7: error: 'cudnnGetRNNDescriptor' was not declared in this scope
       cudnnGetRNNDescriptor(
       ^
```

bvanessen:
> @timmoon10 I run into a compiler error:
> 
> ```
> ../../../src/utils/cudnn.cpp: In copy constructor 'lbann::cudnn::RNNDescriptor::RNNDescriptor(const lbann::cudnn::RNNDescriptor&)':
> ../../../src/utils/cudnn.cpp:581:7: error: 'cudnnGetRNNDescriptor' was not declared in this scope
>        cudnnGetRNNDescriptor(
>        ^
> ```

So it turns out that I am using cuDNN v8 which has changed this API. I can try to work through this in the morning unless @timmoon10 wants to take a crack at it.

benson31 (approved):
Seems ok to me. We should merge this quickly so we don't duplicate any more work on the cuDNN descriptors.

PR #1629, opened 2020-09-16
CUDA kernel for tensor reordering in GRU layer

timmoon10:
The GRU layer requires reordering tensors since we operate with `mini_batch_size x sequence_length x dim` data while cuDNN operates with `sequence_length x mini_batch_size x dim`. This PR replaces a bunch of `cudaMemcpy` s with a single CUDA kernel. The kernel API is very general and can be used for general tensor transposes, although the performance is likely poor for other use-cases. It might be something we want to move over to DiHydrogen in the future.

The GRU layer passes metric checking and gradient checking on Lassen. The mini-batch time in the ATOM VAE model dropped from 0.058 sec to 0.040 sec. [Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM303-1).

### DIFF
--- a/applications/ATOM/models/vae.py
+++ b/applications/ATOM/models/vae.py
@@ -314,6 +314,7 @@ class MolVAE(lbann.modules.Module):
         )
         keep_mask = lbann.LogicalNot(ignore_mask)
         length = lbann.Reduction(keep_mask, mode='sum')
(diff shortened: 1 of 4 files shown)


benson31 (approved):
Looks reasonable.

PR #1634, opened 2020-09-21
Add the K-FAC callback

oyamay:
This PR adds a callback for the K-FAC optimization method.
**This PR requires Hydrogen 1.5.0**.

* The K-FAC callback supports fully-connected layers, convolutional layers, and batch-normalization layers after the two types of layers to optimize.
* Biases of fully-connected layers and convolutional layers are yet to be supported.
* A new unit test `test_unit_callback_kfac.py` is added, which checks whether the callback does not fail with a simple network.
* An MNIST example using the K-FAC callback is added.


oyamay:
#1657 should be merged first to make build work with this update.

benson31 (changes requested):
As a whole, this looks reasonable, just a few small changes needed. As a general matter of C++ style, conventionally function parameters that are declared "by value" are not marked `const` in the public API (because it doesn't matter to the caller that a value within the function is `const`), but they are marked `const` in the definition of the function. For example:
```cpp
// foo.hpp
void foo(int a, double b);

// foo.cpp
void foo(const int a, const double b) {
  accumulating_int_ += a;
  accumulating_dbl_ += b;
  // or whatever...
}
```
They are useful in the implementation, since the function parameter list in the functional definition is no different than any other variable you might declare in the scope of the function.

src/callbacks/kfac/kfac.cpp
```diff
@@ -0,0 +1,482 @@
...
+    prof_region_begin(("kfac-inverse/" + block->get_name()).c_str(), prof_color, prof_sync);
+    // TODO: Add kfac_block::is_bn?
+    const bool is_bn = dynamic_cast<kfac_block_bn<Device>*>(block.get()) != nullptr;
+    block->update_kronecker_inverse(
+        comm, m_use_pi,
+        is_bn ? m_damping_bn_act : m_damping_act,
+        is_bn ? m_damping_bn_err : m_damping_err,
+        m_learning_rate_factor,
+        m_print_matrix, m_print_matrix_summary,
+        m_print_time);
```
timmoon10, line 291: If I understand this function correctly, it inverts the Fisher matrix and applies it to precondition the gradient. This is only called once every `m_update_interval` steps. Would it be better to precondition the gradients at every step, even with a stale Fisher matrix? To do this, we should split this function into two parts.

src/callbacks/kfac/kfac.cpp
```diff
@@ -0,0 +1,482 @@
...
+            "allgather_send_buffer",
+            is_buffer_needed.first ? local_buffer_size : 0,
+            1);
+    El::Matrix<DataType, Device>& global_buffer =
+        get_workspace_matrix(
+            "allgather_recv_buffer",
+            is_buffer_needed.second ? global_buffer_size : 0,
+            1);
+    kfac_util::allgather_blocks(
+        buffers, local_buffer, global_buffer, comm, allgather_mode);
```
timmoon10, line 334: We are preconditioning the gradient at the same time and on the same process as we are inverting the Fisher matrix, so it makes sense to allgather the preconditioned gradients to everyone. However, if we want to precondition the gradient at every step (using a stale Fisher matrix), would it be better to perform the preconditioning in parallel? In other words, all processes should have a copy of all the Fisher matrix inverses so that they can apply them to the gradient without needing any communication. The existing scheme is better if the preconditioning is expensive, the Fisher matrices are large, and the gradients are small. Were these your considerations?

timmoon10 (commented):
I don't see any correctness issues, so this would be fine to merge in.

I do see some room for future work though. In particular, we only apply preconditioning one every few steps. Keeping the stale Fisher matrix inverses and applying them every step might improve the learning behavior.

timmoon10 (approved):

src/callbacks/kfac/kfac_util.cpp
```diff
@@ -0,0 +1,514 @@
...
+      El::OrientationNS::NORMAL,
+      El::UnitOrNonUnitNS::NON_UNIT,
+      El::TypeTraits<DataType>::One(),
+      (const El::AbstractMatrix<DataType> &) Ainv,
+      (El::AbstractMatrix<DataType> &) Linv,
+      true);
+  El::Gemm(
+      El::TRANSPOSE, El::NORMAL,
+      El::TypeTraits<DataType>::One(), Linv, Linv,
+      El::TypeTraits<DataType>::Zero(), Ainv);
```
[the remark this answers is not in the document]
oyamay: A part of my implementation may be too straightforward, so if a better implementation is possible, I think that would be better.

src/callbacks/kfac/kfac.cpp
oyamay: I agree that that will improve the performance (I was mainly using `m_update_interval=1` so I haven't investigated that effect in detail).

src/callbacks/kfac/kfac.cpp
oyamay: One of the advantages of distributing Fisher matrix computation among processes is that it can distribute both the computational workload and the memory consumption. For large dense DNNs, such as the ATOM model, each Fisher matrix inverse consumes memory of the same order of magnitude as the number of parameters, which may take few GBs. So I think it is a good idea to optionally remove all-gather for smaller networks, but I think this function is still possible for some large networks that take a large amount of memory.

benson31 (approved):

bvanessen (approved):

PR #1635, opened 2020-09-21
Distconv's convolution and pooling tests

oyamay:
This PR adds or modifies the following unit tests:
* `test_unit_layer_convolution_distconv.py`: Tests 3x3 and 3x3x3 convolution that are currently supported by Distconv. The height dimension is partitioned by 4 GPUs for both 2D and 3D layers.
* `test_unit_layer_pooling_distconv.py`: Tests 3x3 "same" pooling, and 2x2 and 2x2x2 strided pooling that is currently supported by Distconv. The height dimension is partitioned by 4 GPUs for both 2D and 3D layers.
* `test_unit_layer_pooling.py`: Tests various types of 2D and 3D pooling layers. Distconv is not used in this test.

### DIFF
--- a/bamboo/common_python/tools.py
+++ b/bamboo/common_python/tools.py
@@ -758,13 +758,11 @@ def create_tests(setup_func,
             assert 'procs_per_node' not in _kwargs.keys()
             if procs_per_node == "auto":
                 procs_per_node = lbann.contrib.lc.systems.gpus_per_node()
+                if procs_per_node == 0:
+                    e = 'this test requires GPUs.'
+                    print('Skip - ' + e)
+                    pytest.skip(e)
             _kwargs['procs_per_node'] = procs_per_node
-
-        # FIXME
-        if "procs_per_node" in _kwargs.keys():
-            print("procs_per_node={}".format(_kwargs["procs_per_node"]))
-        else:
-            print("no procs_per_node")
 
         # If the user provided a suffix for the work directory, append it
         if 'work_subdir' in _kwargs:
--- a/bamboo/common_python/tools.py
+++ b/bamboo/common_python/tools.py
@@ -662,7 +662,6 @@ def assert_failure(return_code, expected_error, error_file_name):
 
 def create_tests(setup_func,
                  test_file,
-                 procs_per_node=None,
                  test_name_base=None,
                  **kwargs):
     """Create functions that can interact with PyTest
@@ -689,9 +688,6 @@ def create_tests(setup_func,
             lbann.reader_pb2.DataReader, lbann.Optimizer)`.
         test_file (str): Python script being run by PyTest. In most
             cases, use `__file__`.
-        procs_per_node (int or str, optional): The number of processes
-            per node to pass into `lbann.contrib.launcher.run`.
-            If "auto" is set, gpus_per_node() is passed instead.
         test_name (str, optional): Descriptive name (default: test
             file name with '.py' removed).
         **kwargs: Keyword arguments to pass into
@@ -713,7 +709,7 @@ def create_tests(setup_func,
         # Make sure test name is prefixed with 'test_'
         test_name_base = 'test_' + test_name_base
 
-    def test_func(cluster, executables, dir_name, procs_per_node, compiler_name):
+    def test_func(cluster, executables, dir_name, compiler_name):
         """Function that can interact with PyTest.
 
         Returns a dict containing log files and other output data.
@@ -752,17 +748,6 @@ def create_tests(setup_func,
                                                'experiments',
                                                test_name)
 
-        # Setup the number of processes
-        if procs_per_node is not None:
-            assert 'procs_per_node' not in _kwargs.keys()
-            if procs_per_node == "auto":
-                procs_per_node = gpus_per_node(lbann)
-                if procs_per_node == 0:
-                    e = 'this test requires GPUs.'
-                    print('Skip - ' + e)
-                    pytest.skip(e)
-            _kwargs['procs_per_node'] = procs_per_node
-
         # If the user provided a suffix for the work directory, append it
         if 'work_subdir' in _kwargs:
             _kwargs['work_dir'] = os.path.join(_kwargs['work_dir'], _kwargs['work_subdir'])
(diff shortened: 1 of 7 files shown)


bamboo/common_python/tools.py
```diff
@@ -748,6 +753,17 @@ def test_func(cluster, executables, dir_name, compiler_name):
                                                'experiments',
                                                test_name)
 
+        # Setup the number of processes
+        if procs_per_node is not None:
+            assert 'procs_per_node' not in _kwargs.keys()
+            if procs_per_node == "auto":
+                procs_per_node = lbann.contrib.lc.systems.gpus_per_node()
```
bvanessen, line 760: @timmoon10 should we change this to `lbann.contrib.systems.gpus_per_node()` so that any center gets it?

bamboo/unit_tests/test_unit_layer_identity_distconv.py
```diff
@@ -85,7 +85,8 @@ def construct_model(lbann):
     x = x_lbann
     x = lbann.Reshape(x, dims="4 4 3")
     y = lbann.Identity(x, data_layout='data_parallel',
-                       parallel_strategy=create_parallel_strategy(4))
+                       parallel_strategy=create_parallel_strategy(
+                           lbann.contrib.lc.systems.gpus_per_node()))
```
bvanessen, line 89: Similarly for here, I know that we are only testing on LLNL systems, but this should be generalizable.

bamboo/unit_tests/test_unit_layer_leaky_relu_distconv.py
```diff
@@ -90,7 +90,8 @@ def construct_model(lbann):
     x = lbann.Reshape(x, dims="4 2 6")
     y = lbann.LeakyRelu(x, negative_slope=0.01,
                         data_layout='data_parallel',
-                        parallel_strategy=create_parallel_strategy(4))
+                        parallel_strategy=create_parallel_strategy(
+                            lbann.contrib.lc.systems.gpus_per_node()))
```
bvanessen, line 94: and here.

bamboo/unit_tests/test_unit_layer_leaky_relu_distconv.py
```diff
@@ -121,7 +122,8 @@ def construct_model(lbann):
     x = lbann.Reshape(x, dims="4 2 6")
     y = lbann.LeakyRelu(x, negative_slope=2,
                         data_layout='model_parallel',
-                        parallel_strategy=create_parallel_strategy(4))
+                        parallel_strategy=create_parallel_strategy(
+                            lbann.contrib.lc.systems.gpus_per_node()))
```
bvanessen, line 126: plus here.

bamboo/unit_tests/test_unit_layer_relu_distconv.py
```diff
@@ -89,7 +89,8 @@ def construct_model(lbann):
     x = x_lbann
     x = lbann.Reshape(x, dims="4 2 6")
     y = lbann.Relu(x, data_layout='data_parallel',
-                   parallel_strategy=create_parallel_strategy(4))
+                   parallel_strategy=create_parallel_strategy(
+                       lbann.contrib.lc.systems.gpus_per_node()))
```
bvanessen, line 93: Here as well

bamboo/unit_tests/test_unit_layer_relu_distconv.py
```diff
@@ -119,7 +120,8 @@ def construct_model(lbann):
     x = x_lbann
     x = lbann.Reshape(x, dims="4 2 6")
     y = lbann.Relu(x, data_layout='model_parallel',
-                   parallel_strategy=create_parallel_strategy(4))
+                   parallel_strategy=create_parallel_strategy(
+                       lbann.contrib.lc.systems.gpus_per_node()))
```
bvanessen, line 124: here.

bvanessen (changes requested):
LGTM.  I think that we can make the gpus per node call general.

bamboo/common_python/tools.py
oyamay: I fixed it so that it recognizes other systems that are registered in the Python frontend.

timmoon10 (changes requested):
It will be nice to finally have a test for the pooling layer.

I don't have a good feeling about the `procs_per_node` stuff in `tools.py`. I'd like to revert that file (keeping the `gpus_per_node` function). Inside the distconv tests that require GPUs, we should check if `gpus_per_node() > 0` and otherwise call `pytest.skip()`. That seems to be the appropriate place to make this check, since we could imagine that other layers will skip for different reasons (e.g. if it requires AMD).

bamboo/unit_tests/test_unit_layer_convolution_distconv.py
```diff
@@ -153,126 +154,81 @@ def construct_model(lbann):
...
+            execution_modes='test'))
 
     # ------------------------------------------
     # Gradient checking
     # ------------------------------------------
 
-    callbacks.append(lbann.CallbackCheckGradients(error_on_failure=True))
+    callbacks.append(lbann.CallbackCheckGradients(
+        error_on_failure=True,
+        step_size=20.0, # TODO: Use the default step size.
```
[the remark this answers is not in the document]
oyamay: I found the reason that the pooling test fails on the check gradient callback is the Hybrid tensor shuffler but not the step size, so I reverted that point and tentatively changed the backend to Aluminum, and also created an issue about it (#1659). Now both convolution and pooling tests should succeed.

bamboo/common_python/tools.py
```diff
@@ -688,6 +689,9 @@ def create_tests(setup_func,
             lbann.reader_pb2.DataReader, lbann.Optimizer)`.
         test_file (str): Python script being run by PyTest. In most
             cases, use `__file__`.
+        procs_per_node (int or str, optional): The number of processes
+            per node to pass into `lbann.contrib.launcher.run`.
+            If "auto" is set, gpus_per_node() is passed instead.
```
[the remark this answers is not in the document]
oyamay: I restored this point too.

timmoon10:
Looks good to me! Is this ready to merge?

oyamay:
@timmoon10 Yes, it works on our GPU clusters.

timmoon10 (approved):

PR #1636, opened 2020-09-21
externals-linux-rhel7-power8le.sh fix for Ray

szaman19:
- Added spacing fix on externals-linux-rhel7-power8le.sh 
- Reverted the CUDA version from 10.2.89 to 10.1.168 to work with the CUDNN version.


bvanessen (changes requested):
This PR has a bunch of unnecessary commits on it that concern me.  I will create a clean one.

PR #1637, opened 2020-09-23
Update reconstruction tests to use new sample lists format

bvanessen:
Update the test to use the new sample list format and adjusted the
testing framework to allow for overriding the sample list flags.  Also
tweaked the launch parameters on Lassen to match the rest of the PFEs
launch values.


bvanessen:
@timmoon10 @benson31 I have reverted the expected values back to what we saw.  I am perplexed because these work on pascal, but on Lassen the reconstruction value is ~10.x  and I am not sure why they differ.

timmoon10 (approved):
Makes sense to me. If there is a different value on Lassen, I suspect the data is different or there is a bug in the sample lists. In either case, failure would be the correct behavior.

bvanessen:
Okay, I am going to merge this since it fixes the launch issue and we can have a subsequent PR to ensure that the learning behavior matches.  I will also check that the sample lists are identical.

PR #1638, opened 2020-09-24
Refactor layers to use the cuDNN handle wrappers

benson31:
PR #1626 added some cuDNN handle wrappers, but only used them in GRU layers. This adds the rest of the required set for LBANN and pushes their use throughout LBANN layers.

### DIFF
--- a/src/utils/cudnn.cpp
+++ b/src/utils/cudnn.cpp
@@ -851,7 +851,7 @@ PoolingDescriptor::~PoolingDescriptor()
   }
   catch (std::exception const& e)
   {
-    std::cerr << "Caught exception in ~ActivationDescriptor(). Shutting down."
+    std::cerr << "Caught exception in ~PoolingDescriptor(). Shutting down."
               << std::endl;
     std::terminate();
   }
@@ -974,7 +974,7 @@ LRNDescriptor::~LRNDescriptor()
   }
   catch (std::exception const& e)
   {
-    std::cerr << "Caught exception in ~ActivationDescriptor(). Shutting down."
+    std::cerr << "Caught exception in ~LRNDescriptor(). Shutting down."
               << std::endl;
     std::terminate();
   }
--- a/src/utils/cudnn.cpp
+++ b/src/utils/cudnn.cpp
@@ -198,21 +198,12 @@ void TensorDescriptor::set(
       "and strides (",strides_in.size(),")");
   }
 
-  std::vector<int> dims, strides;
-  if (dims_in.size() < 3)
+  std::vector<int> dims=std::move(dims_in), strides=std::move(strides_in);
+  if (dims.size() < 3)
   {
     dims.resize(3, 1);
-    std::copy(dims_in.rbegin(), dims_in.rend(), dims.rbegin());
-    if (!strides_in.empty())
-    {
-      strides.resize(3, dims_in.front() * strides_in.front());
-      std::copy(strides_in.rbegin(), strides_in.rend(), strides.rbegin());
-    }
-  }
-  else
-  {
-    dims = std::move(dims_in);
-    strides = std::move(strides_in);
+    if (strides.size() < 3)
+      strides.resize(3, 1);
   }
 
   // Assume data is contiguous if no strides are provided
--- a/src/utils/cudnn.cpp
+++ b/src/utils/cudnn.cpp
@@ -306,7 +306,7 @@ void TensorDescriptor::reset(cudnnTensorDescriptor_t desc) {
   desc_ = desc;
 }
 
-cudnnTensorDescriptor_t TensorDescriptor::release() {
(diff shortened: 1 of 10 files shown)


src/utils/cudnn.cpp
```diff
@@ -328,23 +184,26 @@ void TensorDescriptor::create() {
...
+      "with mismatched dimensions (", dims_in.size(), ") ",
+      "and strides (",strides_in.size(),")");
+  }
+
+  std::vector<int> dims=std::move(dims_in), strides=std::move(strides_in);
+  if (dims.size() < 3)
+  {
+    dims.resize(3, 1);
+    if (strides.size() < 3)
+      strides.resize(3, 1);
```
timmoon10, line 206: It would be somewhat better to prepend the ones.

src/utils/cudnn.cpp
```diff
@@ -681,82 +540,486 @@ void RNNDescriptor::set(
...
+{
+  try
+  {
+    this->reset();
   }
-  for (size_t i = 0; i < m_prev_error_signals.size(); ++i) {
-    copy_tensor_desc(other.m_prev_error_signals[i], m_prev_error_signals[i]);
+  catch (std::exception const& e)
+  {
+    std::cerr << "Caught exception in ~ActivationDescriptor(). Shutting down."
```
timmoon10, line 808:
```suggestion
    std::cerr << "Caught exception in ~PoolingDescriptor(). Shutting down."
```

src/utils/cudnn.cpp
```diff
@@ -681,82 +540,486 @@ void RNNDescriptor::set(
...
+{
+  try
+  {
+    this->reset();
   }
-  for (auto&& desc : m_prev_error_signals) {
-    if (desc != nullptr) { cudnnDestroyTensorDescriptor(desc); }
+  catch (std::exception const& e)
+  {
+    std::cerr << "Caught exception in ~ActivationDescriptor(). Shutting down."
```
timmoon10, line 931:
```suggestion
    std::cerr << "Caught exception in ~LRNDescriptor(). Shutting down."
```

timmoon10 (approved):
LGTM

`layer_tensor_manager` delenda est.

src/utils/cudnn.cpp
benson31: The issue is that the first dim is the size, so it's "inter-pending". Suppose `dims_in = {7,5}`. Do you have 7 samples with 5 1x1 feature maps (this would match `develop`, or do you have 7 samples with 1 1x5 (or 5x1?) feature maps? As I mentioned, `develop` will extend this to `{7,5,1,1}`, so I was trying to match that behavior. (At least I got rid of one of the extraneous 1s, so it's just `{7,5,1}` after this PR!)

src/layers/learning/gru.cpp
```diff
@@ -207,7 +207,8 @@ void gru_layer<TensorDataType, Layout, Device>::setup_gpu() {
   // RNN descriptor
   size_t dropout_state_size;
   CHECK_CUDNN(cudnnDropoutGetStatesSize(handle, &dropout_state_size));
-  cudnn::DropoutDescriptor dropout_desc(0.f, nullptr, dropout_state_size, 0);
+  cudnn::DropoutDescriptor dropout_desc;
+  dropout_desc.set(0.f, nullptr, dropout_state_size, 0);
```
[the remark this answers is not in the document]
benson31: On the one hand, that's fair that this will only be available in cuDNN-land for now, but my understanding is that this code will still need to run on HIP-based machines in the future. Is there harm in doing the modifications that we can do now?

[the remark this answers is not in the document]
timmoon10: Most of the changes would be overwritten in the other PR since we're using the new RNN API in cuDNN 8.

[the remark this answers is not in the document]
benson31: Ah, got it. Will revert prior to merge.

src/utils/cudnn.cpp
timmoon10: Hm, matching existing behavior in `develop` is sensible. I'm also thinking about passing the data as-is into cuDNN and letting it throw an error, since it's really an internal problem in cuDNN's implementation.

benson31: Unless we have a way to open a ticket with them and complain, I'm not in favor of throwing the error. This affects models as simple as Lenet, that we have to be able to run. At one level or another, this is something we need to deal with. Since it's a fiddly bit that's specific to cuDNN, I think it belongs here, far away from the actual layer implementation, which ought not concern itself with cuDNN at all.

src/layers/learning/gru.cpp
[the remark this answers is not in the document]
benson31: Done.

bvanessen (approved):
On a quick scan LGTM

PR #1641, opened 2020-09-27
Optimize performance of GRU layer

timmoon10:
This PR makes three changes to the GPU GRU layer implementation:
- Update to use cuDNN 8 API for RNNs.
- Support stacked RNNs.
- Use CUDA graphs to reduce kernel launch overheads.

I've also added wrapper classes for CUDA graph objects. Note that the GRU layer now requires CUDA 11.0 and cuDNN 8.0.4, although LBANN itself can still be built with older versions.

I see a 1.4x speedup in the ATOM VAE model on 1 Lassen node with a 512 mini-batch size. Profiling suggests we are mostly compute-bound, at least within the GRU layer. The GRU layer passes metric checking and gradient checking on Lassen. [I'm running into snafus on Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM305-1), although I need to figure out whether it's a real problem or if my build is messed up.


timmoon10:
I don't see new failures in [Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM305-1).

bvanessen (approved):

PR #1642, opened 2020-09-28
Feature port CUDA API to Hydrogen for ROCm support

mrwyattii:
This PR does the following:
- Port raw CUDA API calls to new Hydrogen API calls
- implement `gpu` and `gpu_lib` namespace aliases to abstract away cuda/rocm api calls (and use them to replace calls to methods in the `cuda::` namespace)
- Replace CUDA `streams` with `multisync` and `sync_info` where possible
- Update code comments to reflect these changes
- Add "TODO: HIP/ROCM notes" where future comments and documentation need to be added

Things that need to be addressed:
- `distconv` related code

### DIFF
--- a/src/layers/misc/channelwise_softmax.cu
+++ b/src/layers/misc/channelwise_softmax.cu
@@ -32,14 +32,14 @@ namespace lbann {
 
 namespace {
 
-using Size3 = cuda::array<size_t,3>;
+using Size3 = gpu_lib::array<size_t,3>;
 
 /** @brief Max functor */
 template <class T>
 struct max_op {
   __device__ __forceinline__
   DataType operator()(const T& x1, const T& x2) const {
-    return cuda::max(x1, x2);
+    return gpu_lib::max(x1, x2);
   }
 };
 
@@ -87,16 +87,16 @@ __global__ void fp_max_kernel(
     for (size_t j = gidy; j < vals_dims[1]; j += nthreadsy) {
 
       // Find largest value for each thread
-      TensorDataType maxval{-cuda::infinity<TensorDataType>()};
+      TensorDataType maxval{-gpu_lib::infinity<TensorDataType>()};
       for (size_t i = gidx; i < vals_dims[2]; i += nthreadsx) {
         const auto& val = vals_buffer[k * vals_strides[0]
                                       + j * vals_strides[1]
                                       + i * vals_strides[2]];
-        maxval = cuda::max(maxval, val);
+        maxval = gpu_lib::max(maxval, val);
       }
 
       // Find largest value for each block
-      maxval = cuda::block_reduce<bdimx,bdimy,bdimz,TensorDataType,max_op<TensorDataType>>(maxval);
+      maxval = gpu_lib::block_reduce<bdimx,bdimy,bdimz,TensorDataType,max_op<TensorDataType>>(maxval);
       if (tid == 0) {
         const auto& pos = (k * maxvals_strides[0]
                            + j * maxvals_strides[1]
@@ -149,13 +149,13 @@ __global__ void fp_denom_kernel(
         const auto& x = input_buffer[k * input_strides[0]
                                      + j * input_strides[1]
                                      + i * input_strides[2]];
-        denom += cuda::exp(x-shift);
+        denom += gpu_lib::exp(x-shift);
       }
 
       // Compute contribution from each block
-      denom = cuda::block_reduce<bdimx,bdimy,bdimz>(denom);
+      denom = gpu_lib::block_reduce<bdimx,bdimy,bdimz>(denom);
       if (tid == 0) {
-        cuda::atomic_add(&denoms[j+k*input_dims[1]], denom);
+        gpu_lib::atomic_add(&denoms[j+k*input_dims[1]], denom);
       }
 
     }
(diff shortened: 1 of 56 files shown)


src/layers/transform/concatenate.cu
```diff
@@ -283,16 +282,13 @@ void fp_compute_impl(
...
-    l.m_workspace.size(),
-    sync_info);
-  unsigned char* device_workspace_ptr = device_workspace.data();
-  cudaMemcpyAsync(device_workspace_ptr,
-                  l.m_workspace.data(),
-                  l.m_workspace.size(),
-                  cudaMemcpyHostToDevice,
-                  stream);
-  l.m_workspace_event.record(stream);
+  gpu_lib::thrust::vector<unsigned char> device_workspace(l.m_workspace.size());
```
timmoon10, line 285: `hydrogen::simple_buffer` is usually preferable to `thrust::vector`. Thrust does a lot of synchronization internally, so we should only use it when we need it (mostly for sorting).

src/layers/transform/slice.cu
```diff
@@ -276,16 +275,13 @@ void fp_compute_impl(
...
-    l.m_workspace.size(),
-    sync_info);
-  unsigned char* device_workspace_ptr = device_workspace.data();
-  cudaMemcpyAsync(device_workspace_ptr,
-                  l.m_workspace.data(),
-                  l.m_workspace.size(),
-                  cudaMemcpyHostToDevice,
-                  stream);
-  l.m_workspace_event.record(stream);
+  gpu_lib::thrust::vector<unsigned char> device_workspace(l.m_workspace.size());
```
timmoon10, line 278: Prefer `hydrogen::simple_buffer` to `thrust::vector`.

timmoon10 (changes requested):
Mostly good, couple typos. Have we run this through Bamboo?

Also, does `hydrogen::gpu:LaunchKernel` cast the input arguments to match the kernel's signature? If I am interpreting he head of the `hydrogen` branch correctly, we just pass the argument pointers to `cudaLaunchKernel`:

https://github.com/LLNL/Elemental/blob/0a94d35a9602781b0b74150a705206d945b3a1d7/include/hydrogen/device/gpu/cuda/CUDALaunchKernel.hpp#L24

I'd expect this to run into problems at runtime if there is a type mismatch.

src/layers/regularizers/batch_normalization.cu
```diff
@@ -484,38 +480,42 @@ void batch_normalization_layer<TensorDataType, T_layout, Dev>::fp_compute() {
...
-          local_width,
-          num_channels,
-          channel_size,
-          local_input.LockedBuffer(), local_input.LDim(),
-          local_mean.Buffer(),
-          local_var.Buffer());
+      hydrogen::gpu::LaunchKernel(
+        fp_sums_kernel<TensorDataType, block_size>,
+        grid_dims, block_dims, 0, multisync,
+        channel_size, local_width,
```
benson31, line 494:
```suggestion
        local_width, num_channels, channel_size,
```

src/layers/misc/channelwise_softmax.cu
```diff
@@ -438,21 +445,28 @@ void bp_impl(size_t num_channels,
   // dot(y,dL/dy)
   LocalMat local_y_dot_dy(num_channels, local_mini_batch_size);
   El::Zero(local_y_dot_dy);
+
+  auto multisync = El::MakeMultiSync(gpu::get_sync_info(local_input_grad),
+                                     gpu::get_sync_info(local_output_grad),
+                                     gpu::get_sync_info(local_output),
+                                     gpu::get_sync_info(local_y_dot_dy));
```
benson31, line 452: In this case, `local_y_dot_dy` must come first, because `local_input_grad` doesn't participate in the first kernel below. In a futuristic multistream world, this could cause issues. Of course, `local_y_dot_dy` could, in that universe, be setup to have the right syncinfo so this could never happen.
```suggestion
  auto multisync = El::MakeMultiSync(gpu::get_sync_info(local_y_dot_dy),
                                     gpu::get_sync_info(local_output_grad),
                                     gpu::get_sync_info(local_output),
                                     gpu::get_sync_info(local_input_grad));
```

benson31 (changes requested):
Just a few minor things.

benson31 (approved):
LGTM

src/layers/transform/concatenate.cu
mrwyattii: addressed with ba62dcb06925539311074dd079b5e52ff46524ff

src/layers/transform/slice.cu
mrwyattii: addressed with ba62dcb06925539311074dd079b5e52ff46524ff

benson31:
@timmoon10 IMO, this is ready to go.

> Also, does hydrogen::gpu:LaunchKernel cast the input arguments to match the kernel's signature?

No, but that's not this PR's problem. I have a fix that I'll work into Hydrogen soon. I _believe_ that HIP does validate the parameter types.

timmoon10:
If `hydrogen::gpu:LaunchKernel` doesn't properly cast the argument types, won't this break all of the CUDA kernels?

timmoon10 (approved):
Once https://github.com/LLNL/Elemental/pull/111 merges and Bamboo passes, I'd say this is ready to merge.

benson31:
This is passing as much of Bamboo as `develop`. Merging.

PR #1644, opened 2020-09-29
Updated Hydrogen to v1.5.0 and Aluminum to v0.5.0

bvanessen:
Update LBANN to use the latest version of Hydrogen and Aluminum.


scripts/install_lbann.sh
```diff
@@ -262,23 +262,23 @@ ${STD_PACKAGES}
     aluminum:
       buildable: true
       version:
-      - 0.4.0
+      - master
```
mrwyattii, line 265: Do we rely on commits to aluminum since release v0.5.0?  If not, would `0.5.0` be more appropriate here?

bvanessen: Yes we do unfortunately.

PR #1649, opened 2020-10-03
Gb integration branch

bvanessen:


bvanessen:
So far this PR include #1641 and #1646

PR #1651, opened 2020-10-03
Feature rng ltfb no sync

bvanessen:
Use dedicated RNGs to compute the LTFB tournament partners rather than rank 0 computing and scattering it.  This will remove the global sync at the start of every tournament.  The new RNG is seeded off of the root seed and is identical across trainers.


timmoon10 (changes requested):
When my suggestion is merged, I don't see any correctness issues. That said, I don't think this is the right approach. The generator is logically a private member of the LTFB callback (and once we have an LTFB training algorithm, it belongs in `ltfb_execution_context`).

bvanessen:
@timmoon10 I agree that as soon as we have LTFB as a training algorithm, not a callback, it would be better to implement the RNG as a private variable.

PR #1652, opened 2020-10-03
Feature smiles dr hack

bvanessen:


bvanessen:
While this is a hack it is suitable to be merged until the smiles data reader is rewritten.

PR #1655, opened 2020-10-04
Disable global statistics

bvanessen:
Updated the print_statistics and timer callbacks so that global
statistics are not computed by default.  This removes several
inter-trainer collectives that impact LTFB performance.  There is now
a command line argument --ltfb_allow_global_statistics and an
environment variable LBANN_LTFB_ALLOW_GLOBAL_STATISTICS that enable
the collectives and computation plus reporting of global statistics.


ndryden (approved):
Looks fine, but should watch out that this could cause output interleaving, making things hard to read, if you don't redirect each rank's output to a separate file.

bvanessen:
@ndryden I agree that this will increase output interleaving.  We are still trying to balance the best way to make sure that everything is reported and that the output is still manageable.

timmoon10 (approved):

PR #1666, opened 2020-10-14
Add autoencoder pilot1 front end

graham63:


applications/autoencoders/data/pilot1/data_reader_candle_pilot1.prototext
```diff
@@ -0,0 +1,25 @@
+data_reader {
+  reader {
+    name: "csv"
```
mrwyattii, line 3: perhaps "P1B3" would be a more appropriate name?

timmoon10: I think this is right. The fields in the data reader don't have the best names. A better name would be "type", so having a CSV data reader is what we want.

applications/autoencoders/data/pilot1/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    protobuf_file = os.path.join(data_dir, data_reader_file)
+    message = lbann.lbann_pb2.LbannPB()
+    with open(protobuf_file, 'r') as f:
+        google.protobuf.text_format.Merge(f.read(), message)
+    message = message.data_reader
+
+    # Set paths
+    for reader in message.reader:
+        reader.data_filedir = data_dir
+
```
timmoon10, line 25: This isn't necessary since we aren't downloading the data into this directory. 
```suggestion
```

applications/autoencoders/data/pilot1/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+import google.protobuf.text_format
+import lbann
+
+# Paths
+data_dir = os.path.dirname(os.path.realpath(__file__))
+
+# FIXME: Should I add anything to check paths like in cifar10, or
+# download anything like in mnist?
+
```
timmoon10, line 12:
```suggestion
```
I think we can assume the data is where it's supposed to be. Since it's not a common dataset, it'll be on the user to get it right.

applications/autoencoders/autoencoder_candle_pilot1/model_dnn_chem_ecfp.py
```diff
@@ -0,0 +1,183 @@
...
+    help='scheduler job name (default: lbann_CANDLE_pilot1)')
+parser.add_argument(
+    '--mini-batch-size', action='store', default=128, type=int,
+    help='mini-batch size (default: 128)', metavar='NUM')
+parser.add_argument(
+    '--num-epochs', action='store', default=4, type=int,
+    help='number of epochs (default: 4)', metavar='NUM')
+parser.add_argument(
+    '--random-seed', action='store', default=0, type=int,
+    help='random seed for LBANN RNGs', metavar='NUM')
```
timmoon10, line 64: This doesn't seem to be used:
```suggestion
```

applications/autoencoders/autoencoder_candle_pilot1/model_autoencoder_chem_ecfp.py
```diff
@@ -0,0 +1,244 @@
...
+    help='scheduler job name (default: lbann_CANDLE_pilot1)')
+parser.add_argument(
+    '--num-epochs', action='store', default=4, type=int,
+    help='number of epochs (default: 4)', metavar='NUM')
+parser.add_argument(
+    '--mini-batch-size', action='store', default=128, type=int,
+    help='mini-batch size (default: 128)', metavar='NUM')
+parser.add_argument(
+    '--random-seed', action='store', default=0, type=int,
+    help='random seed for LBANN RNGs', metavar='NUM')
```
timmoon10, line 64: This doesn't seem to be used:
```suggestion
```

timmoon10 (approved):
After merging my suggestions, I say it's ready to go.

benson31 (approved):

bvanessen (changes requested):
Overall this looks good, but shouldn't the PR deleted the model_dnn_chem_ecfp.prototext file since it is subsumed by the PFE file?

applications/autoencoders/data/pilot1/data_reader_candle_pilot1.prototext
mrwyattii: oh, that is confusing - especially because other application examples don't use that field as "type": https://github.com/LLNL/lbann/blob/develop/applications/vision/data/mnist/data_reader.prototext

mrwyattii (approved):

graham63:
> Overall this looks good, but shouldn't the PR deleted the model_dnn_chem_ecfp.prototext file since it is subsumed by the PFE file?

Probably- I'm just scared to delete things 😂

PR #1668, opened 2020-10-15
Enable mini-batch size smaller than number of ranks

bvanessen:
Disable check for samples that will throw an unnecessary error on
ranks that will not receive samples because the mini-batch size is too
small.


timmoon10:
Closed by https://github.com/LLNL/lbann/pull/1707.

PR #1671, opened 2020-10-17
Fixed the resource launcher on lassen for the tests to set the correct

bvanessen:
number of tasks.

### DIFF
--- a/bamboo/common_python/tools.py
+++ b/bamboo/common_python/tools.py
@@ -241,13 +241,13 @@ def get_command(cluster,
                 option_gpu_per_resource = ' --gpu_per_rs ALL_GPUS'
                 option_launch_distribution = ' --launch_distribution packed'
                 # Avoid `nrs (32) should not be greater than rs_per_host (1) * number of servers available (16).`
-                # By default there should be 4 prcesses per node (especially when using GPUs)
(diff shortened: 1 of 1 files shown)


bamboo/common_python/tools.py
```diff
@@ -241,15 +241,21 @@ def get_command(cluster,
...
                 # Avoid `nrs (32) should not be greater than rs_per_host (1) * number of servers available (16).`
+                # By default there should be 4 prcesses per node (especially when using GPUs)
+                resources_per_node = 4
                 if num_nodes is None:
-                    num_nodes = 1
+                    num_nodes = math.ceil(float(num_processes)/4)
+                else:
+                    if disable_cuda == True:
+                        # When CUDA is disabled, allow the number of resources per node to be overridden
+                        resources_per_node = math.ceil(float(num_processes)/num_nodes)
```
timmoon10, line 251: What if you specify the number of nodes and run on CPU?
```suggestion
            if disable_cuda:
                # When CUDA is disabled, allow the number of resources per node to be overridden
                resources_per_node = math.ceil(float(num_processes)/num_nodes)
```

timmoon10 (changes requested):

PR #1672, opened 2020-10-19
Fix build errors when Distconv is enabled

oyamay:
This PR fixes some build errors when Distconv is enabled, mainly by replacing the MPI CUDA backend with the host-transfer backend.


bvanessen (approved):
LGTM

PR #1673, opened 2020-10-19
Adding the CosmoFlow and the 3D U-Net models

oyamay:
This PR adds two CNN models, the CosmoFlow network and the 3D U-Net.
These models are partitioned in the depth dimension among multiple GPUs (see https://arxiv.org/abs/2007.12856 for details).

They requires that Distconv is installed.


ndryden (approved):
Looks fine to me.

bvanessen (approved):
LGTM

PR #1674, opened 2020-10-19
Abstract CUDA backend implementation

mrwyattii:
Refactors the CUDA backend function definitions.  CUDA-specific code is kept in `cuda.hpp` headers and the non-specific (i.e., works with ROCm and CUDA) code is moved to a `gpu_lib.hpp` headers.  I also modified 2 files in `src` where we had previously missed a `cuda` --> `gpu_lib` replacement.

This is part 2 of 3 in porting the gpu_lib backend to HIP/ROCm.

Note that this PR includes changes from #1670 and will need to be rebased after that PR is merged.


include/lbann/utils/gpu/helpers.hpp
```diff
@@ -38,16 +38,22 @@
...
 namespace lbann {
 namespace gpu_lib = ::lbann::hip;
 }// namespace lbann
-
 #endif // LBANN_HAS_CUDA
+
+// Include gpu arch-independent stuff
+#if defined LBANN_HAS_GPU
+#include "gpu_lib.hpp"
+#endif // LBANN_HAS_GPU
```
timmoon10, line 57: Should we put the declarations before the implementations?

timmoon10 (commented):
Once this is rebased on #1670, this looks good. I am curious that we move a lot of the GPU math function implementations from the `impl/cuda.hpp` header to `impl/gpu_lib.hpp`. We should be aware that we are assuming that all vendors will use the same function names and will support all of the functions. It's not a bad assumption since these are the standard names used in the STL and SciPy, but I am slightly paranoid that Nvidia/AMD/Intel could do something weird.

benson31 (approved):
LGTM

PR #1661, opened 2020-10-07
Update the CMake for HIP/ROCm integration

benson31:
This has no impact on CUDA builds. Bamboo is as happy as it ever is. Let's merge quickly.

PR #1675, opened 2020-10-19
Port CUDA backend for ROCm support

mrwyattii:
Ports the CUDA-specific code identified in #1674 to ROCm in `rocm.hpp`.  Minus some implementation details, `rocm.hpp` is a copy/paste of `cuda.hpp` with CUDA types replaced with ROCm/HIP types.  This PR also does not attempt to support any versions of ROCm/HIP prior to the current version (3.8.0).

This is part 3 of 3 in porting the gpu_lib backend to HIP/ROCm.

Note that this PR includes changes from #1670 and #1674; it will need to be rebased after those PR are merged.  This PR also relies on the CMake changes in #1661.


include/lbann/utils/gpu/rocm.hpp
```diff
@@ -0,0 +1,212 @@
...
+/** Thrust execution policy. */
+using execute_on_stream
+#if THRUST_MAJOR_VERSION > 1 || THRUST_MINOR_VERSION >= 9
+  = ::thrust::hip_rocprim::execute_on_stream; // >= 1.9.1
+#elif THRUST_MAJOR_VERSION == 1 && THRUST_MINOR_VERSION == 8
+  = ::thrust::system::hip::detail::execute_on_stream;
+#else
+  = std::nullptr_t;
+  static_assert(false, "Thrust 1.8 or newer is required");
+#endif
```
timmoon10, line 163: Can we assume Thrust >=1.9 for ROCm >=3.8?
```suggestion
using execute_on_stream = ::thrust::hip_rocprim::execute_on_stream;
```

include/lbann/utils/impl/rocm.hpp
```diff
@@ -0,0 +1,254 @@
...
+  return *old_as_half;
+}
+__device__ __forceinline__
+float gpu_lib::atomic_add(float* address, float val) {
+  return atomicAdd(address, val);
+}
+__device__ __forceinline__
+double gpu_lib::atomic_add(double* address, double val) {
+  return atomicAdd(address, val);
+}
```
timmoon10, line 69: As far as I can tell, ROCm doesn't support `atomicAdd` for `double`. If we want full compatibility, we can copy the implementation from https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions.

timmoon10 (commented):
LGTM once #1674 is merged and any changes are propagated through.

include/lbann/utils/gpu/rocm.hpp
mrwyattii: I think we can, good catch!

src/utils/rocm.cpp
```diff
@@ -0,0 +1,226 @@
+#include "hip/hip_runtime.h"
```
timmoon10, line 1: OCD: Let's move this after the boilerplate.

timmoon10 (approved):
LGTM

src/utils/rocm.cpp
mrwyattii: hmm that's odd - I think the hipify script might have added this here 👎

timmoon10: lol

benson31 (approved):
LGTM

PR #1679, opened 2020-10-23
Refactor namespace cudnn to dnnlib

mrwyattii:
This establishes the `dnn_lib` namespace and lays the groundwork for supporting multiple DNN libraries in that namespace:

- `cudnn::` --> `dnn_lib::`
- replace `cudnn` in variable and function names with `dnn`
- aliasing for DNN library types
- establish the `dnn_lib/helpers.hpp` header that will load the appropriate DNN library code for LBANN layers
- replace `LBANN_HAS_CUDNN` with `LBANN_HAS_DNN_LIB` where appropriate
  - Some things will remain cuDNN-exclusive, primarily the RNN stuff in the GRU layer, as there is not equivalent for MIOpen
- cuDNN-specific code (i.e., `RNNDataDescriptor`) is moved to `cudnn.hpp` header while the rest lives in `dnn_lib.hpp`
- Moves the layer class enums (e.g., conv algos) to their own file (`dnn_enums.hpp`)
- Moves the `to_{dnn_lib}` and `from_{dnn_lib}` functions to `cudnn.hpp`

This PR contains changes from #1674 and #1675 and will need to be rebased when those are merged.

Note: this does not address some of the larger changes necessary in the convolution and pooling layers.  Those changes are captured in #1686 and #1687, respectively.

### DIFF
--- a/include/lbann/layers/activations/log_softmax.hpp
+++ b/include/lbann/layers/activations/log_softmax.hpp
@@ -29,7 +29,7 @@
 
 #include "lbann/layers/data_type_layer.hpp"
 #if defined LBANN_HAS_CUDNN
-#include "lbann/utils/cudnn.hpp"
+#include "lbann/utils/dnn_lib/helpers.hpp"
 #endif
 
 namespace lbann {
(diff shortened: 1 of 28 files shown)


timmoon10 (approved):
#1686 looks good to go, so I presume this PR should also be reasonable.

mrwyattii:
Changes are captured in #1686 - closing

PR #1682, opened 2020-10-27
Improve the serialization of Optimizers with Cereal

benson31:
This PR will result in fully serializable optimizers, complete with matrix data as needed. The initial step of establishing the optimizer serialization tests is now complete. These only compare metadata of the original and restored optimizers, but will expand to compare matrix data as well, once that capability is fully integrated. (Comparing just the metadata already identified two omissions from the serialization, which this PR also fixes.) The serialization of matrices is in-progress now, and I'll update this description when it's complete. Any initial feedback before I go too much further would be appreciated.


include/lbann/utils/serialization/serialize_matrices.hpp
```diff
@@ -0,0 +1,307 @@
...
+  if (mat.DistRank() == mat.Root())
+    ar(height, width);
+  El::mpi::Broadcast(height, mat.Root(), mat.DistComm(),
+                     El::SyncInfo<El::Device::CPU>{});
+  El::mpi::Broadcast(width, mat.Root(), mat.DistComm(),
+                     El::SyncInfo<El::Device::CPU>{});
+  mat.Resize(height, width);
+}
+
+#ifndef DOXYGEN_SHOULD_SKIP_THIS
```
bvanessen, line 178: I don't understand why this guard is here.

bvanessen (approved):
Overall LGTM

include/lbann/utils/serialization/serialize_matrices.hpp
benson31: It's there so that Doxygen doesn't add documentation for these. But I should remove this since I changed the static dispatch from the original version in which I added that.

include/lbann/optimizers/sgd.hpp
```diff
@@ -73,7 +73,9 @@ class sgd : public Cloneable<sgd<TensorDataType>,
   /** Archive for checkpoint and restart */
   template <class Archive> void serialize(Archive & ar) {
     ar(cereal::base_class<data_type_optimizer<TensorDataType>>(this),
-       CEREAL_NVP(m_momentum));
+       CEREAL_NVP(m_momentum),
+       CEREAL_NVP(m_nesterov));
+    // CEREAL_NVP(m_velocity));
```
timmoon10, line 78: The optimizer state should be serialized, or else we'll se wonky learning after deserializing.

timmoon10 (commented):
I've gone through everything except the unit tests. My question is whether Cereal automatically serializes/deserializes all the optimizer member variables, or if that's still in progress. The optimizer state (velocity in SGD or moment1/moment2 in Adam) should be serialized or else it will disrupt learning after deserializing.

include/lbann/utils/serialization/serialize_matrices.hpp
```diff
@@ -0,0 +1,402 @@
...
+          lbann::utils::WhenNotTextArchive<ArchiveT> = 1>
+void load(ArchiveT& archive,
+          El::Matrix<T, El::Device::CPU>& mat)
+{
+  LBANN_ASSERT(!mat.Viewing());
+  El::Int height, width;
+  archive(height, width);
+  mat.Resize(height, width);
+  archive(binary_data(mat.Buffer(),
+                      mat.Height()*mat.Width()*sizeof(T)));
```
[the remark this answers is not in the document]
benson31: Yes, see the test. I suppose I should be a bit more careful, since there's an implicit assumption that on load, `mat.Height() == mat.LDim()`. The `Resize` should take care of this in current Hydrogen, but I don't actually like that either.

include/lbann/optimizers/sgd.hpp
benson31: Do you mean that `m_velocity` should be serialized? Or are you referring to something else? If you're referring to `m_velocity`, this is the issue we discussed yesterday. Distributed matrices cannot yet be serialized inside LBANN classes. The reason for this is that the LBANN load functions will read an XML archive on rank 0, pipe it to a binary archive, broadcast the archive to the rest of the trainer, and then restore from binary. This will change as we continue the refactor, so I have just left these here as comments so it's easy to restore them in the next PR that will address a wider view of the "checkpoint workflow", if you will.

src/utils/unit_test/serialize_distmatrix_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+  El::DistMatrix<T, El::STAR, El::MC  , El::ELEMENT, D>,
+  El::DistMatrix<T, El::STAR, El::MD  , El::ELEMENT, D>,
+  El::DistMatrix<T, El::STAR, El::MR  , El::ELEMENT, D>,
+  El::DistMatrix<T, El::STAR, El::STAR, El::ELEMENT, D>,
+  El::DistMatrix<T, El::STAR, El::VC  , El::ELEMENT, D>,
+  El::DistMatrix<T, El::STAR, El::VR  , El::ELEMENT, D>,
+  El::DistMatrix<T, El::VC  , El::STAR, El::ELEMENT, D>,
+  El::DistMatrix<T, El::VR  , El::STAR, El::ELEMENT, D>>;
+
+// Now get all the devices. For now, just do CPU testing.
```
timmoon10, line 30:
```suggestion
```

timmoon10 (approved):

PR #1684, opened 2020-10-27
Abstract cuDNN backend implementation

mrwyattii:
Builds on #1679 and establishes `dnn_lib.hpp` header files.  Code for MIOpen + cuDNN is in `dnn_lib.hpp` while cuDNN-specific code remains in `cudnn.hpp`.  This also moves the enum class definitions to their own file in `ml_enums.hpp`.


mrwyattii:
The work here is captured in #1679 and #1686 - closing

PR #1685, opened 2020-10-28
Utility function to detect system in Bamboo tests

timmoon10:
This PR uses the approach from PR #1635 to allow us to skip Bamboo tests on unsupported systems. For example, if we wanted to only run the LeNet test on Pascal, we would add the following to the `setup_experiment` function in `test_integration_lenet.py`:
```python
    if tools.system(lbann) != 'pascal':
	message = f'{os.path.basename(__file__)} is only supported on pascal system'
	print('Skip - ' + message)
	pytest.skip(message)
```
Pinging @samadejacobs 

I've also taken the liberty of enabling GPU-only tests for the uniform hash and GRU layers. [Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM307-1).

### DIFF
--- a/bamboo/common_python/tools.py
+++ b/bamboo/common_python/tools.py
@@ -3,6 +3,7 @@ import copy
 import math
 import os
 import re
+import socket
 import sys
 import numpy as np
 import pytest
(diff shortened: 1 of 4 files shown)


timmoon10:
[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM307-2).

timmoon10:
[Bamboo is green on Lassen](https://lc.llnl.gov/bamboo/browse/LBANN-TIM307-6), and the remaining failures match the nightly build.

PR #1686, opened 2020-10-28
Refactor convolution algorithm selection and math type

mrwyattii:
This PR addresses the issue of `from_cudnn` wrapper functions around convolution algorithm selection.  The `from_cudnn` is moved inside the algorithm selection functions defined in `cudnn.cpp`.  In doing so, we allow for both MIOpen and cuDNN algorithm selection functions to return generic enum classes instead of library-specific data types.  At the same time, we conserve the current method for memoization of optimal algorithm selection.

Also, the functions for selecting the tensor math type are moved to the `dnn_lib` namespace.

This PR includes work from #1679 and should be rebased.


timmoon10 (approved):
LGTM

benson31 (approved):
LGTM

PR #1688, opened 2020-10-30
Data reader hdf5

davidHysom:
This PR implements a generalized HDF5/conduit data reader that utilizes both user supplied sample lists and data schemas. Two classes are introduced: data_reader_sample_list, which inherits from generic_data_reader, and data_reader_HDF5, which inherits from data_reader_sample_list. The intent is that additional readers that use sample lists, but whose data is not in HDF5, will also inherit from data_reader_sample_list.


bvanessen:
@davidHysom I haven't read this too closely, but on a quick overview this looks like a good direction.

davidHysom:
> @davidHysom I haven't read this too closely, but on a quick overview this looks like a good direction.

this is what I think should replace fetch_datum|label|response:
template<typename T> T* fetch(int index, const std::string &field_name);

Let me know if you have something else in mind.

bvanessen:
@davidHysom I have also rebased this PR.

include/lbann/data_readers/data_reader_HDF5_impl.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+namespace lbann {
+
+template<typename T_from, typename T_to>
+void hdf5_data_reader::coerceme(const T_from* data_in, size_t n_bytes, std::vector<T_to> & data_out) {
+  size_t n_elts = n_bytes / sizeof(T_from);
+  data_out.resize(0);
+  data_out.reserve(n_elts);
+  for (size_t j=0; j<n_elts; j++) {
+    data_out.push_back(*data_in++);
```
bvanessen, line 41: Do we need an explicit cast between types here?

davidHysom: the value that is copied (data_in) should always be a primitive C data type, so any compiler should correctly cast. But @benson31  would better answer.

benson31: A compiler _might_ complain about a narrowing cast (`int->size_t`, for example), but it's not a big deal.

bvanessen (approved):

PR #1691, opened 2020-11-05
Replace raw layer pointers with smart pointers

timmoon10:
We would like to use Cereal to serialize a model's layer graph, but it doesn't support serializing raw pointers. However, it does support serializing `std::weak_ptr`. This PR goes through and replaces all raw layer pointer state with `std::weak_ptr<Layer>`. The only significant substantive functional change is in the unpooling layer (use the hint layer instead of maintaining a pointer to a pooling layer). [Bamboo has no new failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM308-7).

Note that `std::weak_ptr` points to data in a `std::shared_ptr`. This is suboptimal since we would like for layers to be explicitly owned by models, and hence we would like to keep the layers in `std::unique_ptr<Layer>` s. There is an experimental [`std::observer_ptr`](https://en.cppreference.com/w/cpp/experimental/observer_ptr), but it is not available as of C++20 and [Bjarne doesn't like it](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1408r0.pdf). To make usage clear, I've typedef'ed `OwningLayerPtr=std::shared_ptr<Layer>` and `ViewingLayerPtr=std::weak_ptr<Layer>`.

### DIFF
--- a/src/layers/layer.cpp
+++ b/src/layers/layer.cpp
@@ -272,7 +272,7 @@ int Layer::get_input_size(size_t input_index) const {
 }
 
 std::vector<int> Layer::get_output_dims(size_t output_index) const {
-  const auto num_outputs = get_num_children();
+  const size_t num_outputs = get_num_children();
   if (m_output_dims_list.size() != num_outputs) {
     std::stringstream err;
     err << "attempted to access dimensions of output tensor "
@@ -301,11 +301,11 @@ int Layer::get_output_size(size_t output_index) const {
 }
 
 void Layer::set_output_dims(std::vector<int> dims, size_t output_index) {
-  if (m_output_dims_list.size() != get_num_children()
+  if (static_cast<int>(m_output_dims_list.size()) != get_num_children()
       || m_output_dims_list.size() <= output_index) {
     // Handles case where dims are set before child layers are set
-    m_output_dims_list.resize(std::max(get_num_children(),
-                                       output_index + 1));
+    m_output_dims_list.resize(El::Max(get_num_children(),
+                                      output_index + 1));
   }
   m_output_dims_list[output_index] = dims;
 }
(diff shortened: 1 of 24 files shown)


src/layers/layer.cpp
```diff
@@ -582,50 +528,189 @@ void Layer::write_proto(lbann_data::Layer* proto) const {
...
-void Layer::add_parent_layer(const Layer* parent) {
-  const auto parent_pos = std::find(m_parent_layers.begin(),
-                                    m_parent_layers.end(),
-                                    parent);
-  if (parent != nullptr
-      && parent != this
-      && parent_pos == m_parent_layers.end()) {
-    m_parent_layers.push_back(parent);
+ViewingLayerPtr Layer::get_parent_layer_pointer(size_t index) const {
+  if (index >= m_child_layers.size()) {
```
bvanessen, line 682: @timmoon10 this should be m_parent_layers.size()

timmoon10: Good catch.

bvanessen (approved):
Overall LGTM.  I was able to build locally.  Haven't tested, but since bamboo runs, we should be good.

src/proto/factories/layer_graph_factory.cpp
```diff
@@ -194,17 +150,18 @@ std::vector<std::unique_ptr<Layer>> construct_layer_graph(
...
     // Construct layer
-    std::unique_ptr<Layer> l;
+    OwningLayerPtr l;
 #define TEMPLATE_INSTANTIATION(TensorDataType, T_layout, T_device)      \
     do {                                                                \
       if (proto_datatype == TypeToProtoDataType<TensorDataType>::value  \
           && layout == T_layout                                         \
           && device == T_device) {                                      \
-        l = construct_layer<TensorDataType, T_layout, T_device>(        \
+        auto l_ = construct_layer<TensorDataType, T_layout, T_device>(  \
```
benson31, line 159: `std::shared_ptr<T>` is both assignable and constructible from `std::unique_ptr<T>` rvalues.
```suggestion
        l = construct_layer<TensorDataType, T_layout, T_device>(  \
```

src/proto/factories/layer_graph_factory.cpp
```diff
@@ -194,17 +150,18 @@ std::vector<std::unique_ptr<Layer>> construct_layer_graph(
...
           comm,                                                         \
           training_dr_linearized_data_size,                             \
           num_parallel_readers,                                         \
           proto_layer);                                                 \
+        l.reset(l_.release());                                          \
```
benson31, line 164:
```suggestion
```

benson31 (approved):
Overall this seems fine. Certainly sufficient to accomplish the goals.

timmoon10:
I'll address the comments in the next PR.

PR #1693, opened 2020-11-07
Minor tweaks for half data type constants

bvanessen:
Changed the instantiation of constants from using TensorDataType to use the El::To<TensorDataType> pattern


timmoon10 (approved):
:shipit:

PR #1694, opened 2020-11-11
Replace raw weights pointers with smart pointers

timmoon10:
This is a followup to PR #1691 that replaces `weights*` with `std::weak_ptr<weights>`. [No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM311-2).

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -114,15 +114,15 @@ model::model(const model& other) :
   }
 
   // Copy weights
-  std::unordered_map<weights*,weights*> weights_map;
+  std::unordered_map<weights*,ViewingWeightsPtr> weights_map;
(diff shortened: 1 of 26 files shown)


src/models/model.cpp
```diff
@@ -415,29 +423,6 @@ void model::add_metric(metric *m) {
   m_metrics.push_back(m);
 }
 
-void model::replace_weights(std::vector<weights*>& new_weights) {
```
bvanessen, line 418: I am trying to remember if there is any reason that this would still be necessary.  I believe that it was used in the previous multi-model approach, where weights were moved.

bvanessen (approved):
Overall LGTM.

benson31 (approved):
LGTM

PR #1698, opened 2020-11-13
Remove a shadowing declaration

benson31:
These lines shouldn't be necessary.


mrwyattii (approved):
Agreed, these were added mistakenly.  LGTM

PR #1699, opened 2020-11-14
Improved spack external definitions

bvanessen:
Added feature to build required spack packages for building LBANN
documentation.  Specifically, breathe, sphinx, and sphinx-rtd-theme.
Also added a feature to use a generic spack externals file that is
selected via spack platform and then specialized by the full spack
architecture.  If there is a more specific external file, that will be
preferred over a generic external.


mrwyattii (approved):
LGTM.

I tested this on 2 systems and in the end, the spack install works, but the first time running `install_lbann.sh -d` results in a spack error, where spack claims a dependency has been uninstalled:
```
Error: Cannot proceed with $PACKAGE: 1 uninstalled dependency: $DEPENDENCY
```
Running the script again, no error occurs and the installation is successful

benson31 (approved):
I didn't spend the necessary several hours to get this to build, but the changes to the scripts and inputs seem reasonable. +1 on the `SPACK_ARCH` updates.

PR #1700, opened 2020-11-16
Adding support for 3D Matrix Multiplication

szaman19:
- Test file located in bamboo/test_unit_batched_matmul.py
- Updated implementation in matmul.cpp

### DIFF
--- a/src/layers/math/matmul.cpp
+++ b/src/layers/math/matmul.cpp
@@ -197,12 +197,19 @@ void fp_compute_impl(matmul_layer<TensorDataType, data_layout::DATA_PARALLEL,El:
 
   const El::Int input0_height = *(input0_dims.rbegin()+1);
   const El::Int input0_width = *(input0_dims.rbegin());
+  const El::Int input1_height = *(input1_dims.rbegin()+1);
   const El::Int input1_width = *(input1_dims.rbegin());
   const El::Int output_height = *(output_dims.rbegin()+1);
   const El::Int output_width = *(output_dims.rbegin());
 
   const auto num_matrices = mat_depth * local_mini_batch_size;
-  
+  const auto input0_stride = input0_height * input0_width;
+  const auto input1_stride = input1_height * input1_width; 
+  const auto output_stride = output_height * output_width;
+
+  // Compute gradients for each mini-batch sample
+  // Note: cuBLAS expects matrices in Fortran layout while LBANN
+  // tensors are in C layout.
   {
     using namespace hydrogen;
     auto multisync = MakeMultiSync(gpu::get_sync_info(local_output),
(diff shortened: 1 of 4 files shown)


src/layers/math/matmul.cpp
```diff
@@ -164,32 +192,35 @@ void fp_compute_impl(matmul_layer<TensorDataType, data_layout::DATA_PARALLEL,El:
...
       transpose_input0 ? input0_height : input0_width,
       El::TypeTraits<TensorDataType>::One(),
-      local_input1.LockedBuffer(), input1_width, local_input1.LDim(),
-      local_input0.LockedBuffer(), input0_width, local_input0.LDim(),
+      local_input1.LockedBuffer(), input1_width, local_input1.LDim() / mat_depth,
+      local_input0.LockedBuffer(), input0_width, local_input0.LDim() / mat_depth,
       El::TypeTraits<TensorDataType>::Zero(),
-      local_output.Buffer(), output_width, local_output.LDim(),
-      local_mini_batch_size,
+      local_output.Buffer(), output_width, local_output.LDim() / mat_depth,
```
timmoon10, line 222: This won't be correct if the matrices aren't contiguous. I'd recommend replacing the matrix strides from `LDim()/mat_depth` to `width*height`. In case we get non-contiguous data, we can have something like:
```c++
if (!local_input1.Contiguous() || !local_input2.Contiguous() || !local_output.Contiguous()) {
  LBANN_ERROR(
    this->get_type()," layer \"",this->get_name(),"\" ",
    "has non-contiguous data buffers");
}
```
We can implement this case properly if we run into it.

timmoon10 (changes requested):
Mostly looks good, although there is an issue with the matrix strides when we call `gpu_blas::GemmStridedBatched`.

src/layers/math/matmul.cpp
szaman19: I'll change it to `width*height`, that should've been written that way. Actually, while we have contiguous buffers, that would make it more general and have MatMul defined for two matrices with shapes ( *, m, k) and  ( *, k, n), and mat1.shape[:-2] == mat2.shape[:-2]. It'd only change one value in the GEMM kernel, I believe.  CPU implementation would need a few changes though.

timmoon10: I'm thinking of situations where the Hydrogen matrices have `LDim() != Height()`, so the columns are not contiguous in memory. I believe the CPU implementation handles this case correctly, it's just the GPU implementation. If we wanted to support it, I believe cuBLAS exposes an API for batched GEMM with non-uniform strides, but that would be more effort than it would be worth right now.

timmoon10 (changes requested):
We should also have a guard against non-contiguous data in backprop.

szaman19:
> We should also have a guard against non-contiguous data in backprop.

Added! Not sure if the best way to implement it though. Lot's of code copying. But putting it in the friend impl functions didn't look good

timmoon10 (approved):
LGTM. Whenever you're ready we can merge this in.

szaman19:
:shipit:

PR #1705, opened 2020-11-18
Better support for unknown systems in Python contrib launcher

timmoon10:
`lbann.contrib.launcher.run` fails when a user runs on an unknown system within a known computing center. This PR changes the behavior so that it falls back to `lbann.run` if it doesn't recognize the system.


timmoon10:
Closes #1516.

PR #1706, opened 2020-11-19
Remove pointer to LBANN comm in layer class

timmoon10:
This removes the LBANN communicator pointer that's maintained by the layer class. The layer now accesses the communicator via its model. The goal of this work is to make the base layer class default-constructible. However, this PR does not touch any of the layer constructors to avoid merge conflicts with other work.

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM318-1).

### DIFF
--- a/include/lbann/layers/layer.hpp
+++ b/include/lbann/layers/layer.hpp
@@ -207,6 +207,7 @@ class Layer {
 
 public:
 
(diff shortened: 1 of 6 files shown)


bvanessen (approved):
LGTM

benson31 (approved):
LGTM

PR #1707, opened 2020-11-19
Allow LBANN to work with fewer samples than ranks per trainer

bvanessen:
Disable check for samples that will throw an unnecessary error on
ranks that will not receive samples because the mini-batch size is too
small.

Replaces PR #1668 to avoid conflicts due to buffered data coordinator refactor.


bvanessen:
@szaman19 Here is the PR that you need.  I will test it and then merge it.  But if you want to get a head start.

PR #1713, opened 2020-11-23
Fix assorted things that caused warnings when building with Clang

benson31:
[Letting Bamboo take this for a spin](https://lc.llnl.gov/bamboo/browse/LBANN-BEN43-WPT-1), but my local GCC builds are still fine (better, in fact, since this incidentally fixes an NVCC warning, too).

### DIFF
--- a/include/lbann/execution_contexts/execution_context.hpp
+++ b/include/lbann/execution_contexts/execution_context.hpp
@@ -105,7 +105,7 @@ public:
 
   /** Grab the trainer from the execution context */
   const trainer& get_trainer() const {
-    return m_trainer;
+    return *m_trainer;
   }
 
   trainer& get_trainer() {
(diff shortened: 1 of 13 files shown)


include/lbann/execution_contexts/execution_context.hpp
```diff
@@ -146,9 +146,9 @@ class execution_context {
 
 private:
   /** Pointer to the training context (execution environment) for the training algorithm */
-  trainer& m_trainer;
+  trainer* m_trainer;
 
-  training_algorithm& m_training_algorithm;
+  training_algorithm* m_training_algorithm;
```
timmoon10, line 151: I presume Cereal struggles with serializing reference members? In the future we may be want to replace the raw pointers with `std::shared_ptr`, e.g. if we use Cereal to checkpoint the execution context. That might be overkill though, since the execution context belongs to the trainer, which can set these fields in its setup phase.

timmoon10 (approved):

include/lbann/execution_contexts/execution_context.hpp
benson31: No, reference members should not be used in general. Pointers are superior in every way.

bvanessen (approved):

PR #1714, opened 2020-11-24
Make sure CPU softmax layer has access to softmax mode enum

timmoon10:
PR #1686 moved the `softmax_mode` mode enum out of `include/lbann/layers/activations/softmax.hpp` into `include/lbann/utils/dnn_enums.hpp`. We got compile errors on CPU systems because the softmax layer didn't have access to the enum, so the easy fix is to just include the header. I think this should be straightforward enough to merge once [Bamboo finishes](https://lc.llnl.gov/bamboo/browse/LBANN-TIM322-1), but I'm pinging @mrwyattii in case he has any thoughts after he gets back.


benson31 (approved):
:shipit:

PR #1715, opened 2020-12-01
Per-trainer prototext generation

samadejacobs:
Minor extension to LBANN to generate per trainer prototext.

### DIFF
--- a/python/lbann/launcher/__init__.py
+++ b/python/lbann/launcher/__init__.py
@@ -13,6 +13,7 @@ from lbann.util import make_iterable
 
 def run(trainer, model, data_reader, optimizer,
         work_dir=None,
+        proto_file_name='experiment.prototext',
         nodes=1,
         procs_per_node=1,
         time_limit=None,
@@ -81,6 +82,7 @@ def run(trainer, model, data_reader, optimizer,
     if not work_dir:
         work_dir = experiment_dir
     script = make_batch_script(work_dir=work_dir,
+                               proto_file_name=proto_file_name,
                                nodes=nodes,
                                procs_per_node=procs_per_node,
                                time_limit=time_limit,
@@ -98,7 +100,7 @@ def run(trainer, model, data_reader, optimizer,
     # Batch script invokes LBANN
     lbann_command = [lbann_exe]
     lbann_command.extend(make_iterable(lbann_args))
-    prototext_file = os.path.join(script.work_dir, 'experiment.prototext')
+    prototext_file = os.path.join(script.work_dir, proto_file_name)
     lbann.proto.save_prototext(prototext_file,
                                trainer=trainer,
                                model=model,
@@ -124,6 +126,7 @@ def run(trainer, model, data_reader, optimizer,
 
 def make_batch_script(script_file=None,
                       work_dir=None,
+                      proto_file_name='experiment.prototext',
                       nodes=1,
                       procs_per_node=1,
                       time_limit=None,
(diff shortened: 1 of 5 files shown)


python/lbann/launcher/__init__.py
```diff
@@ -124,6 +126,7 @@ def run(trainer, model, data_reader, optimizer,
 
 def make_batch_script(script_file=None,
                       work_dir=None,
+                      proto_file_name='experiment.prototext',
```
timmoon10, line 129: This is unused.
```suggestion
```

bvanessen (approved):

python/lbann/launcher/__init__.py
```diff
@@ -81,6 +82,7 @@ def run(trainer, model, data_reader, optimizer,
     if not work_dir:
         work_dir = experiment_dir
     script = make_batch_script(work_dir=work_dir,
+                               proto_file_name=proto_file_name,
```
timmoon10, line 85: No need to pass this into `make_batch_script`.
```suggestion
```

src/utils/protobuf_utils.cpp
```diff
@@ -62,6 +63,13 @@ parse_prototext_filenames_from_command_line(
     if (equal_sign != std::string::npos) {
       std::string which = s.substr(2, equal_sign-2);
       std::string fn = s.substr(equal_sign+1);
+      //check if trainer is append to filename string
+      size_t t_pos = fn.find("trainer");
+      if(t_pos != std::string::npos) {
+         //append appropriate trainer id to prototext filename
+         std::string fname = fn.substr(0,t_pos+7)+ std::to_string(trainer_rank);
+         fn = fname;
+      }
```
timmoon10, line 72: I don't like how we hard-code this naming scheme where all prototext files have to end in ".trainer\<num\>". It would be better if we provided a list and trainer n could load the nth file in the list.

python/lbann/launcher/__init__.py
```diff
@@ -13,6 +13,7 @@
 
 def run(trainer, model, data_reader, optimizer,
         work_dir=None,
+        proto_file_name='experiment.prototext',
```
timmoon10, line 16: If I'm honest, I don't think modifying this function is the right way to go. It's supposed to be a convenience driver, and adding more knobs to handle multiple models will make it bloated, convoluted, and more likely to fail when a user puts in a weird combination of options.

I would prefer if we wrote a function built off of `make_batch_script` and put in an application dir. This function would take in multiple models and save each one of them with `save_proto`. It would then construct a batch script with the correct LBANN invocation to specify multiple models.

samadejacobs: I believe we discussed this at a previous LBANN meeting. This does no harm IMO and it is just one line change instead of having user write the function you suggested for every application. I agree with that the _run_ function has a long list of arguments already but I don't think this one argument will do any harm; it is optional but also provide users flexibility to rename the default prototext filename if so desire.

python/lbann/launcher/__init__.py
samadejacobs: Fixed, thanks.

python/lbann/launcher/__init__.py
timmoon10: From what I gather, we require many more changes for your use-case. You need to save multiple models, each with a different name. This function is still only capable of saving a single model, although now you can change the filename. If `run` won't work for your eventual use-case, we might as well transition to a new launcher sooner rather than later.

PR #1716, opened 2020-12-01
data_converter class

davidHysom:
Provides methods to construct an El::Matrix<> from an lbann::DataType*, and to run a transform pipleline. Matrix dimensions; the transforms to run; etc, are specified in a conduit::Schema

The immediate need for this class is as an intermediary between the new hdf5_data_reader -- which knows nothing about matrices -- and the transform pipeline -- which does.


davidHysom:
no longer part of development pipeline

PR #1719, opened 2020-12-03
Feature modular trainer split

bvanessen:
Separate the splitting of the MPI communicator from the creation of the trainer object.  Change the procs_per_trainer option to only be a command line or environment argument, not a trainer prototext field.


bvanessen:
@samadejacobs I have done a bit of testing with the `train_atom_wae.py` launch script.  Can you please take a look at the rest of the ATOM scripts to make sure that they launch correctly.

timmoon10 (approved):
To be clear, we are now making `procs_per_trainer` an expert feature. If we want to expose it in the Python front-end, we can put it in the run functions.

benson31 (approved):
Looks fine.

samadejacobs:
> @samadejacobs I have done a bit of testing with the `train_atom_wae.py` launch script. Can you please take a look at the rest of the ATOM scripts to make sure that they launch correctly.

@bvanessen this should be application independent and should work with any application. I can't test really until I am able to resolve my spack issue.

samadejacobs (approved):
LGTM, PFE/prototext files that use procs_per_trainer need to be modified, it seems you've covered that with applications we care about.

PR #1720, opened 2020-12-03
Bugfixes for distconv with CosmoFlow

bvanessen:
Updated the HDF5 data reader to use the proper ordered map
(m_supported_input_types) that tracks which types of information are
available from a data reader, specifically if responses and labels are
supported.

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -32,6 +32,12 @@ Build system:
 
 Bug fixes:
  - Fixed a bug where LBANN didn't set the Hydrogen RNG seed
+ - Fixed both CosmoFlow and UNet models PFE as well as addressed
+   issues in the data reader and data coordinator.
+ - Fixed the HDF5 data reader to properly specify the supported I/O
+   types
+ - Fixed calculation of the linearized response size
+ - Fixed the data coordinator's interface to input_layer
 
 Retired features:
 
--- a/applications/physics/cosmology/cosmoflow/cosmoflow.py
+++ b/applications/physics/cosmology/cosmoflow/cosmoflow.py
@@ -319,7 +319,7 @@ if __name__ == "__main__":
 
     parser.add_argument(
         '--dynamically-reclaim-error-signals', action='store_true',
-        help='Allow DiHydrogen to reclaim error signals buffers (default: False)')
+        help='Allow LBANN to reclaim error signals buffers (default: False)')
 
     parser.add_argument(
         '--batch-job', action='store_true',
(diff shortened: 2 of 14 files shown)


oyamay (approved):
Looks good to me.

bvanessen:
@oyamay With the latest fixes I got it to go through 1 epoch and then it crashed in backprop.  I don't know if it ran out of memory or something else happened.

```
model0 (instance 0) validation mini-batch time statistics : 0.0117616s mean, 0.0117616s max, 0.0117616s min, N/A stdev
[1] Epoch : stats formated [tr/v/te] iter/epoch = [1/1/1]
            global MB = [   1/   1/   1] global last MB = [   1  /   1  /   1  ]
             local MB = [   1/   1/   1]  local last MB = [   1+0/   1+0/   1+0]
Model 0 GPU memory usage statistics : 3.33 GiB mean, 3.17 GiB median, 3.5 GiB max, 3.16 GiB min (15.8 GiB total)
Caught signal 11 (SIGSEGV - invalid memory reference) on rank 3
Stack trace:
   0: lbann::stack_trace::get()
   2: /usr/WS1/vanessen/lbann_bvanessen.git/build/gnu.Release.lassen.llnl.gov/install/lib64/liblbann.so(+0x15b56fc) [0x2000016556fc] (could not find stack frame symbol)
   3: __kernel_sigtramp_rt64 (demangling failed)
   4: __kernel_sigtramp_rt64 (demangling failed)
   5: lbann::Layer::back_prop()
   6: lbann::model::backward_prop()
   7: lbann::sgd_training_algorithm::train_mini_batch(lbann::sgd_execution_context&, lbann::model&, lbann::data_coordinator&)
   8: lbann::sgd_training_algorithm::train(lbann::sgd_execution_context&, lbann::model&, lbann::data_coordinator&, unsigned long, unsigned long)
   9: lbann::trainer::train(lbann::model*, long long, long long)
  10: /usr/WS1/vanessen/lbann_bvanessen.git/build/gnu.Release.lassen.llnl.gov/install/bin/lbann() [0x1000e2c8] (could not find stack frame symbol)
  11: /lib64/libc.so.6(+0x25300) [0x200042ca5300] (could not find stack frame symbol)
  12: __libc_start_main (demangling failed)
```

bvanessen:
@timmoon10 @oyamay This does not fix all of the problems, but does fix some.  We may want to merge this in if it also helps address some issues that @timmoon10 is seeing on other LTFB problems.

oyamay:
@bvanessen I confirmed that the latest version works for the CosmoFlow network for both 128^3 (with 4-way) and 512^3 (8-way) cubes. Note that the number of 512^3 samples should be sufficiently small to run on the minimum number of nodes (2 nodes).

bvanessen:
@benson31 @oyamay @timmoon10 This PR is now working on Lassen.  I have updated the PFE for this and UNet to ensure that error signal buffers are always kept, since it is required for DistConv support.  There is now a flag to enable buffer reclamation.

timmoon10 (approved):

PR #1722, opened 2020-12-05
Bugfix and unit tests for LTFB

timmoon10:
It appears that PR #1538 has broken LTFB. The root cause is because the data coordinator resized all of its data matrices at every step. However, LTFB involves switching between training and validation execution modes. In the process of configuring the data matrices to load validation data, it would clobber the data matrices for training data. I've modified the arguments to `buffered_data_coordinator::fp_setup_data` so that it only interacts with the data matrices for the current execution mode.

To help catch these kinds of errors in the future, I've also added two Bamboo unit tests:
- `test_unit_callback_ltfb.py`: Check that LTFB weight exchange is working as expected.
- `test_unit_callback_ltfb_data.py`: Check that switching between training and validation execution modes in LTFB works correctly.

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM325-1).

### DIFF
--- a/bamboo/unit_tests/test_unit_callback_ltfb.py
+++ b/bamboo/unit_tests/test_unit_callback_ltfb.py
@@ -1,0 +1,263 @@
+"""Test to check weight exchanges in LTFB.
+
+Each model has a randomly initialized weights object. An LTFB round is
+performed after every training step and winners are chosen randomly.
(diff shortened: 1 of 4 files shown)


bvanessen (approved):
LGTM and bamboo has passed.

PR #1723, opened 2020-12-08
Refactor dnn_lib convolution parameter type from distmatrix to matrix

mrwyattii:
In the `dnn_lib` functions for the convolution layer, several matrices had type `El::DistMatrix`.  This PR refactors those parameters to be of type `El::Matrix` and avoid the requirement of distributed matrices for these function calls.


benson31 (approved):
LGTM

PR #1724, opened 2020-12-08
Feature add tournament execution mode

bvanessen:
Adds a new execution mode for LTFB tournaments.  Creates the ability to split a training data reader into validation and tournament data sets.  Removed some deprecated features, namely the ability to set a lead JAG data reader and partition the data reader index lists between data readers that used to be attached to individual models.

### DIFF
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -70,6 +70,7 @@ class trainer;
  */
 class generic_data_reader {
  public:
+  using unused_index_map_t = std::map<execution_mode,std::vector<int>>;
 
  #define JAG_NOOP_VOID if (m_jag_partitioned) { return; }
  #define JAG_NOOP_INT if (m_jag_partitioned) { return 0; }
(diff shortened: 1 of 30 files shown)


include/lbann/data_readers/data_reader.hpp
```diff
@@ -545,15 +542,27 @@ class generic_data_reader {
     return (int)m_shuffled_indices.size();
   }
   /// Get the number of unused samples in this dataset.
-  int get_num_unused_data() const {
-    return (int)m_unused_indices.size();
+  int get_num_unused_data(execution_mode m) const {
+    if(m_unused_indices.count(m)) {
+      return (int)m_unused_indices.size();
```
timmoon10, line 547:
```suggestion
      return (int)m_unused_indices.at(m).size();
```

timmoon10 (approved):
If there are no new Bamboo errors, we can merge this.

I don't like how we're baking in LTFB into our data ingestion model. It would be better if we got rid of the rigid data split and just allowed the user provide an arbitrary number of data readers, whatever is required by the application (e.g. normal training needs train and val sets, transfer learning requires two train sets, inference only needs an inference set). Deadlines are deadlines though.

bvanessen:
> I don't like how we're baking in LTFB into our data ingestion model. It would be better if we got rid of the rigid data split and just allowed the user provide an arbitrary number of data readers, whatever is required by the application (e.g. normal training needs train and val sets, transfer learning requires two train sets, inference only needs an inference set). Deadlines are deadlines though.

I agree with @timmoon10 that we want to make things more flexible.  As a future PR we can find a way to push towards a more flexible data sets.

bvanessen:
There are no new bamboo errors. Just an instability in the atom integration test, where it will occasionally get too good of a result.

PR #1728, opened 2020-12-10
Perform LTFB model exchanges with temporary models

timmoon10:
This is some refactoring I've done in preparation for LTFB with heterogeneous architectures. In particular, we have problems because the LTFB callback is owned by a model. What happens when LTFB loads a checkpoint and replaces all the callbacks? The "right answer"® is to make LTFB a training algorithm, but even so I think this PR makes LTFB slightly easier to reason about.

Instead of reloading checkpoints into the active model, I figure it will be easier to load into a temporary model. If the remote model is better than the local model, then we swap their data (everything except callbacks). This adds the overhead of creating the temporary model at every LTFB round and doubles the memory footprint, but also reduces the number of times we need to load checkpoints (previously, we needed to load a checkpoint to restore the local model if it is better than the remote model). I've also done some refactoring to concentrate the differences between the "sendrecv weights" and "checkpoint file" communication algorithms, so they only differ in a single function call.

LTFB with "checkpoint file" communication runs and appears _mostly_ correct (checkpointing doesn't handle multiple metrics correctly, but that will be moot when #1709 is merged and we get full model serialization with Cereal). Once we have checkpoints that store model topology, it will just be a matter of making sure that LTFB is calling the correct checkpointing API in the "checkpoint file" communication algorithm, and then heterogeneous LTFB should just work.

[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM325-2).

### DIFF
--- a/bamboo/unit_tests/test_unit_callback_ltfb.py
+++ b/bamboo/unit_tests/test_unit_callback_ltfb.py
@@ -89,7 +89,11 @@ def construct_model(lbann):
     ]
     callbacks = [
         lbann.CallbackPrint(),
-        lbann.CallbackLTFB(batch_interval=1, metric='random'),
+        lbann.CallbackLTFB(
+            batch_interval=1,
+            metric='random',
+            communication_algorithm='checkpoint_file',
+        ),
     ]
 
     # Construct model
--- a/bamboo/unit_tests/test_unit_callback_ltfb.py
+++ b/bamboo/unit_tests/test_unit_callback_ltfb.py
@@ -40,7 +40,7 @@ def initialize_rng():
 
 # Sample access functions
 _mini_batch_size = 2
-_num_epochs = 3
+_num_epochs = 5
 def get_sample(index):
     initialize_rng()
     return (random.gauss(0,1),)
--- a/include/lbann/callbacks/ltfb.hpp
+++ b/include/lbann/callbacks/ltfb.hpp
@@ -55,8 +55,6 @@ namespace callback {
  *      or a subset of the weights? Hyperparameters?
  *    - Can this be used to explore model architectures?
  *
- *  @todo Exchange optimizer state.
- *  @todo Support heterogeneous models.
  */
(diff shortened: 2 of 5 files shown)


samadejacobs:
LGTM. Few comments or clarification:
(1) For incorrect metrics, does this mean that merging this PR would have to wait for the appropriate fix in PR #1709 ?
(2) How do we reproduce current (existing) LTFB approach of just exchanging model weights? To ask differently, it seems by default, this refactor would swap layers, metrics, objective function, and weights, I am wondering what if user want to swap weights only, or for a practical scenario of swapping everything but the objective function(s)?
(3) I assume you made an assumption that a large portion of memory needed for storing (local copy) of a model goes to model weights, I think that is a fair assumption.

timmoon10:
1. I've rebased on PR #1709 and the `checkpoint_file` communication algorithm works as expected. LTFB should now work with heterogeneous models.
2. No substantive change has been made to the default communication algorithm (`sendrecv_weights`). It will only transfer weight data, but it requires that all models have the same architecture. If you want heterogeneous architectures or hyperparameters, then you need to checkpoint the entire model. Checkpointing subsections of the model (e.g. just the weights or subsections of the layer graph) is very difficult and messy. At that point, I think it's better to checkpoint the entire model and separately extract out the parts that you want.

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM326-1).

benson31:
Re (2): Serializing subgraphs of the model is not supported (though it won't error out, it just won't work right because of the ... suboptimal pointer relationships). However, serializing just the weights will work fine (since they have no model topology information in them). One could serialize, e.g., the set of weights whose names appear in some set to a single archive with no problem. One would have to take care, in this case, either to prepend a count of the weights that will appear in the archive or to deserialize in a way that won't overflow the end of the archive.

samadejacobs (approved):

PR #1730, opened 2020-12-14
Upgrade superbuild for CUDA 11

oyamay:
This PR applies the following changes to superbuild and the LC build script to adapt to recent changes to LBANN and its dependencies:
* The default CUDA, cuDNN, and NCCL versions are changed to 11.1, 8.0.4, and 2.7.8 respectively.
* CUB build is disabled by default because it is now a part of CUDA 11.
* Aluminum's CUDA build is enabled when Distconv is enabled.
* The CMake version is changed to 3.18.1 as 3.17 or more is required by LBANN.


timmoon10 (approved):
Does this work for you? If so, then we can ship it.

If you don't mind switching to Spack, I can also send you the build script I've been using.

oyamay:
@timmoon10 It works on Lassen except for a CMake configuration error of Conduit, but it will be patched soon (according to @benson31).

PR #1731, opened 2020-12-15
Support nvprof in the Python frontend

oyamay:
This PR adds nvprof options to the Python frontend.
When `nvprof` of `lbann.contrib.launcher.run` is set, nvprof generates per-process tracing files in the working directory.
`LBANN_WITH_NVPROF` CMake variable should be set to use this function.


python/lbann/util/__init__.py
```diff
@@ -28,3 +29,23 @@ def make_nd_array(*dims):
...
+
+    Args:
+        nvprof_exe (str, optional): nvprof executable.
+        work_dir (str, optional): Working directory.
+        output_name (str, optional): File name to output nvprof trace.
+
+    """
+
+    if output_name is None:
+        output_name = "prof_%q{OMPI_COMM_WORLD_RANK}.nvvp"
```
timmoon10, line 46: This is somewhat less clear, but supports non-OpenMPI systems:
```suggestion
        output_name = "prof-%h-%p.nvvp"
```

timmoon10 (approved):

PR #1733, opened 2020-12-16
Move model checkpointing from end of training step to beginning

timmoon10:
We have an off-by-one error for applications that require synchronized checkpointing and dumping weights/activations/metrics. Checkpointing happens at the end of the training step, after we apply the optimizer step, so the checkpointed model is one step later than the other dumped objects. We can fix this by checkpointing at the beginning of the training step.

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM327-1).

### DIFF
--- a/include/lbann/utils/visitor_hooks.hpp
+++ b/include/lbann/utils/visitor_hooks.hpp
@@ -1,0 +1,80 @@
+////////////////////////////////////////////////////////////////////////////////xecu
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+// may not use this file except in compliance with the License.  You may
+// obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the license.
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef LBANN_VISITOR_HOOKS_HPP_INCLUDED
+#define LBANN_VISITOR_HOOKS_HPP_INCLUDED
+
+// #include "lbann_config.hpp"
+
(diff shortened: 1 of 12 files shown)


bvanessen (changes requested):
This is not working quite the way that we want yet.  I am trying it on our checkpoint / restart overhaul to test.

bvanessen:
@timmoon10 I have rebased this PR.

bvanessen:
@timmoon10 @benson31 Please take a look at this updated version of the PR.  I added a `visitor_hook` enum to identify the phase and that could eventually be exposed to the PFE to allow the user to select which phases a callback executes on.

include/lbann/utils/visitor_hooks.hpp
```diff
@@ -0,0 +1,75 @@
+////////////////////////////////////////////////////////////////////////////////xecu
```
benson31, line 1:
```suggestion
////////////////////////////////////////////////////////////////////////////////
```

include/lbann/utils/visitor_hooks.hpp
```diff
@@ -0,0 +1,75 @@
...
+    train_end,
+    phase_end,
+    epoch_begin,
+    epoch_end,
+    batch_begin,
+    batch_end,
+    test_begin,
+    test_end,
+    validation_begin,
+    validation_end,
```
timmoon10, line 50: Visitor hooks should be independent of execution modes, so the validation and test epochs should be indistinguishable. If the callback wants to know the execution mode, it should explicitly ask the execution context.
```suggestion
    evaluate_begin,
    evaluate_end,
```

include/lbann/utils/visitor_hooks.hpp
```diff
@@ -0,0 +1,75 @@
...
+    forward_prop_begin,
+    forward_prop_end,
+    backward_prop_begin,
+    backward_prop_end,
+    optimize_begin,
+    optimize_end,
+    batch_evaluate_begin,
+    batch_evaluate_end,
+    evaluate_forward_prop_begin,
+    evaluate_forward_prop_end,
```
timmoon10, line 60: Training and evaluation forward prop are identical, so I think the callback should explicitly query the execution context if it wants to know the execution mode.
```suggestion
```

timmoon10 (commented):
I like the idea of the visitor hooks since it gives the callback more flexibility to choose when it is run. I'd like to replace all of the `on_batch_begin`/`on_epoch_begin`/`on_test_begin`/etc virtual functions with a `run` function that is called at every visitor hook point (we may need separate `run` functions to handle layer-wise callbacks, model callbacks, and trainer callbacks).

One caveat: we should be careful to keep visitor hooks distinct from execution modes. The visitor hook only tells where the callback is being invoked, and the callback should explicitly ask the execution context if it wants to know the execution mode. But I realize the existing callback infrastructure doesn't do a good job with that distinction either, so we shouldn't adopt my suggestions yet in the interest of keeping this PR as a straight refactor. The suggestions are more guides for future work.

include/lbann/utils/visitor_hooks.hpp
bvanessen: Okay @timmoon10, I have updated the visitor hook phases to remove execution mode specific ones and create new `execution_mode_[begin|end]` phases and so on that are combined with the execution mode when they are displayed.

bvanessen (approved):

PR #1738, opened 2020-12-28
Support DiHydrogen's mean squared error layers

oyamay:
This PR makes mean squared error layers available for Distconv-enabled networks (https://github.com/LLNL/DiHydrogen/pull/45). The architectural requirements of mean squared error layers are the same as those of cross-entropy loss layers.

This PR also allows users to specify the number of channels of the "label" tensors by setting `DISTCONV_LABEL_NUM_CHANNELS`. In previous implementations, this number of channels was calculated as one, assuming a segmentation task, but this feature is necessary when it is not true, such as in reconstruction tasks.


benson31 (approved):
LGTM

samadejacobs (approved):
LGTM

bvanessen (approved):

PR #1739, opened 2021-01-04
Add complex numbers support

graham63:


timmoon10 (approved):
LGTM.

Perhaps we should start thinking about making the type names more manageable. I'd support getting rid of the `Tensor` and making it just `InputDataType`/`OutputDataType`. We can defer this until later though.

include/lbann/layers/data_type_layer.hpp
```diff
@@ -64,44 +64,56 @@ using supported_layer_data_type = h2::meta::TL<
...
-  using AbsDistMatrixType = El::AbstractDistMatrix<TensorDataType>;
+  using AbsDistMatrixType = El::AbstractDistMatrix<InputTensorDataType>;
+  using InputAbsDistMatrixType = El::AbstractDistMatrix<InputTensorDataType>;
+  using OutputAbsDistMatrixType = El::AbstractDistMatrix<OutputTensorDataType>;
 
   /** @brief The proxy tensor type expected in this object. */
   template <El::Device D>
-  using AbsDistMatReadProxyType = El::AbstractDistMatrixReadDeviceProxy<TensorDataType, D>;
+  using AbsDistMatReadProxyType =
+    El::AbstractDistMatrixReadDeviceProxy<InputTensorDataType, D>;
```
bvanessen, line 82: Since AbsDistMatReadProxyType is just a synonym of InputAbsDistMatReadProxyType, I am thinking that we should get rid of it and only have Input and Output versions.

include/lbann/layers/data_type_layer.hpp
```diff
@@ -64,44 +64,56 @@ using supported_layer_data_type = h2::meta::TL<
...
+  template <El::Device D>
+  using InputAbsDistMatReadProxyType =
+    El::AbstractDistMatrixReadDeviceProxy<InputTensorDataType, D>;
+  template <El::Device D>
+  using OutputAbsDistMatReadProxyType =
+    El::AbstractDistMatrixReadDeviceProxy<OutputTensorDataType, D>;
 
   /** @brief The local tensor type expected in this object. */
-  using AbsMatrixType = El::AbstractMatrix<TensorDataType>;
+  using AbsMatrixType = El::AbstractMatrix<InputTensorDataType>;
```
bvanessen, line 91: Same as above, we should get rid of AbsMatrix Type.

src/layers/data_type_layer.cpp
```diff
@@ -282,45 +290,57 @@ auto data_type_layer<TensorDataType>::get_error_signals(int parent_index) const
...
+get_local_error_signals(int parent_index) const -> const InputAbsMatrixType& {
   return get_error_signals(parent_index).LockedMatrix();
 }
 
 // Accessing matrices corresponding to parent/child layer
-template <typename TensorDataType>
-auto data_type_layer<TensorDataType>::get_activations(const Layer& child) const -> const BaseDistMat& {
+template <typename InputTensorDataType, typename OutputTensorDataType>
+auto data_type_layer<InputTensorDataType, OutputTensorDataType>::
+get_activations(const Layer& child) const -> const BaseDistMat& {
```
bvanessen, line 343: Why is this returning a BaseDistMat not an Output matrix?

src/layers/data_type_layer.cpp
```diff
@@ -335,8 +355,9 @@ auto data_type_layer<TensorDataType>::get_activations(const Layer& child) const
   }
   return get_activations(child_index);
 }
-template <typename TensorDataType>
-auto data_type_layer<TensorDataType>::get_error_signals(const Layer& parent) const -> const BaseDistMat& {
+template <typename InputTensorDataType, typename OutputTensorDataType>
+auto data_type_layer<InputTensorDataType, OutputTensorDataType>::
+get_error_signals(const Layer& parent) const -> const BaseDistMat& {
```
bvanessen, line 360: Similarly, shouldn't this be an input mat?

bvanessen (changes requested):

src/layers/data_type_layer.cpp
bvanessen: Okay, I see that this is getting another layer's activations, so it may be a pain to figure out what it's output type, but shouldn't we know this statically?

bvanessen:
@benson31 @timmoon10 I have added support for distconv and addressed my own comments.

bvanessen (approved):

timmoon10 (approved):

benson31 (approved):
LGTM.

PR #1740, opened 2021-01-05
Build OpenCV with imgcodecs module

timmoon10:
I encounter build problems when I build OpenCV with Spack, especially from this line:
https://github.com/LLNL/lbann/blob/42a7c5c2f4d46505aaeb1b6148281b1620b31d25/src/callbacks/save_images.cpp#L35
It is probably related to https://github.com/spack/spack/pull/20378, which added the `+imgcodecs` variant to the OpenCV package. Manually enabling with variant fixed the problem for me.


bvanessen (approved):

bvanessen:
This should be added for the x86 environment as well. Or we can wait to merge my PR.

PR #1732, opened 2020-12-15
Feature reworked install process

bvanessen:
Updated the spack build scripts and instructions to use the `spack dev-build` and `spack build-env` commands.  Also makes it simpler to find dependencies using `spack externals find`.

PR #1742, opened 2021-01-06
Remove the superbuild ecosystem

benson31:
Spack is the preferred way to build, and #1732 should clean up some lingering rough edges. Developers should be using that ecosystem anyway. I do not want to continuing maintaining this deprecated build system. Thus I suggest simply deleting it.


timmoon10 (approved):
:shipit: 
It's the passing of an era.

bvanessen (approved):

PR #1744, opened 2021-01-08
Move state from data reader to data coordinator

bvanessen:
Mini-batch size is now passed into the data reader fetch functions
from the data coordinator, rather than having the data reader
calculate what is the current mini-batch size.


bvanessen:
@benson31 Bamboo has no new failures

PR #1745, opened 2021-01-09
More LTFB refactoring

benson31:
1. Adds a new "checkpoint_binary" approach to LTFB model exchange.
2. Refactor the LTFB implementation into a strategy-based approach. The strategies are private for now.

~~The tests still pass for `checkpoint_file`; however, I'm still debugging some setup issues for the `checkpoint_binary` approach.~~ The LTFB tests appear to pass. I have updated `test_unit_callback_ltfb.py` to use `checkpoint_binary` instead of `checkpoint_file`.

~~Bamboo is running now.~~ Bamboo failures match `develop`.

### DIFF
--- a/src/callbacks/ltfb.cpp
+++ b/src/callbacks/ltfb.cpp
@@ -220,6 +220,25 @@ El::Int get_partner_trainer(lbann_comm& comm,
   return send_buffer[comm.get_trainer_rank()];
 }
 
+void restore_model_weights(
+  model& m,
+  std::unordered_map<std::string, std::unique_ptr<weights>>& restore_weights)
+{
+  // Restore weights that shouldn't be exchanged
+  if (restore_weights.empty())
+    return;
+
+  // FIXME: Generalize this; enable ptr move??
+  for (auto w : m.get_weights()) {
+    if (restore_weights.count(w->get_name()) > 0) {
+      using TensorDataType = DataType;
+      using WeightsType = data_type_weights<TensorDataType>;
+      dynamic_cast<WeightsType&>(*w)
+        = dynamic_cast<WeightsType&>(*restore_weights[w->get_name()]);
+    }
+  }
+}
+
 /** @class SendRecvWeights
  *  @brief Exchange model weights directly using sendrecvs.
  *  @todo More general approach to exchange optimizer state. Currently
@@ -456,17 +475,7 @@ public:
       p.close_restart();
     }
 
-    // Restore weights that shouldn't be exchanged
-    if (!restore_weights.empty()) {
-      for (auto w : m.get_weights()) {
-        if (restore_weights.count(w->get_name()) > 0) {
-          using TensorDataType = DataType;
-          using WeightsType = data_type_weights<TensorDataType>;
-          dynamic_cast<WeightsType&>(*w)
-            = dynamic_cast<WeightsType&>(*restore_weights[w->get_name()]);
-        }
-      }
-    }
+    restore_model_weights(m, restore_weights);
   }
 private:
   std::string ckpt_basedir_;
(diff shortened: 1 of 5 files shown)


src/callbacks/ltfb.cpp
```diff
@@ -34,283 +33,551 @@
...
+    std::string save_model_ckpt = oss.str(), load_model_ckpt;
+    if (comm.am_trainer_master())
+    {
+      std::size_t save_size=save_model_ckpt.size(), load_size=0;
+      comm.sendrecv(&save_size, 1, partner_trainer, 0,
+                    &load_size, 1, partner_trainer, 0,
+                    El::SyncInfo<El::Device::CPU>{});
+      load_model_ckpt.resize(load_size);
+
+      while (save_size && load_size)
```
timmoon10, line 536: We should handle the case where the send and load models require a different number of data blocks:
```suggestion
      while (save_size || load_size)
```

src/callbacks/ltfb.cpp
```diff
@@ -34,283 +33,551 @@
...
+        int this_blk_send_size =
+          (save_size > max_blk_size_size_t ? max_blk_size : save_size);
+        int this_blk_recv_size =
+          (load_size > max_blk_size_size_t ? max_blk_size : load_size);
+        comm.sendrecv(
+          reinterpret_cast<El::byte const*>(save_model_ckpt.data()),
+          this_blk_send_size, partner_trainer, 0,
+          reinterpret_cast<El::byte*>(load_model_ckpt.data()),
+          this_blk_recv_size, partner_trainer, 0,
+          El::SyncInfo<El::Device::CPU>{});
```
timmoon10, line 551:
```suggestion
        comm.sendrecv(
          reinterpret_cast<El::byte const*>(save_model_ckpt.data() + save_model_ckpt.size() - send_size),
          this_blk_send_size, partner_trainer, 0,
          reinterpret_cast<El::byte*>(load_model_ckpt.data() + load_model_ckpt.size() - load_size),
          this_blk_recv_size, partner_trainer, 0,
          El::SyncInfo<El::Device::CPU>{});
```
Also, does sendrecv gracefully handle the case where one of the buffers is empty? If we want to be paranoid, we could replace it with a send or recv.

src/callbacks/ltfb.cpp
```diff
@@ -34,283 +33,551 @@
...
-    p.open_restart(recv_dir);
-    m.load_from_checkpoint_shared(p);
-    p.close_restart();
+    restore_model_weights(m, restore_weights);
   }
+private:
+  std::string ckpt_basedir_;
+};// class CheckpointFile
+
+class CheckpointBlob final
```
timmoon10, line 484: As compelling as "blob" is, it's better to have consistent naming:
```suggestion
class CheckpointBinary final
```

timmoon10 (changes requested):
The class-based approach for communication algorithms is much cleaner than the previous enum-based approach. I think there are some correctness issues with very large models though.

src/callbacks/ltfb.cpp
benson31: That was my intent; brain fail. Thanks!

src/callbacks/ltfb.cpp
benson31: Oh yeah. One also has to bump the pointers?? Epic brain fail. Solving this in a different way.

benson31: Based on a test with MVAPICH2, it handles the case where one of the buffers is empty. I did test this idea and it worked for the cases (a) both buffers same size, (b) buffers off by one block, (c) one buffer completely empty. Apparently I just failed to copy-paste from my prototype code... A lot...

timmoon10 (approved):
LGTM

samadejacobs (approved):
Ran fine, sort of, the hangs issue (with and without fixed batch size) is independent of this.

bvanessen (approved):

PR #1743, opened 2021-01-07
Refactor - Port to MIOpen

mrwyattii:
Adds support for running LBANN with MIOpen:

- Duplicates the CUDNN-specific `dnn_lib` stuff for MIOpen
- CUDNN and MIOpen are aliased by the `dnn_lib` namespace
- Adds `dnn_lib` unit tests

PR #1746, opened 2021-01-09
Fix bug in cuDNN filter desc wrapper copy constructor

timmoon10:
I've found that [`cuDNNGetFilterNdDescriptor`](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnGetFilterNdDescriptor) will return `CUDNN_BAD_PARAM` if we pass in `nbDimsRequested=0`. I've modified the copy constructor for `lbann::dnn_lib::cudnn::FilterDescriptor` to handle this. We may want to make a similar change to other cuDNN objects, e.g. `TensorDescriptor`.

This bugfix emerged out a bug @samadejacobs noticed in LTFB, so I've also updated the unit test to handle the recent PR that added a "tournament" execution mode.

### DIFF
--- a/bamboo/unit_tests/test_unit_callback_ltfb.py
+++ b/bamboo/unit_tests/test_unit_callback_ltfb.py
@@ -113,6 +113,8 @@ def construct_data_reader(lbann):
 
     """
 
+    # Note: The training data reader should be removed when
+    # https://github.com/LLNL/lbann/issues/1098 is resolved.
     message = lbann.reader_pb2.DataReader()
     message.reader.extend([
         tools.create_python_data_reader(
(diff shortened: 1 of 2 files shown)


mrwyattii (approved):
LGTM

mrwyattii:
@timmoon10 have you tested if this bug arises for `cuDNNGetTensorNdDescriptor`?  I will also look into whether this is a necessary change for the MIOpen port #1743

bvanessen (approved):

timmoon10:
@mrwyattii I haven't tested with `cuDNNGetTensorNdDescriptor`.

PR #1750, opened 2021-01-12
Remove LBANN-written python module in favor of CMake module

benson31:
@timmoon10 Please verify this works for your workflow. Please let me know if you have any issues and what they are.


timmoon10 (approved):
:shipit:

PR #1751, opened 2021-01-13
Refactor OpenCV Optional Dependency

mrwyattii:
This PR makes OpenCV an optional dependency and adds an `LBANN_WITH_VISION` option for enabling/disabling OpenCV.  I've kept most of the changes in CMake files, with a few `#ifdef LBANN_HAS_OPENCV` in source and header files where necessary.


benson31 (approved):
Looks good :shipit:

timmoon10 (approved):
:shipit:
I'm not thrilled with the name `LBANN_HAS_VISION` since we still have non-OpenCV vision operations (e.g. convolution, cropping) that aren't affected by this flag. We've already discussed this though, and it's not a major deal.

PR #1753, opened 2021-01-15
Update Python frontend to correctly specify procs_per_trainer

timmoon10:
We've recently moved `procs_per_trainer` out of the trainer and required that it be passed directly into the LBANN driver. This PR updates the Python front-end to reflect this change, mainly by removing the option from the `Trainer` class. For convenience, I've also added a `procs_per_trainer` option to the run scripts.


samadejacobs (approved):
LGTM

PR #1754, opened 2021-01-16
Add option to specify file format of "dump weights" callback

timmoon10:
Options are text, binary, and distributed binary files. The latter is intended to avoid hangs when dealing with large weight tensors, e.g. with more than 2^32 entries.

An API change: The file names are just the weight name. Previously, they included the weight matrix dimensions.


timmoon10:
[https://lc.llnl.gov/bamboo/browse/LBANN-TIM333-1](No new Bamboo failures).

PR #1760, opened 2021-01-23
Fixed preprocessor guards for building without CNPY library for some

bvanessen:
data reader headers and utility functions.


benson31 (approved):
:shipit:

PR #1763, opened 2021-01-26
oneDNN support on CPU

benson31:


timmoon10 (changes requested):
Overall looks good. The only critical change is that we don't initialize the workspace for the GPU model-parallel impl.

benson31:
It occurred to me that I have not yet run Bamboo on this. This is in progress. I will comment again when it's through.

timmoon10 (approved):
:shipit: , pending no Bamboo problems.

benson31:
@bvanessen @mrwyattii This is passing Bamboo for me.

bvanessen (approved):

PR #1766, opened 2021-01-28
Scatter and gather layers

timmoon10:
This is an implementation of scattering and gathering with 1D tensors:
```
Scatter: y[ind[i]] = x[i]
Gather: y[i] = x[ind[i]]
```
Each layer takes two input tensors (values and indices) and outputs one tensor. This may be useful for applications that involve sparsity or categorical data (pinging @szaman19). 

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM338-1).


timmoon10:
[It looks like this is breaking the GRU unit tests in Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM338-LAS-3/test/case/86209274). This is odd since we don't touch it at all.

timmoon10:
It turns out the GRU failure is unrelated to this PR (see PR #1785). When I [run Bamboo with that fix](https://lc.llnl.gov/bamboo/browse/LBANN-TIM338-5), I see no new failures.

PR #1769, opened 2021-02-02
Clean up handling of GPU errors

benson31:
This is the LBANN part of a cleanup that will also occur in Hydrogen and Aluminum.

The primary change is to remove `cudaDeviceReset()` from the error handling macros. Since I was there, I cleaned up the messy error string building code, too.

[cudaDeviceReset](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gef69dd5c6d0206c2b8d099abac61f217) is a very big hammer, causing all allocations to be destroyed and cleans up "all resources". Our error checking macros will call this, but then they will throw an exception. This has a few issues:

1. Since all allocations are destroyed, any attempt to recover from this error would be futile. If a fatal error occurs (i.e., one from which recovery is not meant to be possible), throwing an exception is not the correct behavior (maybe: print error message, print stack, `MPI_Abort`/`std::abort`).
2. While the exception is finding its way up the stack, there will still be runtime calls, say, in destructors that will try to deallocate an objects internal (GPU) memory buffers. But these have already been deleted, so it's likely that a user will get a cryptic error message similar to:
```cpp
{
    File: /path/to/hydrogen/include/El/core/Memory/impl.hpp
    Line: 177
    Mesg: CUDA error detected in command: "hydrogen::cub::MemoryPool().DeviceFree(ptr)"

    Error Code: 400
    Error Name: cudaErrorInvalidResourceHandle
    Error Mesg: invalid resource handle
}
```
This looks very bad and very scary, and it is, but it's also highly likely that it is not the "real" error. Moreover, the initial error that triggered this whole chain of events may never be reported anyway, rendering the entire reporting infrastructure essentially moot. To get any real information, a user would have to run through a debugger with breakpoints set correctly, and we may has well have printed "An error occurred. Ask GDB about it."

Anyway, this just removes the `cudaDeviceReset()` so the "real" error has a chance to make it to the `catch` clause (or to make its way out of `main()`, where it will likely be reported by the runtime).

### DIFF
--- a/include/lbann/utils/dnn_lib/cudnn.hpp
+++ b/include/lbann/utils/dnn_lib/cudnn.hpp
@@ -38,10 +38,9 @@
   do {                                                          \
     const cudnnStatus_t status_CHECK_CUDNN = (cudnn_call);      \
     if (status_CHECK_CUDNN != CUDNN_STATUS_SUCCESS) {           \
-      cudaDeviceReset();                                        \
-      LBANN_ERROR(std::string("cuDNN error (")                  \
-                  + cudnnGetErrorString(status_CHECK_CUDNN)     \
-                  + std::string(")"));                          \
+      LBANN_ERROR("cuDNN error (",                              \
+                  cudnnGetErrorString(status_CHECK_CUDNN),      \
+                  ")");                                         \
     }                                                           \
   } while (0)
 #define CHECK_CUDNN_DEBUG(cudnn_call)                           \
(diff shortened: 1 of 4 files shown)


timmoon10 (approved):
LGTM

mrwyattii (approved):
LGTM and this addresses problems I was having when debugging LBANN-core code!

PR #1772, opened 2021-02-04
LBANN-core

mrwyattii:
Adds a minimal interface for downstream applications to load trained LBANN models and perform inference on a set of samples:

- Adds `batch_functional_inference_algorithm` execution algorithm for running inference on a model with provided samples
- Modifies `input_layer` to allow the execution algorithm to insert sample tensors without a data coordinator and data reader
- Adds `initialize_lbann` and `finalize_lbann` that initialize (and finalize) LBANN with a provided MPI_COMM
- Adds `load_inference_model` and `infer` as helper-functions for using new functionality
- Adds core-driver application to demonstrate how downstream applications can use the new inference algorithm
- Adds a unit test for `batch_functional_inference_algorithm`; We test the insertion of data into the input layer and verify that the correct categorical labels are returned for inference


applications/ATOM/utils/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ add_executable(compute_vocab
 install( 
   TARGETS compute_vocab
   OPTIONAL
-  EXPORT LBANNTargets
+  #EXPORT LBANNTargets
```
benson31, line 13:
```suggestion
```

benson31 (approved):
Just minor comments.

mrwyattii:
Made minor changes based on @benson31 feedback, including moving the core driver application to using the LBANN argument parser.

include/lbann/execution_algorithms/batch_functional_inference_algorithm.hpp
```diff
@@ -48,35 +58,110 @@ class batch_functional_inference_algorithm {
...
-        size_t mbs=0);
+  /** @brief Run model inference on samples and return predicted categories.
+   * @param[in] model A trained model
+   * @param[in] samples A distributed matrix containing samples for model input
+   * @param[in] mbs The max mini-batch size
+   * @return Matrix of predicted labels (by index)
+   */
+  template <typename DataT, El::Dist CDist, El::Dist RDist, El::DistWrap DistView, El::Device Device>
+  El::Matrix<int, El::Device::CPU>
+  infer(observer_ptr<model> model,
```
bvanessen, line 78: I don't think that we need to do this for this PR, but we should move most of this to an impl.hpp file.

bvanessen (approved):

PR #1773, opened 2021-02-05
Make OpenCV optional when building Catch2 tests

timmoon10:
I run into build errors when I disable OpenCV and enable Catch2. A minor tweak to the CMake gets it to work for me.

[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM340-1).


benson31 (approved):
LGTM

PR #1774, opened 2021-02-05
Add clang-format specification

benson31:
A good time to merge this sort of PR will be at the release of v1.0.

This will automatically add a target to the build system called `clang-format`. This will call `clang-format` on every file that's registered as a source file of the `lbann` target.

I recommend that we add a hook to run `clang-format` as a part of the commit process for `develop`.

The style is sort of a mixture of the base LLVM style with 2-space indents with a few aspects of Allman-style bracing. I will argue for class and function breaks before braces:
```cpp
class foo
{
...
};

void bar()
{
}
```
but I don't care as much about nested blocks such as conditionals or for loops (I still prefer the break, but I think this is a reasonable compromise):
```cpp
// This is fine
if (whatever) {
}

// As is this
for (auto x : y) {
}
```
Most of the other main changes it makes is a consistent formatting style for arguments, initializers, and inheritance, as well as a consistent ordering of `#include` statements.

See [the clang10 docs](https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormatStyleOptions.html) for info about what each key in `.clang-format` means.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,165 @@
+###############################################################################
+# Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+# Produced at the Lawrence Livermore National Laboratory.
+# Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+# the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+#
+# LLNL-CODE-697807.
+# All rights reserved.
+#
+# This file is part of LBANN: Livermore Big Artificial Neural Network
+# Toolkit. For details, see http://software.llnl.gov/LBANN or
+# https://github.com/LLNL/LBANN.
+#
+# Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+# may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at:
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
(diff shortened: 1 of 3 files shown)


benson31:
One thing I would like to propose (this is not in this format spec) is one argument per line. As we interact with ~ancient technology~ C APIs every now and again and these tend to be ~obnoxious and terrible~ very long and convoluted, I find it nice to have things a bit more expanded:
```cpp
  foo(a,
      b);// A little silly, but not _that_ bad

  // I find this structure overwhelming.
  // As the argument count increases, errors and typos are harder to spot
  some_c_func(some_scalar, descriptor, tensor_name, another_descriptor, another_tensor, 
              another_scalar, yet_another_descriptor, yet_another_tensor, param, another_param);

  // IMO, this structure makes it easier to count arguments (think: "error: no match for argument 8")
  // and it just makes it easier to read.
  some_c_func(some_scalar,
              descriptor,
              tensor_name,
              another_descriptor,
              another_tensor,
              another_scalar,
              yet_another_descriptor,
              yet_another_tensor,
              param,
              another_param);
```
The extra vertical space should not be of much concern, as functions that take this many arguments should be nearly exclusively C API calls (it's less-than-ideal practice for C++ APIs to be this bad (it's probably a symptom of poor data encapsulation)), and those should be few and far between (and increasingly isolated to "backend" code).

Thoughts?

ndryden:
I'd be fine with these changes. It will be good to have a standardized format. Definitely need the commit hook to keep things neat.

benson31:
#1804 applied this format to {comm.hpp, comm_impl.hpp, comm.cpp}, for a concrete look at the style.

bvanessen:
@benson31 What do we need to do to finish this up?

benson31:
Just click merge.

PR #1778, opened 2021-02-08
Link/include embedded Python lib with CMake

timmoon10:
I've been running into build errors where the compiler can't find `Python.h`. It turns out [FindPython.cmake](https://cmake.org/cmake/help/latest/module/FindPython.html) exposes a target for the Python/C API. Linking against this in CMake fixed the issue for me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -407,7 +407,7 @@ endif ()
 
 # Find Python
 if (LBANN_WITH_PYTHON)
-  find_package(Python REQUIRED COMPONENTS Interpreter Development)
+  find_package(Python REQUIRED COMPONENTS Interpreter Development.Embed)
   set(LBANN_HAS_PYTHON "${Python_FOUND}")
   if (NOT Python_VERSION_MAJOR EQUAL 3)
     set(LBANN_HAS_PYTHON FALSE)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -629,6 +629,9 @@ if (LBANN_HAS_OPENCV)
   target_link_libraries(lbann PUBLIC ${OpenCV_LIBRARIES})
 endif ()
 target_link_libraries(lbann PUBLIC ${CONDUIT_LIBRARIES})
+if (Python_Development.Embed_FOUND)
+  target_link_libraries(lbann PUBLIC Python::Python)
```
benson31, line 633: This library is [already linked](https://github.com/LLNL/lbann/blob/develop/CMakeLists.txt#L644). Is it sufficient to just make the `Development.Embed` change above?

timmoon10: Hm, I wonder why it wasn't working for me before. `Development.Embed` should be automatically activated when the `Development` module is chosen, so that change was not strictly needed.

timmoon10:
I've rebuilt with just the `Development.Embed` change and the original, and it's all working for me now. Build systems work in mysterious ways.

PR #1781, opened 2021-02-09
fixed bug causing segfault during cudnnHandle destructor

mrwyattii:
Calling `cudnnDestroy` on the cudnn handle was causing a segfault during the `[dnn_lib]` unit tests.  I added a call to `dnn_lib::destory()` that frees the unique ptr for the handle and avoids the segfault.  Also changed `LBANN_HAS_GPU` to `LBANN_HAS_DNN_LIB` because we are testing functions in `dnn_lib::`.


unit_test/SequentialCatchMain.cpp
```diff
@@ -26,12 +26,14 @@
 
 #define CATCH_CONFIG_RUNNER
 #include <catch2/catch.hpp>
+#include <lbann/utils/dnn_lib/helpers.hpp>
 #include <lbann/utils/random_number_generators.hpp>
 
 int main(int argc, char* argv[]) {
-#ifdef LBANN_HAS_GPU
+#ifdef LBANN_HAS_DNN_LIB
   hydrogen::gpu::Initialize();
```
benson31, line 34: We should probably explicitly call `dnn_lib::initialize()`, too, for consistency if nothing else.
```suggestion
  hydrogen::gpu::Initialize();
  dnn_lib::initialize();
```

benson31 (approved):

PR #1784, opened 2021-02-09
Added support for the compiler tests to use the spack build script.

bvanessen:
Also setup the run script to find the spack installed lbann executable.


bvanessen:
@benson31 We need to change how the tests deal with failures to ensure that LBANN gets installed no matter if the `catch` tests fail or not.

PR #1792, opened 2021-02-17
Updated the build script for ROCm on Corona

bvanessen:
Added external paths for the HIP/ROCm tools.

### DIFF
--- a/scripts/customize_build_env.sh
+++ b/scripts/customize_build_env.sh
@@ -54,7 +54,7 @@ set_center_specific_gpu_arch()
                 GPU_ARCH_VARIANTS="cuda_arch=60"
                 CMAKE_GPU_ARCH="60"
                 ;;
-            "zen2")
+            "zen" | "zen2")
                 GPU_ARCH_VARIANTS="amdgpu_target=gfx906"
                 ;;
             *)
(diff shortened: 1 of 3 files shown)


scripts/customize_build_env.sh
```diff
@@ -78,8 +81,8 @@ set_center_specific_modules()
             "ivybridge")
                 MODULE_CMD="module --force unload StdEnv; module load gcc/8.3.1 mvapich2/2.3 python/3.7.2"
                 ;;
-            "zen")
-                MODULE_CMD="module --force unload StdEnv; module load gcc/8.3.1 mvapich2/2.3 python/3.7.2"
+            "zen2")
+                MODULE_CMD="module --force unload StdEnv; module load gcc/8.3.1 mvapich2/2.3 python/3.7.2 opt rocm/3.10.0 hwloc/2.1"
```
mrwyattii, line 85: We will need hwloc@2.3+ because of this include: https://github.com/LLNL/Aluminum/blob/5dcf8756285caece49ffee97ce35d9d8d109a777/src/progress.cpp#L35

scripts/customize_build_env.sh
```diff
@@ -180,8 +187,47 @@ set_center_specific_externals()
...
+    rdma-core:
+      buildable: False
+      version:
+      - 20
+      externals:
+      - spec: rdma-core@20 arch=${spack_arch}
+        prefix: /usr
+EOF
+                ;;
+            "zen2")
```
mrwyattii, line 201: `spack arch -t` returns `zen` for me on corona.  We should probably have this as `"zen" | "zen2"`

bvanessen: Yes, it depends on if you are on the head node or compute node.

PR #1796, opened 2021-02-24
CPU GRU layer

timmoon10:
This PR implements a CPU GRU layer using oneDNN. I've also done some refactoring to better encapsulate the cuDNN and oneDNN implementations. It takes advantage of some of the generic `dnn_lib` functionality, but it is still rather messy and intertwined with oneDNN and cuDNN. A CPU-only build on Pascal passes the GRU unit test (metric checking and gradient checking).

Miscellaneous comments on oneDNN GRUs for future reference:
- If the GRU has >1 layer, then the input and hidden vector sizes must be identical.
- In order to match cuDNN, we use the Linear-Before-Reset GRU instead of the vanilla GRU.
- The matrix weights are expected in LDIGO format in forward prop and LDGOI in backprop. The matrix weight gradients are expected in LDIGO format. I don't believe this is documented anywhere.
- The matrix weights have a gate order of {update, reset, output}. LBANN, PyTorch, and cuDNN use {reset, update, output}.
- There is a single set of biases with a gate order of {update, reset, output, before-reset}. LBANN and PyTorch have separate "ih_bias" and "hh_bias" with a gate order of {reset, update, output}.
- oneDNN is not big on informative error messages. Setting `export DNNL_VERBOSE=2` will print out info as primitives execute, but it was not much of a help since most of my problems happened when constructing primitive descriptors (with a message of "could not create a primitive descriptor iterator").

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM348-1). Given our deadlines, it should be fine to merge this immediately if Bamboo doesn't show anything worrying.


timmoon10:
No new Bamboo failures. Keep in mind the CPU version isn't being tested by Bamboo since I'm not sure if the Catalyst and Corona builds include oneDNN support.

benson31:
They do not include oneDNN support. I consider it "experimental" right now. When we move to more flexible CI, it would be good to test with and without it.

benson31 (approved):
This looks good to me. I was able to build locally. As there are no new Bamboo failures, I strongly vote we merge this now to accommodate deadlines, even if we find issues later on.

bvanessen (approved):
LGTM

PR #1797, opened 2021-02-24
Separate GRU layers in ATOM WAE decoder

timmoon10:
Due to limitations in oneDNN, the CPU GRU layer in PR #1796 does not support multi-layer GRUs with different input and hidden vector sizes. The decoder RNN in the ATOM WAE model has three layers and converts from a latent space to a character embedding space. The easiest workaround is to just split up the decoder into three separate GRU layers. Alternatively, it would suffice to just split the first GRU layer from the other two.

I just made a beeline to get this working on CPU with synthetic data. I haven't checked correctness yet, although it's such a simple change I don't expect problems. Note that this will significantly hurt performance on GPUs. It may be worthwhile adding options so that we can construct the WAE with or without a stacked RNN.


bvanessen:
@timmoon10 @samadejacobs I haven't look too closely at this yet, but shouldn't we create a new WAE model instantiation one for stacked and one for non-stacked so that we don't lose the stacked capability?

samadejacobs:
@timmoon10, let's make this optional; preferably stacked for NVIDIA GPU and unstacked (this PR) for CPU or other accelerators (may be). The option (a flag in argument parser in the driver possibly) should be transparent to user.

bvanessen (approved):

PR #1798, opened 2021-02-25
Implement ATOM WAE reconstruction loss with gather layer

timmoon10:
When training the ATOM WAE model, we compute the cross entropy loss for each character in the output sequence. Our old approach involved computing the log-softmax and applying a mask that was built out of one-hot vectors (one per sequence position). This incurred the overhead of launching many small kernels to construct the one-hot vectors. Since we have implemented a gather layer (PR #1766), we can replace all of the one-hot vectors with a single gather layer. Note that this is mathematically equivalent.

Running with synthetic data on a Lassen node, per-step time drops from 0.021 sec to 0.018 sec (1.14x speedup). Running a CPU build on a Pascal node (using PR #1797), there is not a significant change in per-step time (1.5 sec). The reconstruction losses are similar between the new and old implementations (except on GPU after it goes below the eps for half-precision).

### DIFF
--- a/applications/ATOM/eval_atom_wae.py
+++ b/applications/ATOM/eval_atom_wae.py
@@ -150,7 +150,8 @@ def construct_model(run_args):
         for idx in range(len(l.weights)):
           l.weights[idx].optimizer = lbann.NoOptimizer()
       weights.update(l.weights)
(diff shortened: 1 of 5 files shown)


samadejacobs (approved):
LGTM. Same optimization will nice for VAE, not required for this PR though.

PR #1799, opened 2021-02-27
Add OpenMPI launcher to Python front-end

timmoon10:
This is an initial implementation of a launcher for OpenMPI and its various forks. I'd prefer to have a generic MPI launcher, but as far as I can tell the standard offers no way to specify process placement. I'd be satisfied calling it an "MPI launcher" if it only worked for MPICH and OpenMPI, but their `mpiexec`/`mpirun` arguments don't seem compatible.


bvanessen (approved):
LGTM

PR #1801, opened 2021-03-02
Batch-wise reduce-sum layer

timmoon10:
Some of our applications require batch-wise metrics, e.g. R2 over a mini-batch. In principle, we can implement a lot of these with a sum that goes over the batch dimension (e.g. `torch.sum(x, dim=0)` in PyTorch). For example, if we want to compute the (biased) variance of a scalar over a mini-batch, we can now do:
```
# var(x) = mean(x^2) - mean(x)^2
mini_batch_size = lbann.MiniBatchSize()
mean = lbann.Divide(lbann.BatchwiseReduceSum(x), mini_batch_size)
sqmean = lbann.Divide(lbann.BatchwiseReduceSum(lbann.Square(x)), mini_batch_size)
var = lbann.Subtract(sqmean, lbann.Square(mean))

# var(x) = mean( (x-mean(x))^2 )
# Note: This is better than the above for numerical accuracy
mini_batch_size = lbann.MiniBatchSize()
mean = lbann.Divide(lbann.BatchwiseReduceSum(x), mini_batch_size)
var = lbann.Divide(lbann.BatchwiseReduceSum(lbann.Square(lbann.Subtract(x, mean))), mini_batch_size)
```
Note that this layer requires blocking communication, so it will hurt performance.

[No new Bamboo errors](https://lc.llnl.gov/bamboo/browse/LBANN-TIM353-1).

### DIFF
--- a/bamboo/unit_tests/test_unit_layer_batchwise_reduce_sum.py
+++ b/bamboo/unit_tests/test_unit_layer_batchwise_reduce_sum.py
@@ -1,0 +1,155 @@
+import functools
+import operator
+import os
+import os.path
+import sys
+import numpy as np
+
+# Bamboo utilities
+current_file = os.path.realpath(__file__)
+current_dir = os.path.dirname(current_file)
+sys.path.insert(0, os.path.join(os.path.dirname(current_dir), 'common_python'))
+import tools
+
+# ==============================================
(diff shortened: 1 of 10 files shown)


samadejacobs (approved):
LGTM

PR #1803, opened 2021-03-02
Add Batch Inference Algorithm

mrwyattii:
Adds a `batch_inference_algorithm` to `execution_algorithms` and also adds `execution_algorithm` class such that:
```
                     execution_algorithm
                       /            \
batch_inference_algorithm   training_algorithm
                                      |
                                    SGD
```

The `batch_inference_algorithm` takes in a model and data source and then runs forward prop.  The data source can be either a `data_coordinator` or distributed matrix.  For the distributed matrix case, the functions are incomplete until a future PR which adds the ability to insert samples into input layers without a `data_coordinator`.


mrwyattii:
Based on a conversation with @benson31 , I have remove the `execution_algorithm` class as it doesn't add anything useful at this time.  If we find that we need it, it can be quickly implemented in a future PR

benson31 (approved):
LGTM

timmoon10:
I would think it would be better to flatten the hierarchy:
```
                    execution_algorithm
                    /                             \
batch_inference_algorithm       sgd_training_algorithm
```
"Training algorithm" isn't meaningful by itself, so might as well has SGD inherit directly from the base class.

mrwyattii:
I've made updates after our discussion:

- `execution_algorithm` is just the name of the directory containing the inference and training algorithms (and no longer a class)
- The inference algorithm is now called `batch_functional_inference_algorithm`
- The inference algorithm no longer supports data from a `data_coordinator`.  It's not simplified to a single `infer` and `infer_mini_batch` that expect samples to be passed in via `El::AbstractDistMatrix`
- I've added a function that returns the label value given a softmax output

This compiles, but I have not properly tested.  I'll bring these changes into #1822 and test with an external driver application.

bvanessen (approved):

PR #1805, opened 2021-03-03
Fix error when not building with FP16 support

benson31:
Affects the Catch2 optimizer unit tests.

### DIFF
--- a/src/optimizers/unit_test/optimizer_common.hpp
+++ b/src/optimizers/unit_test/optimizer_common.hpp
@@ -28,18 +28,21 @@ std::string desc_string(ObjectType const& opt)
 // archive, the cadr is the input archive. Accessor metafunctions are
 // defined below.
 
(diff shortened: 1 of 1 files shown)


bvanessen (approved):

src/optimizers/unit_test/optimizer_common.hpp
```diff
@@ -28,18 +28,21 @@ std::string desc_string(ObjectType const& opt)
 // archive, the cadr is the input archive. Accessor metafunctions are
 // defined below.
 
-using Fp16Types = TL<lbann::cpu_fp16
+using FpTypes = TL<float, double,
```
timmoon10, line 31:
```suggestion
using FpTypes = TL<float, double
```

timmoon10 (changes requested):

timmoon10 (approved):
:shipit:

PR #1809, opened 2021-03-04
Added support for the Fugaku PJM launcher.

bvanessen:


timmoon10 (approved):
Overall looks good. It would be better if we removed that Fugaku-specific incantation and put it in a contrib launcher for R-CCS systems.

PR #1811, opened 2021-03-05
Add initial mockup of transform documentation

benson31:
I'm not volunteering to flesh these out, but this is a shell to get someone started.

Question: Do we like the single file (i.e., a single web page), or would we prefer multiple files (i.e., multiple web pages). AFAIK, there's no way to break the 1:1 file:web page relationship.


benson31:
Unrelated to the content of this PR, we should get around to exposing Transforms to the python front-end.

ndryden:
I think one page is fine for these.

timmoon10 (approved):
LGTM

bvanessen (approved):

PR #1807, opened 2021-03-03
Cleanup of some of the comm guts; fix const correctness issues.

benson31:
This twiddles some of the API decorations in the `lbann_comm` class. It also aggressively pushes `const` parameters through the implementation of the `lbann_comm` interface.

PR #1813, opened 2021-03-05
Actually-non-blocking allreduce without Aluminum

benson31:
This sort of punts on (a) noncontiguous matrices (but I don't see those showing up, at least not in the ATOM model I'm testing with) and (b) GPU matrices (but in this case one _should_ be using Aluminum or just accept bad performance).

I'm slightly torn: it makes some sense to migrate this code into Hydrogen so it's hidden a bit better and then all the actual MPI calls are coming from Hydrogen. On the other hand, `libHydrogen` doesn't need nonblocking collectives for its own purposes. Any thoughts?

### DIFF
--- a/include/lbann/comm.hpp
+++ b/include/lbann/comm.hpp
@@ -95,6 +95,7 @@ struct request
   mpi_req_type mpi_req = mpi_null_req;
   nccl_req_type nccl_req = nccl_null_req;
   mpicuda_req_type mpicuda_req = mpicuda_null_req;
+  MPI_Request raw_mpi_req = MPI_REQUEST_NULL;
 };
 
 } // namespace Al
--- a/include/lbann/comm_impl.hpp
+++ b/include/lbann/comm_impl.hpp
@@ -718,8 +718,8 @@ void lbann_comm::nb_allreduce(T* data,
                               Al::request& req,
                               const El::mpi::Op op) const
 {
-#ifdef LBANN_HAS_ALUMINUM
   m_bytes_sent += count * sizeof(T);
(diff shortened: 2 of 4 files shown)


benson31:
Depends on #1807.

timmoon10 (approved):
This seems like a sensible approach to make progress. I figure if we want non-blocking communication in Hydrogen, we'll want to take the effort to the get the API right.

If I understand correctly, replacing the `MPI_Wait` with a spinlock on `MPI_Test` would avoid deadlocks. If we make sure the data store only communicates on thread 0 and uses its own communicator, then the multiple threads should never interfere with each other (thanks to `MPI_THREAD_SERIALIZED`) and we should be golden, right?

benson31:
The tricky bit is non-blocking comms via Aluminum in Hydrogen (we need to extend `El::mpi::Request` to hide the Aluminum requests -- this isn't a big deal, it's just the bulk of the work). The other issue is nonblocking communication of noncontiguous matrix data requires *either* custom vector data types (representing columns) *or* a way to manage the temporary memory if we use the copy-to-contiguous-copy-and-communicate-the-contiguous-buffer approach (which is what the blocking matrix allreduce does, for example).

My initial local (rather small) tests show that the spin-`Test` does, indeed, resolve the deadlock that I was seeing, if all the other `lbann_comm` interface is locked. @bvanessen is working on sorting out data communication stuff. In theory, we might actually be able to get to a `MPI_THREAD_SERIAL`-compatible mode without special care or explicit locks.

ndryden (approved):

bvanessen (approved):

timmoon10 (approved):

PR #1814, opened 2021-03-07
hdf5 reader documentation: all about the yaml files

davidHysom:
draft


bvanessen (approved):

docs/data_ingestion.rst
```diff
@@ -0,0 +1,39 @@
...
+into distinct objects, but it is still a work in progress.  As a
+result there are some "legacy" data readers that represent both of
+these features, and some new data readers that focus more on task 1
+and incorporate the use of a sample list to help with task 2.
+
+At this time, LBANN can only ingest static data sets; work on
+streaming data is in progress.
+
+
+Legacy Data Readers
```
benson31, line 24: Is this information really so significant that it needs its own section? It's not that interesting, and a sentence per category could just be appended to the paragraph above.

benson31 (approved):
This is probably fine.

docs/data_ingestion.rst
bvanessen: I was planning to fill it out more later.

PR #1815, opened 2021-03-08
Fix bug in MIOpen dropout

mrwyattii:
The `noise_shape` tensor descriptor passed to the MIOpen `dropout{Forward,Backward}` was causing a nullptr dereference error.  The `noise_shape` input is not used according to MIOpen docs: https://rocmsoftwareplatform.github.io/MIOpen/doc/html/dropout.html#miopendropoutforward

I replaced the `nullptr` with the input tensor descriptor for `noise_shape` to fix this bug.


benson31 (approved):
LGTM :shipit:

PR #1819, opened 2021-03-09
Handle weights with dims >2.1B

timmoon10:
@KIwabuchi has observed integer overflow errors when running node2vec on very large graphs. The fix is to store weight dims in `size_t` instead of `int`. With this change, I've been able to run on Lassen with 5B embeddings and synthetic data. The GPU implementation runs without problem, but the CPU implementation fails with an OpenSHMEM error. I don't see anything substantially different between the CPU and GPU implementations, so maybe it's something in OpenMPI OpenSHMEM?

This is making progress toward #91. [Bamboo is failing](https://lc.llnl.gov/bamboo/browse/LBANN-TIM355-1/artifact) because I forgot the unit tests, but the main code should be ready for review.

### DIFF
--- a/include/lbann/layers/learning/channelwise_scale_bias.hpp
+++ b/include/lbann/layers/learning/channelwise_scale_bias.hpp
@@ -178,7 +178,7 @@ void channelwise_scale_bias_layer<TensorDataType, Layout, Dev>
   auto dist = this->get_prev_activations().DistData();
   dist.colDist = El::STAR;
   dist.rowDist = El::STAR;
-  this->get_weights(0).set_dims({static_cast<int>(num_channels)}, {2});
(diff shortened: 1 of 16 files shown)


timmoon10:
Right now it's only with SHMEM and the message sizes are still small. The offsets are huge, but that's just pointer arithmetic that should be robust to integer overflow (assuming the SHMEM implementation is correct). It's worth keeping the MPI issues in mind though, although I think we're a while away from hitting them.

timmoon10:
[No new Bamboo errors](https://lc.llnl.gov/bamboo/browse/LBANN-TIM355-3).

benson31 (approved):
:shipit:

PR #1822, opened 2021-03-10
Modify input layer to load samples without DC

mrwyattii:


mrwyattii:
Closed, changes are included in #1772

PR #1823, opened 2021-03-11
Implementation of perturb learning rate callback

samadejacobs:
Callback for learning rate exploration in tournament (population)-based training. Current implementation supports 
random perturbation, performed in log space.


src/callbacks/perturb_lr.cpp
```diff
@@ -0,0 +1,159 @@
...
+     CEREAL_NVP(m_weights_names));
+}
+
+void perturb_lr::setup(model* m) {
+  perturb(*m);
+}
+
+void perturb_lr::on_batch_begin(model* m) {
+  const auto& c = m->get_execution_context();
+  if (m_perturb_during_training && c.get_step() > 0) {
```
timmoon10, line 74: We should only perturb once per interval:
```suggestion
  if (m_perturb_during_training
      && c.get_step() % m_batch_interval == 0
      && c.get_step() > 0) {
```

timmoon10 (changes requested):
Two requests:
- Change name to "perturb_learning_rate". The meaning of LR is not obvious to me.
- Make sure we don't perturb at every step (I now see why my experiments with perturbing Adam were so volatile...).

samadejacobs:
Two requests by @timmoon10 addressed in latest push.

timmoon10 (approved):
LGTM. If there aren't any new failures in Bamboo, we can merge.

PR #1824, opened 2021-03-11
Updated build environment

bvanessen:
Updated the build script to use the new environment aware feature of
Spack that allows the spack build-env lbann -- bash command to find
the proper version of LBANN that is in the active environment.  This
simplifies the user interface so that the develop does not have to
remember the full set of Spack variants.  As a result the build script
flag --build-env-only is deprecated.  Additionally, cleaned up the
implicit dependencies on Hydrogen and DiHydrogen to properly work when
Aluminum is disabled.

### DIFF
--- a/docs/building_lbann.rst
+++ b/docs/building_lbann.rst
@@ -193,7 +193,8 @@ platform and the nominal options in the CMake build environment.
 
     .. code-block:: console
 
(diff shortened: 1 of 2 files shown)


mrwyattii (approved):
tested and working, LGTM

samadejacobs:
WFM.

PR #1826, opened 2021-03-12
Use size_t for weights positions

timmoon10:
PR #1819 introduced some compile errors in the debug build since the weights positions (in `int`) don't match the dimensions (in `size_t`).

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM357-1). Pinging @davidHysom.


benson31 (approved):
LGTM.

PR #1832, opened 2021-03-16
Parallelize computation of Gaussian random numbers

timmoon10:
This replaces calls to `El::Gaussian` with an OpenMP-parallelized loop. Running the ATOM WAE on a Pascal node, I see a 3.6x speedup in the Gaussian layers.

I haven't properly tested this yet and I'd like to implement some basic normality tests in Catch. But deadlines being what they are, we can merge this in if we're willing to accept the risk of a bug.

### DIFF
--- a/include/lbann/utils/random.hpp
+++ b/include/lbann/utils/random.hpp
@@ -169,6 +169,20 @@ template <typename TensorDataType>
 void uniform_fill_procdet(El::AbstractDistMatrix<TensorDataType>& mat, El::Int m, El::Int n,
                           TensorDataType center = 0.0, TensorDataType radius = 1.0);
 
+/**
+ * Make mat into an m x n matrix where each entry is independently
+ * drawn from a Gaussian distribution with given mean and standard
+ * deviation. Entries are generated in parallel, so there are no
+ * guarantees of thread/process indendence.
+ */
+template <typename TensorDataType>
+void gaussian_fill_parallel(
+  El::AbstractDistMatrix<TensorDataType>& mat,
+  El::Int m,
+  El::Int n,
+  TensorDataType mean = 0.0,
+  TensorDataType stddev = 1.0);
+
 bool save_rng_to_checkpoint_shared(persist& p, lbann_comm* comm);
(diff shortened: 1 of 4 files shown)


timmoon10:
I've added unit tests to estimate whether `gaussian_fill` is actually outputting Gaussian numbers. It runs the [Anderson-Darling test](https://en.wikipedia.org/wiki/Anderson%E2%80%93Darling_test) three times for each matrix type and fails if it gets three 0.05 significance events. We do this for O(50) matrix types, so the probability of a fluky failure is <2%.

 In retrospect I think this approach is both overkill and insufficient. I'm not sure if it can detect some likely failures like duplicated values. But in any case, it gives us a sign that `gaussian_fill` is producing something that roughly looks like the desired distribution.

timmoon10:
It looks like [`fugaku-develop`](https://github.com/LLNL/lbann/tree/fugaku-develop) fails Bamboo due to a `-Werror` complaint. See: https://github.com/LLNL/lbann/blob/17cd056a1ce198d8b8397f89d811fc7d01c08095/src/data_readers/data_reader.cpp#L225

To keep this PR minimal, I've rebased and changed it to merge into `develop`.

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM359-2).

timmoon10:
[Bamboo is being uncooperative](https://lc.llnl.gov/bamboo/browse/LBANN-TIM362-1). I'm still in the process of debugging.

timmoon10:
[Latest Bamboo build](https://lc.llnl.gov/bamboo/browse/LBANN-TIM362-6).

The test failures for "Testing gaussian_fill" are fixed with https://github.com/LLNL/Elemental/pull/118. I think the compiler errors are related to Catch failures, since the compiler script also runs the Catch tests. I think this is ready to merge. If we want to be sure, we can rebuild Hydrogen for Bamboo and rerun the tests.

timmoon10:
[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM362-7). This is ready to go.

benson31 (approved):
Seems reasonable.

PR #1833, opened 2021-03-16
Fixed bug with MIOpen convolution algorithm selection

mrwyattii:
Initial implementation of the miopen algorithm selection treated the `miopenFindConvolution*Algorithm` functions the same as their cuDNN equivalents, causing an issue with algorithm selection.  Specifically, the `GEMM` convolution algorithms were always being selected, showing a measured performance time of "0", despite those being the worst performing algorithms.  Changes include:

- Remove the for loop of trials which repeatedly calls `miopenFindConvolution*Algorithm`.  This was done because repeated calls to the functions returned the same exact results.  Even with `MIOPEN_FIND_ENFORCE=3` (https://rocmsoftwareplatform.github.io/MIOpen/doc/html/perfdatabase.html#miopen-find-enforce) the exact same performance results are returned after the first call.
- Resize `perf_results` vector based on the number of algorithms tested.  The number of algorithms _tested_ is not guaranteed to be the number of algorithms _requested_ and sometimes the vector will contain `miopenConvAlgoPerf_t` structs with default values (e.g., GEMM with time "0").  Pruning the results to the number of algorithms tested fixes this problem.
- Added convolution algorithm selection to the `dnn_lib` unit test

A few more things to document/note:
- It seems that the MIOpen performance database should occasionally or always be forced to update by setting `MIOPEN_FIND_ENFORCE=4`.  Otherwise, the algorithm selection will pull from pre-calculated performance data in the DB
- It's unclear if `find_best_algorithm` and `find_best_heuristic_algorithm` are necessary.  The MIOpen functions return sorted vectors (based on time only) and selecting the first algorithm may be sufficient to replace those functions.

### DIFF
--- a/src/utils/miopen.cpp
+++ b/src/utils/miopen.cpp
@@ -1192,6 +1192,7 @@ miopenConvFwdAlgorithm_t get_fwd_algo_heuristic(
                 perf_results.data(),
                 ws, ws_size,
                 true));
+  perf_results.resize(num_tested_algos);
   return find_best_heuristic_algorithm(perf_results, nondet_fwd_algos,
                                        deterministic, ws_size);
 }
@@ -1220,6 +1221,7 @@ miopenConvBwdDataAlgorithm_t get_bwd_data_algo_heuristic(
                 perf_results.data(),
                 ws, ws_size,
                 true));
+  perf_results.resize(num_tested_algos);
   return find_best_heuristic_algorithm(perf_results, nondet_bwd_data_algos,
                                        deterministic, ws_size);
 }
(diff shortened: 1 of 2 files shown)


timmoon10 (approved):
:shipit: 

It is annoying that MIOpen doesn't match cuDNN here. I think the two use-cases we want are a cheap, compute-free heuristic and an exhaustive search. If MIOpen caches performance results, I suppose that's an _ok_ approximation of a heuristic. Forcing recalculation when doing an exhaustive search seems like the right behavior to me though.

benson31 (approved):
LGTM

PR #1834, opened 2021-03-18
Configure trainer grid dims with environment variable

timmoon10:
By default, Hydrogen initializes process grids to be as square as possible. However, we have some applications with very tall-skinny matrices that would benefit from making the process grid 1D. This PR adds the option to configure the grid height with the `LBANN_TRAINER_GRID_HEIGHT` environment variable.

@benson31 and I have talked about exposing this more visibly, perhaps in the Protobuf and Python interfaces. It's not a bad idea, although I'm not sure how to do it right yet. I think the future is moving toward per-layer, multi-dimensional process grids with user-specified or introspected dimensions, e.g. for distconv or sub-graph parallelism, so this will come up again.

[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM360-1).

### DIFF
--- a/include/lbann/comm.hpp
+++ b/include/lbann/comm.hpp
@@ -205,9 +205,9 @@ public:
     return get_rank_in_world() == get_world_master();
   }
   /** Return a grid to use for this trainer. */
-  inline El::Grid& get_trainer_grid() noexcept { return *m_grid; }
+  inline El::Grid& get_trainer_grid() { return *m_grid; }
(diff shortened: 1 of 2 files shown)


benson31 (approved):

bvanessen:
Why wasn't this added as a proper argument in our new command line argument parser?  That way it would be centralized, have documented help, and more maintainable.

PR #1836, opened 2021-03-20
Add command-line arg to set trainer grid height

timmoon10:
This is a followup to suggestions from PR #1834. It adds a command-line argument `--trainer_grid_height` to set the trainer grid height.

[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM363-1).


benson31 (approved):
LGTM

PR #1837, opened 2021-03-23
Optimizing tessellate, concatenate, and slice layers on CPU

timmoon10:
This PR is tinkering with the CPU implementations of the tessellate, concatenate, and slice layers to make them more amenable to vectorization. In particular, as much as possible, I remove function calls from inside inner loops. If profiling on Fugaku turns out well, this should be ready to merge.

[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM364-1).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 3.17)
 
 project(LBANN CXX)
 
+# C isn't required by LBANN but packages such as Conduit are forcing C dependencies in MPI
+enable_language(C)
+
 # Prevent in-source builds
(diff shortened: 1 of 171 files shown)


timmoon10:
Running the ATOM WAE model on a Fugaku node, I see a 1.5x speedup in the tessellate layer, 2.2x in the concat layer, and 1.9x in the slice slayer. The tessellate and concat layers still take a moderate amount of total runtime (7.5% and 4.4% respectively), but further gains will probably require more careful optimization of data access patterns or specialized optimizations.

Just in case, [Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM364-2).

PR #1839, opened 2021-03-24
CMake modifications inspired by work on Fugaku

benson31:
While these are driven by some of the issues we had on Fugaku, I believe they're relevant to `develop`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -698,7 +698,7 @@ target_link_libraries(lbann PUBLIC
 
 if (LBANN_FLAG_Werror_OK)
   target_compile_options(lbann
-    PRIVATE $<BUILD_INTERFACE:-Werror>)
+    PRIVATE $<$<COMPILE_LANGUAGE:CXX>:$<BUILD_INTERFACE:-Werror>>)
 endif ()
 
 if (LBANN_HAS_ALUMINUM)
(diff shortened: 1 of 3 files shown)


timmoon10 (approved):
LGTM

benson31:
Bamboo is happy enough for both CPU and GPU systems.

CMakeLists.txt
```diff
@@ -424,7 +397,8 @@ endif ()
 # TODO: This probably shouldn't be the case and should disable specific
 # callbacks when OpenCV is not available
 if (LBANN_WITH_TBINF AND NOT LBANN_HAS_OPENCV)
-  message(FATAL_ERROR "TensorBoard requires OpenCV. Enable LBANN_WITH_VISION to use TensorBoard.")
+  message(FATAL_ERROR "TensorBoard requires OpenCV. "
```
mrwyattii, line 400: Is this check necessary?  `LBANN_WITH_TBINF` is set to `LBANN_WITH_VISION` (line 135), so this *should* never be triggered

mrwyattii (approved):
Left one comment/suggestion, but overall looks good

CMakeLists.txt
benson31: That's the default value only. You might manually say `LBANN_WITH_VISION=ON`, `LBANN_WITH_TBINF=OFF` (this should be fine) but you could just as easily say: `LBANN_WITH_VISION=OFF`, `LBANN_WITH_TBINF=ON`. This is the case we're protecting against.

bvanessen (approved):

PR #1840, opened 2021-03-24
Manually set Spack spec for OpenBLAS on LC x86 systems

timmoon10:
This matches how we handle Power systems. See PRs #1792 and #1800.


timmoon10:
Deprecated by #1846.

PR #1843, opened 2021-03-30
Avoid GPU embedding and tessellate layers in ATOM WAE test

timmoon10:
These layers are correct, but have non-deterministic floating-point errors since they accumulate into atomics. The ATOM WAE test is incredibly sensitive, even when computing in double-precision, so bit-wise reproducbility is necessary. When I run the ATOM WAE test (1 Lassen node, deterministic debug build, double-precision data), I see a 5x slowdown relative to `develop`.


timmoon10:
If we'd like to avoid the performance hit, we could alternatively replace Adam with SGD. In my experiments, that makes the learning behavior much more stable.

bvanessen (approved):

PR #1845, opened 2021-03-31
Add GRU layer support to the K-FAC callback

oyamay:
This PR extends the K-FAC callback (#1634) to support GRU layers. 
* It supports both the cuDNN (GPU) and oneDNN (CPU) backends.
* It does not support (cuDNN's) stacked GRU layers at the moment.
* K-FAC is not applied to biases at the moment. `learning_rate_factor_gru` is multiplied to the learning rate only for six weight matrices of each GRU layer to use K-FAC and another optimizer for other layer types at the same time.


bvanessen:
@timmoon10 @benson31 I have rebased this on the head of develop so that it is easier to review.

timmoon10 (approved):
I think this is ready to go. I've updated it with the bugfix from PR https://github.com/LLNL/lbann/pull/1886.

[Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM375-1).

PR #1849, opened 2021-04-07
Cleanup execution context

benson31:
Severely pare down the API of `execution_context`. Critically examines the API of the base `execution_context` class and eliminates some cruft, notably removing the nonsensical `training_algorithm*` member and externalizing and globalizing the `trainer` object.

Further treatment of this infrastructure is necessary, but this is an initial step in this direction. (Stay tuned for the rest of the steps.)

### DIFF
--- a/src/layers/io/input_layer.cpp
+++ b/src/layers/io/input_layer.cpp
@@ -26,8 +26,11 @@
 
 #define LBANN_INPUT_LAYER_INSTANTIATE
 #include "lbann/layers/io/input_layer.hpp"
-#include "lbann/utils/profiling.hpp"
+
 #include "lbann/callbacks/imcomm.hpp"
+#include "lbann/execution_contexts/execution_context.hpp"
+#include "lbann/execution_contexts/sgd_execution_context.hpp"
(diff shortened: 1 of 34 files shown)


src/layers/io/input_layer.cpp
```diff
@@ -26,8 +26,11 @@
 
 #define LBANN_INPUT_LAYER_INSTANTIATE
 #include "lbann/layers/io/input_layer.hpp"
-#include "lbann/utils/profiling.hpp"
+
 #include "lbann/callbacks/imcomm.hpp"
+#include "lbann/execution_contexts/execution_context.hpp"
```
bvanessen, line 31: Why do we need both includes here?

bvanessen (approved):

src/layers/io/input_layer.cpp
benson31: Strictly speaking, we don't _need_ both. However, from an IWYU perspective, we use the `execution_context` API as well as the more specific `sgd_execution_context` API. There should be no harm to both includes. When we eventually get rid of the `sgd`-specific stuff, we'll only need the `execution_context` include.

PR #1851, opened 2021-04-14
Fix LTFB's SendRecvWeights algorithm for handling different optimizers.

benson31:
This factors out some common code and then implements a version of sendrecv weights that handles the case that a weights object in trainer `i` has a different optimizer than the corresponding object in trainer `j`.


samadejacobs (approved):

timmoon10 (approved):
LGTM

PR #1858, opened 2021-04-23
Added numpy and pandas to the build

bvanessen:


timmoon10:
I'm not a fan of putting these app-specific dependencies in the main build script, especially since building NumPy is somewhat expensive. Using the Clingo concretizer, it is pretty clean to include additional dependencies:
```bash
build_lbann.sh --install-deps --dependencies-only
spack env activate lbann-local-power9le
spack add py-numpy
spack add py-pandas
spack install
```

PR #1861, opened 2021-04-27
Add onnx callback

graham63:
Template for pulling the necessary data from lbann is in place. I could use some help determining what to pass to some (most) of the onnx fields. 

One of the biggest ones is the op_type that I need to fill in for the layers. I think I might need to import and opset, and make sure the op_type for each layer matches the corresponding layer in the opset, but I wasn't able to figure out if there is some kind of "default" we could match. Each layer needs to return the correct value for op_type.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,6 +230,9 @@ endif ()
 set(LBANN_HAS_DIHYDROGEN TRUE)
 message(STATUS "Found DiHydrogen: ${DiHydrogen_DIR}")
 
+# Include onnx dependency
+find_package(ONNX CONFIG REQUIRED)
+
 # Inherit half-precision stuff from Hydrogen
 set(LBANN_HAS_HALF ${HYDROGEN_HAVE_HALF}) # This is CPU-only
 
@@ -657,6 +660,7 @@ target_include_directories(lbann PUBLIC
 target_compile_features(lbann PUBLIC cxx_std_17)
 
 # Use the IMPORTED targets when possible.
+target_link_libraries(lbann PUBLIC onnx)
 target_link_libraries(lbann PUBLIC LbannProto)
 target_link_libraries(lbann PUBLIC Threads::Threads)
 target_link_libraries(lbann PUBLIC clara::clara)
--- a/include/lbann/callbacks/export_onnx.hpp
+++ b/include/lbann/callbacks/export_onnx.hpp
@@ -30,6 +30,9 @@
 #define LBANN_CALLBACKS_EXPORT_ONNX_HPP_INCLUDED
 
 #include "lbann/callbacks/callback.hpp"
+
+#include <google/protobuf/message.h>
+#include <lbann/base.hpp>
 #include <onnx/onnx_pb.h>
 #include <iostream>
 #include <memory>
@@ -47,20 +50,29 @@ class export_onnx : public callback_base {
 
 public:
   /** @brief export_onnx Constructor. */
-  export_onnx();
+  export_onnx(std::shared_ptr<lbann_summary> const& summarizer);
 
+  /** @brief Copy constructor */
+  callback_base* copy() const override {
+    LBANN_ERROR( "This callback is not copyable.");
+    return nullptr;
+  }
 
+  /** @brief Return name of callback */
+  std::string name() const override { return "export_onnx"; }
 
 private:
 
+  /* @brief lbann_summary object */
+  std::shared_ptr<lbann_summary> m_summarizer;
 
 
 }; // class export_onnx
 
 std::unique_ptr<callback_base>
 build_export_onnx_callback_from_pbuf(
-  const google::protobuf::message&,
-  const std::shared_ptr<lbann_summary>&);
+  const google::protobuf::Message& proto_msg,
+  const std::shared_ptr<lbann_summary>& summarizer);
 
 } // namespace callback
 } // namespace lbann
--- a/include/lbann/callbacks/export_onnx.hpp
+++ b/include/lbann/callbacks/export_onnx.hpp
@@ -61,7 +61,16 @@ public:
   /** @brief Return name of callback */
   std::string name() const override { return "export_onnx"; }
 
+  /* @brief gather model info */
+  void on_setup_end(model* m) override;
+
+  /* @brief gather graph/layer info */
+  void on_train_begin(model* m) override;
+
 private:
+
+  /* @brief onnx ModelProto object */
+  onnx::ModelProto mp_;
 
   /* @brief lbann_summary object */
   std::shared_ptr<lbann_summary> m_summarizer;
--- a/include/lbann/callbacks/export_onnx.hpp
+++ b/include/lbann/callbacks/export_onnx.hpp
@@ -1,0 +1,68 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+// may not use this file except in compliance with the License.  You may
+// obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the license.
+//
+// export_onnx .hpp .cpp - Exports trained model to onnx format
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef LBANN_CALLBACKS_EXPORT_ONNX_HPP_INCLUDED
+#define LBANN_CALLBACKS_EXPORT_ONNX_HPP_INCLUDED
+
+#include "lbann/callbacks/callback.hpp"
+#include <onnx/onnx_pb.h>
+#include <iostream>
+#include <memory>
+#include <vector>
+
+using namespace onnx;
+
+namespace lbann {
+namespace callback {
+
+/** @class export_onnx
+ *  @brief Callback to export a trained model to onnx format
+ */
+class export_onnx : public callback_base {
+
+public:
+  /** @brief export_onnx Constructor. */
+  export_onnx();
+
+
+
+private:
+
+
+
+}; // class export_onnx
+
+std::unique_ptr<callback_base>
+build_export_onnx_callback_from_pbuf(
+  const google::protobuf::message&,
+  const std::shared_ptr<lbann_summary>&);
+
+} // namespace callback
+} // namespace lbann
+
+#endif  // LBANN_CALLBACKS_EXPORT_ONNX_HPP_INCLUDED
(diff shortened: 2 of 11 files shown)


include/lbann/layers/transform/split.hpp
```diff
@@ -72,7 +72,23 @@ class split_layer : public data_type_layer<TensorDataType> {
...
 
-
+#ifdef LBANN_HAS_ONNX
+  std::string get_onnx_op_type() const override { return "Identity"; }
+
+  void fill_onnx_node(onnx::GraphProto& graph) const override {
+    for(auto const* child : this->get_child_layers()) {
+      auto* node = graph.add_node();
+      for(auto const* parent : this->get_parent_layers())
+        node->add_input(parent->get_name() + "_0");
```
timmoon10, line 82: We'll run into issues if the parent has multiple children:
```suggestion
    const auto& parent = this->get_parent_layer();
    const size_t idx_in_parent = parent.find_child_layer_index(this);
    for(auto const* child : this->get_child_layers()) {
      auto* node = graph.add_node();
      node->add_input(parent->get_name() + "_" + std::to_string(idx_in_parent));
```

include/lbann/layers/transform/evaluation.hpp
```diff
@@ -119,13 +119,55 @@ class evaluation_layer : public abstract_evaluation_layer<TensorDataType> {
...
+    graph_output_type->mutable_tensor_type()->set_elem_type(1);
+
+    auto* dims = graph_output_type->mutable_tensor_type()->mutable_shape()->add_dim();
+    dims->set_dim_param("batch");
+    dims = graph_output_type->mutable_tensor_type()->mutable_shape()->add_dim();
+    dims->set_dim_value(1);
+    dims = graph_output_type->mutable_tensor_type()->mutable_shape()->add_dim();
+    dims->set_dim_value(28);
+    dims = graph_output_type->mutable_tensor_type()->mutable_shape()->add_dim();
+    dims->set_dim_value(28);
```
timmoon10, line 151: The evaluation layer should output a `batch_size x 1` tensor (e.g. the accuracy for each mini-batch sample):
```suggestion
    auto* dims = graph_output_type->mutable_tensor_type()->mutable_shape()->add_dim();
    dims->set_dim_param("batch");
    dims = graph_output_type->mutable_tensor_type()->mutable_shape()->add_dim();
    dims->set_dim_value(1);
```

include/lbann/layers/transform/evaluation.hpp
```diff
@@ -119,13 +119,55 @@ class evaluation_layer : public abstract_evaluation_layer<TensorDataType> {
...
+
+    //initializer->add_dims(1);
+    //auto const parents = this->get_parent_layers();
+    //for( auto const& dim : parents[0]->get_output_dims() ) {
+      //initializer->add_dims(dim);
+      //dims = graph_output_type->mutable_tensor_type()->mutable_shape()->add_dim();
+      //dims->set_dim_value(dim);
+      //dims->set_denotation("N/A");
+    // }
```
timmoon10, line 160:
```suggestion
```

include/lbann/layers/loss/mean_squared_error.hpp
```diff
@@ -103,6 +103,41 @@ class mean_squared_error_layer : public data_type_layer<TensorDataType> {
...
 
+#ifdef LBANN_HAS_ONNX
+  void fill_onnx_node(onnx::GraphProto& graph) const override {
+    auto* diff = graph.add_node();
+    for(auto const* parent : this->get_parent_layers()) {
+      size_t idx = parent->find_child_layer_index(*this);
+      diff->add_input(parent->get_name() + "_" + std::to_string(idx));
+    }
+    diff->add_output("diff_0");
+    diff->set_name("diff");
```
timmoon10, line 114: We should make sure that all ONNX nodes and tensors have unique names if the model has multiple MSE layers:
```suggestion
    diff->add_output(this->get_name() + "_diff_0");
    diff->set_name(this->get_name() + "_diff");
```
Similar changes should be made to the rest of the function.

timmoon10 (changes requested):
I see some edge cases. I also have some design suggestions for the next stage of this work, mostly imitating the "dump weights" or "dump outputs" callbacks.

Once the issues are addressed, can you run it with Bamboo?

CMakeLists.txt
```diff
@@ -473,20 +473,20 @@ endif (LBANN_WITH_HWLOC)
...
+#  if (NOT HDF5_FOUND)
+#    find_package(HDF5 CONFIG QUIET)
+#  endif ()
+#  if (NOT HDF5_FOUND)
+#    enable_language(C) # WHY??????????????
+#    find_package(HDF5 REQUIRED)
+#    set(HDF5_FOUND_WITH_MODULE TRUE)
+#  else ()
+#    message(STATUS "Found HDF5: ${HDF5_DIR}")
+#  endif ()
```
timmoon10, line 489:
```suggestion
  find_package(HDF5 CONFIG QUIET
    HINTS ${HDF5_DIR} $ENV{HDF5_DIR}
    PATH_SUFFIXES share/cmake/hdf5
    NO_DEFAULT_PATH)
  if (NOT HDF5_FOUND)
    find_package(HDF5 CONFIG QUIET)
  endif ()
  if (NOT HDF5_FOUND)
    enable_language(C) # WHY??????????????
    find_package(HDF5 REQUIRED)
    set(HDF5_FOUND_WITH_MODULE TRUE)
  else ()
    message(STATUS "Found HDF5: ${HDF5_DIR}")
  endif ()
```

timmoon10 (changes requested):
Looks like some debugging code slipped in, but otherwise looks good. Have you been able to run this with Bamboo?

timmoon10 (approved):
If this passes Bamboo, it's good to go.

I anticipate there will be conflicts with PR #1955. I think we should merge this first, with the understanding that the operator refactor will disrupt support for some layers. The important thing is to get the base layer class infrastructure. What do you think @benson31?

src/callbacks/unit_test/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 set_full_path(THIS_DIR_MPI_CATCH2_TEST_FILES
   print_statistics_test.cpp
+  export_onnx_test.cpp
```
benson31, line 3: Either add this file or 
```suggestion
```

include/lbann/layers/loss/mean_squared_error.hpp
```diff
@@ -103,6 +103,41 @@ class mean_squared_error_layer : public data_type_layer<TensorDataType> {
   data_layout get_data_layout() const override { return T_layout; }
   El::Device get_device_allocation() const override { return Dev; }
 
+#ifdef LBANN_HAS_ONNX
```
benson31, line 106: This class is a prime example of why `get_onnx_op_type` should not be public (see my general comment, point 2). This class does not have an override for that function (and looking at the impl of `fill_onnx_node`, it seems there's not a single right answer for what to return in that case). So there's inconsistent behavior; a user calling `get_onnx_op_type()`, which should be valid for this onnx-ified class, will get an exception, but `fill_onnx_node` will be fine.

include/lbann/callbacks/export_onnx.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+#ifdef LBANN_HAS_ONNX
+#include <onnx/onnx_pb.h>
+#endif // LBANN_HAS_ONNX
+#include "lbann/callbacks/callback.hpp"
+#include <google/protobuf/message.h>
+#include <lbann/base.hpp>
+#include <iostream>
+#include <memory>
+#include <vector>
```
benson31, line 40:
```suggestion
#include "lbann/callbacks/callback.hpp"
#include "lbann/base.hpp"

#ifdef LBANN_HAS_ONNX
#include <onnx/onnx_pb.h>
#endif // LBANN_HAS_ONNX
#include <google/protobuf/message.h>

#include <iostream>
#include <memory>
#include <vector>
```

include/lbann/callbacks/export_onnx.hpp
```diff
@@ -0,0 +1,97 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
```
benson31, line 2:
```suggestion
// Copyright (c) 2014-2021, Lawrence Livermore National Security, LLC.
```

src/callbacks/export_onnx.cpp
```diff
@@ -0,0 +1,107 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
```
benson31, line 2:
```suggestion
// Copyright (c) 2014-2021, Lawrence Livermore National Security, LLC.
```

src/callbacks/export_onnx.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "lbann/layers/io/input_layer.hpp"
+
+#include "lbann/proto/helpers.hpp"
+#include "lbann/utils/factory.hpp"
+#include "lbann/utils/summary_impl.hpp"
+
+#include <callbacks.pb.h>
+
+#include <string>
+
```
benson31, line 42:
```suggestion
#include "lbann/callbacks/export_onnx.hpp"

#include "lbann/layers/io/input_layer.hpp"
#include "lbann/proto/helpers.hpp"
#include "lbann/utils/factory.hpp"
#include "lbann/utils/summary_impl.hpp"

#include <callbacks.pb.h>

#include <fstream>
#include <iostream>
#include <string>
```

include/lbann/callbacks/export_onnx.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+public:
+  /** @brief export_onnx Constructor.
+   *  @param output_file Output filename (default = lbann_output.onnx)
+   *  @param print_debug_string Option to print debug string to stdout
+   */
+  export_onnx(bool print_debug_string = false,
+              std::string output_file = "lbann_output.onnx");
+
+  /** @brief Copy constructor */
```
benson31, line 58:
```suggestion
  /** @brief Copy interface */
```
But also, why is it not copyable? Protobuf messages are copyable. (FWIW, I don't think it makes sense to copy a callback ever, but somebody thought it made sense to add the `copy()` function. Since this object actually is copyable, and makes as much sense to copy as any other callback, we should support this function as well as any other callback...)

benson31 (commented):
A few comments:
1. I'm concerned about having a conditional API in the layers (that is, `fill_onnx_node(...)` of the `Layer` API protected by a preprocessing macro). A possible workaround is to define LBANN objects that wrap the ONNX objects but are trivial structs when ONNX is not supported.
2. After reading the PR more thoroughly and thinking about it, I don't see the value of `get_onnx_op_type()` being public. It's only used virtually as an implementation detail of the default implementation of `fill_onnx_node`. Moreover, someone adding onnx support for a new layer will either be OK with the default implementation of `fill_onnx_node`, in which case they need only override `get_onnx_op_type`, OR they'll need to write a new implementation of `fill_onnx_node` and it will be less typing to just hard-code the op name in the `set_op_type` call rather than overriding `get_onnx_op_type` and then calling `this->get_onnx_op_type();`. As shown by MSE, there are cases for which `get_onnx_op_type` may not have a reasonable return value, but `fill_onnx_node` will be fine. I recommend making it `private` or `protected` virtual, with the default implementation. Then provide documentation that `get_onnx_op_type` doesn't need to be overridden if `fill_onnx_node` is overridden.
3. There's no documentation around what the default implementation of `fill_onnx_node` will provide, making it difficult for a new developer to understand when they should add their own override or just use the default.

benson31 (changes requested):
Meant to click "Request changes" before but just clicked "Comment", apparently... Oops.

benson31:
I rescind my concern labeled "1." in my above comment. Upon further consideration, it would probably be messier than it's worth to hide this.

src/callbacks/unit_test/export_onnx_test.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <onnx/onnx_pb.h>
+
+#include <iostream>
+#include <memory>
+
+
+using unit_test::utilities::IsValidPtr;
+TEST_CASE("Serializing \"export onnx\" callback",
+          "[mpi][callback][serialize]")
```
benson31, line 46:
```suggestion
          "[mpi][callback][serialize][onnx]")
```

benson31 (changes requested):
We should move the implementations of "fill_onnx_node" out of the header files.

bvanessen (approved):
Overall looks good. As noted by @benson31 we should move the implementation of the fill_onnx_node function to the cpp file.

PR #1864, opened 2021-05-05
Edge Conditioned Graph Neural Networks 

szaman19:
- Added edge conditioned graph neural networks to LBANN
    - Added Scatter-Gather and Matrix Multiplication based implementation of MPNN   
- Example application on OGB Large Scale Competition (LSC) Quantum Chemistry dataset
- Added Synthetic benchmarking datasets

### DIFF
--- a/applications/graph/GNN/NNConvModel.py
+++ b/applications/graph/GNN/NNConvModel.py
@@ -1,0 +1,345 @@
+import lbann
+import math
+from lbann.util import str_list
+from lbann.modules.graph import NNConv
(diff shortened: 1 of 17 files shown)


timmoon10 (approved):
LGTM. My only concern is that the integration test is not included in the continuous integration. We've already talked about how slow the test is and how it is only on Ray right now. Instead of a full integration test, I wonder if we could just have a minimal unit test that makes sure that `NNConv` outputs the same results as a Python implementation in NumPy or some graph package.

bvanessen (approved):

PR #1867, opened 2021-05-06
LTFB support for multiple metrics

benson31:
This adds the multi-metric support feature to RandomPairwiseExchange. All metrics must exist in every model. The local model is favored; a partner model must win ALL metrics to be declared the winner.

The front-ends use associative maps (dictionaries in python, maps in protobuf). See the LTFB test file for a usage example.

### DIFF
--- a/bamboo/unit_tests/test_unit_algo_ltfb.py
+++ b/bamboo/unit_tests/test_unit_algo_ltfb.py
@@ -65,8 +65,7 @@ def setup_experiment(lbann):
     RPE = lbann.RandomPairwiseExchange
     SGD = lbann.BatchedIterativeOptimizer
     metalearning = RPE(
-        metric_name='random',
-        metric_strategy=RPE.MetricStrategy.HIGHER_IS_BETTER)
+        metric_strategies={'random': RPE.MetricStrategy.HIGHER_IS_BETTER})
     ltfb = lbann.LTFB("ltfb",
                       metalearning=metalearning,
                       local_algo=SGD("local sgd",
--- a/src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
+++ b/src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
@@ -44,6 +44,7 @@
 #include <sstream>
 #include <string>
 #include <unordered_map>
+#include <vector>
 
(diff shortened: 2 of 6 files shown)


src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
```diff
@@ -245,11 +314,11 @@ void RandomPairwiseExchange::select_next(model& m,
                            " (trainer ",
                            local_trainer,
                            " score = ",
-                           local_score,
+                           local_scores.begin()->second,
```
samadejacobs, line 317: Might be helpful (informative) to print score for each metrics and not just the first metric.

bamboo/unit_tests/test_unit_algo_ltfb.py
```diff
@@ -65,8 +65,7 @@ def setup_experiment(lbann):
     RPE = lbann.RandomPairwiseExchange
     SGD = lbann.BatchedIterativeOptimizer
     metalearning = RPE(
-        metric_name='random',
-        metric_strategy=RPE.MetricStrategy.HIGHER_IS_BETTER)
+        metric_strategies={'random': RPE.MetricStrategy.HIGHER_IS_BETTER})
```
samadejacobs, line 68: Can we test multiple metrics?

src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
benson31: Yeah, we don't have a canoncal `operator<<(std::ostream&, std::vector<T> const&)` overload. This was just for simplicity. I'll expand it.

bamboo/unit_tests/test_unit_algo_ltfb.py
benson31: We can. I'm working on some Catch tests. In the meantime, I will add a new Bamboo test for this feature. I don't want to modify this one, since this is supposed to mimic `test_unit_callback_ltfb.py`, at least until that callback has been removed.

timmoon10 (approved):
:shipit:

bvanessen (approved):

PR #1868, opened 2021-05-11
Update smiles test

bvanessen:


benson31 (approved):
Haven't run it, but the code seems good.

PR #1877, opened 2021-06-01
Change CUDA arch handling in LBANN

benson31:
This uses a different approach for dealing with the FP16 requirements on CUDA platforms.


timmoon10 (approved):
:shipit: 

This should help reduce the long startup time on Lassen from JIT compilation.

bvanessen (approved):

PR #1886, opened 2021-06-04
Bug fix for K-FAC callback with ROCm

mrwyattii:
This fixes a bug introduced in #1634 that causes ROCm builds to fail.

- Swap out CUDA kernel calls for `hydrogen::gpu::LaunchKernel`
- Use `El::SyncInfo` in place of `cudaStream_t`
- Updated CMakeLists to fix undefined reference with ROCm build


timmoon10 (approved):
:shipit:

benson31 (approved):
:shipit:

PR #1888, opened 2021-06-07
Fix a C++17 compilation issue with SMILES data reader

benson31:
`std::random_shuffle` is deprecated as of C++14 and removed entirely in C++17. This should be a nearly-equivalent-if-not-slightly-better fix.

### DIFF
--- a/src/data_readers/data_reader_smiles.cpp
+++ b/src/data_readers/data_reader_smiles.cpp
@@ -39,6 +39,7 @@
 #include <algorithm>
 #include <mutex>
 #include <cctype>
+#include <random>
 
 namespace lbann {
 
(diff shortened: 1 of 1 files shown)


davidHysom (approved):
LGTM; HOWEVER the modified "shuffle" line of code was put in by Brian, so I may not be best qualified to judge.

benson31:
@davidHysom Fair enough. Thanks for taking a look. I usually mark you as a review for anything matching `data_reader.*` 😄

benson31:
Some additional details: I thought about using one of LBANN's random number generators for this, but I wasn't sure which was the appropriate one, or if doing so would mess with some other subtle, undocumented assumption about their use/state. Also, the old implementations of `std::random_shuffle` would likely have used `std::rand()` under the hood, so spinning up an on-the-fly RNG seemed like the closest approximation to that behavior. I make no comment on whether that's "good" or "correct" behavior; my task here was to fix the compiler error with the smallest diff.

bvanessen (approved):

PR #1889, opened 2021-06-07
Fix some initializers

benson31:
IDK if this is worth merging; it seems like it's fixed differently in the HDF5 data reader updates (#1688).


bvanessen (approved):

bvanessen:
@benson31 I am going to merge this now to fix current bamboo problems.

PR #1891, opened 2021-06-08
Refactor input options to use argument parser class

mrwyattii:
Refactor of LBANN input options to use the `argument_parser` class.  Input options are enumerated in `options.cpp` file, split into different functions (e.g., `std`, `datareader`, and `jag`).  Trying to keep this as a pure refactor, but some notable changes:

- Many options now have invalid default options (e.g., `""`, `-1`) that we use to check if an option was provided by the user.  In some cases, having reasonable default values might give a cleaner implementation
- Some option checking logic is not preserved.  For example, if a user provides an invalid value that happens to be the same as the default invalid value, an incorrect error message will be produced indicating the option was not provided
- There existed a few cases where options values were added within the datareader implementation.  These are now commented out (e.g., `node_sizes_vary` now needs to be passed as an option when using `image_data_reader`)
- Moved some options to more reasonable boolean flags (e.g., `--disable_cuda=1` is now just `--disable_cuda`)
- Modified some Bamboo tests to adapt to the changes described above ^

Some of these issues will need to be addressed in future PRs to provide overall cleaner argument parsing.  Future work is marked with `TODO MRW`


bvanessen (approved):

tests/test_shuffled_indices.cpp
```diff
@@ -50,23 +50,29 @@ int main(int argc, char *argv[]) {
 
   try {
     // Initialize options db (this parses the command line)
-    options *opts = options::get();
-    opts->init(argc, argv);
-    if (opts->has_string("h") or opts->has_string("help") or argc == 1) {
-      print_help(*comm);
+    auto& arg_parser = global_argument_parser();
+    construct_all_options();
+		arg_parser.add_flag(FN,
```
benson31, line 55:
```suggestion
		arg_parser.add_flag("fn",
```

tests/test_shuffled_indices.cpp
```diff
@@ -50,23 +50,29 @@ int main(int argc, char *argv[]) {
...
+
+    if (arg_parser.help_requested() or argc == 1) {
+      if (master)
+        std::cout << arg_parser << std::endl;
       return EXIT_SUCCESS;
     }
 
     //read data_reader prototext file
-    if (not opts->has_string("fn")) {
+    if (arg_parser.get<std::string>(FN) == "") {
```
benson31, line 67:
```suggestion
    if (arg_parser.get<std::string>("fn") == "") {
```

tests/test_shuffled_indices.cpp
```diff
@@ -50,23 +50,29 @@ int main(int argc, char *argv[]) {
...
       std::cerr << __FILE__ << " " << __LINE__ << " :: "
                 << "you must run with: --fn=<string> where <string> is\n"
                 << "a data_reader prototext filePathName\n";
       return EXIT_FAILURE;
     }
 
     lbann_data::LbannPB pb;
-    std::string reader_fn(opts->get_string("fn").c_str());
+    std::string reader_fn(arg_parser.get<std::string>(FN).c_str());
```
benson31, line 75:
```suggestion
    std::string reader_fn = arg_parser.get<std::string>("fn");
```

include/lbann/utils/options.hpp
```diff
@@ -1,111 +1,148 @@
...
-  int m_argc;
-  char **m_argv;
-
-  std::vector<void*> m_ptrs;
-};
+/****** std options ******/
+// Bool flags
+#define DISABLE_BACKGROUND_IO_ACTIVITY "disable_background_io_activity"
+#define DISABLE_CUDA "disable_cuda"
+#define FN "fn"
```
benson31, line 17:
```suggestion
```

benson31 (approved):

tests/test_shuffled_indices.cpp
bvanessen: @benson31 Is there a reason that this one doesn't use the FN macro, where the other tests use it?

PR #1895, opened 2021-06-10
Operator class

bvanessen:
Added an operator class that will provide a simpler approach to
defining the mathematical operation that is executed by a layer.  The
goal is that layers would execute operators that are assigned to them
and that operators act as curried functions (based on parameters
during construction).  Operators should be functional, taking inputs
and producing outputs in the forward direction, and taking error
signalsand inputs to produce new error signals in the backward
direction.

Implemented a data_type and element-wise operator sub-classes that are
specialized on Input/Output data types and tuned for operators that
operate on purely local matrices, respectively.

Rewrote the clamp layer as an operator.


include/lbann/layers/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 # Add the headers for this directory
 set_full_path(THIS_DIR_HEADERS
-  layer.hpp
   data_type_layer.hpp
+  doer.hpp
```
bvanessen, line 4: Isn't this replaced with operator_layer?

benson31 (approved):
Passes Bamboo; LGTM

src/layers/math/unit_test/clamp_test.cpp
```diff
@@ -24,6 +24,8 @@
 // permissions and limitations under the license.
 ////////////////////////////////////////////////////////////////////////////////
 
+#if 0
```
bvanessen, line 27: @benson31 Is this test supposed to be commented out?

src/models/model.cpp
```diff
@@ -2531,6 +2533,25 @@ void model::print_distributions() const {
 }
 #endif // LBANN_HAS_DISTCONV
 
+void foofoobar(El::AbstractMatrix<float>& x) {
```
bvanessen, line 2536: @benson31 Is this function really used?

timmoon10 (approved):
LGTM

include/lbann/operators/operator.hpp
```diff
@@ -0,0 +1,209 @@
...
+ *  defined during construction but do not hold internal state.  A
+ *  operator should also be able to take objective function gradients
+ *  w.r.t. the outputs ("previous error signals") and compute the
+ *  objective function gradients w.r.t. the inputs ("error
+ *  signals"). This allows the model to perform automatic
+ *  differentiation.
+ *
+ *  Operators are specified for unique input and output data types.
+ */
+template <typename InputT, typename OutputT, El::Device D>
```
timmoon10, line 89: We are making an assumption that all inputs have the same type and all outputs will have the same type. This is reasonable for most of our operations and it's best not to prematurely over-optimize, but we should anticipate cases where this may not hold:
- Indexing operations like gather or scatter, where one input is floating point and the other is integer.
- Operations where inputs are on different devices, e.g. a rotation layer where the image is on GPU and the rotation angle is on CPU.
- Operations without inputs (e.g. random distributions) or without outputs (e.g. outputs to objective functions or metrics).

include/lbann/layers/operator_layer_impl.hpp
```diff
@@ -0,0 +1,265 @@
...
+  OperatorLayer const& other)
+  : DataTypeLayer(other), m_ops{clone_ops(other.m_ops)}
+{}
+
+template <typename InputT, typename OutputT, data_layout Layout, El::Device D>
+auto OperatorLayer<InputT, OutputT, Layout, D>::operator=(
+  OperatorLayer const& other) -> OperatorLayer&
+{
+  // This is self-assignment safe
+  data_type_layer<InputT, OutputT>::operator=(*this);
```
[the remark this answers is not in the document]
benson31:
```suggestion
  data_type_layer<InputT, OutputT>::operator=(other);
```

include/lbann/operators/operator.hpp
bvanessen: Yes for now we are making that assumption.  We will address these more complex operations in future designs. Possibly as manual layer, or a more complex operator interface.

PR #1897, opened 2021-06-18
Refactor K-FAC callback as training algorithm

timmoon10:
This PR implements a K-FAC training algorithm based on the SGD training algorithm. At the end of each forward prop step and each backprop step, it applies the K-FAC implementation from the K-FAC callback (see PRs #1634 and #1845). To keep the training algorithm functional, I've separated out the state into an execution context.

Note that the implementation itself hasn't changed. It will require further refactoring to split up Kronecker factor computation/inversion (infrequent) and gradient preconditioning (ideally every step).

Pinging @aj-prime.


timmoon10:
[There are no new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM376-5). I think this is ready to go.

PR #1899, opened 2021-06-28
Add functionality to insert, remove and replace layers in the model

soumyadipghosh:
- Adds insert, remove and replace functions to modify layers in the model
- Adds corresponding unit tests

NOTE: There is an assumption that the layer being modified has just one parent and one child.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -971,6 +971,151 @@ void model::add_split_layers(std::unordered_set<std::string>& layer_names) {
   }
 }
 
+// EXPERIMENTAL: Insert layer in model
+// Arguments - ptr to new layer, name of preceding layer where to insert
+void model::insert(OwningLayerPtr&& new_layer, std::string preceding_layer_name) {
+
+  // Find preceding layer after which to insert new layer
+  int preceding_layer_index = -1;
+  for (int i=0; i<this->get_num_layers(); ++i) {
+    if (this->get_layer(i).get_name() == old_layer_name) {
+      preceding_layer_index = i;
+      break;
+    }
+  }
+  if (preceding_layer_index < 0) {
+    LBANN_ERROR(
+      "Attempted to insert layer after ",
+      "layer \"",preceding_layer_name,"\",",
+      "but no such layer exists");
+  }
+
+  auto& l = get_layer(preceding_layer_index);
+
+  // Set checks to ensure there is only one parent and one child
+  if (l.get_num_parents() != 1) {
+    LBANN_ERROR(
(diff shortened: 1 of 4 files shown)


timmoon10:
[Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM378-1).

samadejacobs (approved):
LGTM

src/models/model.cpp
```diff
@@ -971,6 +971,129 @@ void model::add_split_layers(std::unordered_set<std::string>& layer_names) {
...
+  // Other checks also have to be done - like cannot delete activation function between two layers, etc
+
+  // Parent and child of removable layer
+  auto& parent = const_cast<Layer&>(l.get_parent_layer(0)); // assuming only one parent
+  auto& child = const_cast<Layer&>(l.get_child_layer(0)); // assuming only one child
+
+  // Setup relationship between parent layer and child layer
+  child.replace_parent_layer(l.get_parent_layer_pointer(0), child.find_parent_layer_index(l));
+  parent.clear_child_layers();
+  parent.add_child_layer(l.get_child_layer_pointer(0));
```
timmoon10, line 1046: We should handle the case where `parent` has multiple children:
```suggestion
  parent.replace_child_layer(l.get_child_layer_pointer(0), parent.find_child_layer_index(l));
```

src/models/model.cpp
```diff
@@ -971,6 +971,129 @@ void model::add_split_layers(std::unordered_set<std::string>& layer_names) {
...
+  auto& parent = const_cast<Layer&>(l.get_parent_layer(0)); // assuming only one parent
+  auto& child = const_cast<Layer&>(l.get_child_layer(0)); // assuming only one child
+  
+  // Setup relationship between the new layer and child of old layer (which becomes child of new layer)
+  new_layer->add_child_layer(l.get_child_layer_pointer(0));
+  child.replace_parent_layer(new_layer, child.find_parent_layer_index(l));
+
+  // Setup relationship between parent of old layer (which becomes parent of new layer) and new layer
+  parent.clear_child_layers();
+  parent.add_child_layer(new_layer);
```
timmoon10, line 1087: We should handle the case where `parent` has multiple children:
```suggestion
  parent.replace_child_layer(new_layer, parent.find_child_layer_index(l));
```

timmoon10 (changes requested):
Mostly looks good. Just an edge case if a parent layer has multiple children.

benson31 (approved):
Once @timmoon10 is happy, I'm ok with the rest of this.

timmoon10 (approved):

PR #1903, opened 2021-06-30
Restore global batch/epoch count for validation

benson31:
@samadejacobs Please verify that this restores the functionality that you want.

~Note that the validation count is global *within the `sgd_training_algorithm::train()` call*. Thus, when doing LTFB, for example, it will reset each time local training is called.~

Now, the validation epoch count is global across an LBANN run.

### DIFF
--- a/src/execution_algorithms/sgd_training_algorithm.cpp
+++ b/src/execution_algorithms/sgd_training_algorithm.cpp
@@ -83,6 +83,13 @@ void sgd_training_algorithm::train(sgd_execution_context& c,
                                    data_coordinator& dc,
                                    sgd_termination_criteria const& term)
 {
+  // Setup a "training-global" validation context:
+  using ValidationContext = sgd_execution_context;
+  ValidationContext evaluation_context(
+    execution_mode::validation,
+    dc.get_mini_batch_size(execution_mode::validation));
+  size_t num_validation_epochs = 1UL;
+
   // Initialize some state so it knows we're training now.
   c.set_execution_mode(execution_mode::training);
   model.reset_mode(c, execution_mode::training);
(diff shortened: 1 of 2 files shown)


timmoon10 (approved):
LGTM

benson31:
Comment from @samadejacobs:

> this would not work for LTFB, dump outputs callback for example will override output file at every LTFB super step.

I have updated the PR to make the validation context globally persistent across all training runs in a single instance of LBANN.

bvanessen (approved):

src/execution_algorithms/sgd_training_algorithm.cpp
```diff
@@ -83,6 +84,13 @@ void sgd_training_algorithm::train(sgd_execution_context& c,
                                    data_coordinator& dc,
                                    sgd_termination_criteria const& term)
 {
+  // Setup a "training-global" validation context:
+  using ValidationContext = sgd_execution_context;
+  static ValidationContext evaluation_context(
```
timmoon10, line 89: Won't making this `static` prevent us from juggling multiple instances of `sgd_training_algorithm`?

timmoon10 (changes requested):

src/execution_algorithms/sgd_training_algorithm.cpp
benson31: Why yes it will! But we don't have those. What we do have is legacy code in callbacks. It's easier to change these two lines than to try to fix callbacks. 🤷‍♂️

benson31:
@timmoon10 I made the validation context member data. It will now monotonically increase throughout the lifetime of an instance of `sgd_training_algorithm`. For LTFB, we keep the same local training algorithm instance, so @samadejacobs should get the behavior he wants and you should be able to juggle multiple SGDs.

timmoon10 (approved):

samadejacobs (approved):
LGN, please merge.

PR #1907, opened 2021-07-01
Fix a compilation error in gru layer

benson31:
The differences between CPU and GPU are probably significant enough that partial specialization on device would be a better approach here, but that's a more complicated refactoring.


timmoon10 (approved):
:shipit:

PR #1908, opened 2021-07-06
Adding implementation for callback to perturb weights.

tnat410:
Adding perturb_weights callback.

Note: There is the assumption that weight layer is used for crop operation, i.e. its dimension is 3.
Perturbation values are range from 0.3 to 0.7 .

### DIFF
--- a/include/lbann/callbacks/perturb_weights.hpp
+++ b/include/lbann/callbacks/perturb_weights.hpp
@@ -1,0 +1,90 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+// may not use this file except in compliance with the License.  You may
+// obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the license.
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef LBANN_CALLBACKS_CALLBACK_PERTURB_WEIGHT_HPP_INCLUDED
+#define LBANN_CALLBACKS_CALLBACK_PERTURB_WEIGHT_HPP_INCLUDED
+
+#include "lbann/callbacks/callback.hpp"
+#include "lbann/weights/weights.hpp"
+
+
+namespace lbann {
+namespace callback {
+
+/** @brief Perturb weights for a specific layer in model.
+ *
+ *  
+ */
+class perturb_weights : public callback_base {
+public:
+
+  /** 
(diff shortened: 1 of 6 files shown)


include/lbann/callbacks/perturb_weights.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+   /*
+   *  output_name.
+   */
+  std::string m_output_name;
+  
+  weights* m_output;
+  
+  /** Compute model size. */
+  void perturbed(model& m);
```
timmoon10, line 78: This function name is a bit clearer to me. `perturbed` makes me think that this function is checking whether the model has been previously perturbed.
```suggestion
  void perturb(model& m);

```

src/proto/callbacks.proto
```diff
@@ -81,6 +81,8 @@ message Callback {
     CallbackDumpModelGraph dump_model_graph = 49;
     CallbackPerturbLearningRate perturb_learning_rate = 50;
     CallbackComputeModelSize compute_model_size = 51;
+    CallbackKFAC kfac = 52;
```
timmoon10, line 84: Looks like there was some git mixup. This callback has been deleted.
```suggestion
```

src/proto/callbacks.proto
```diff
@@ -408,4 +410,43 @@ message Callback {
...
+    string disable_layers = 14; // List of layers to be ignored by the callback
+
+    // Factor to be multiplied to the learning rate (default: 1)
+    float learning_rate_factor = 15;
+    // Factor to be multiplied to the learning rate for GRU layers
+    // (default: learning_rate_factor)
+    float learning_rate_factor_gru = 16;
+  }
+
+
```
timmoon10, line 447:
```suggestion
```

src/callbacks/perturb_weights.cpp
```diff
@@ -0,0 +1,166 @@
...
+  const auto thres = 0.5;
+  const auto lower = 0.3;
+  const auto upper = 0.7;
+  
+  // RNG
+  auto& gen = get_generator();
+  std::normal_distribution<DataType> dist(zero, one);
+
+
+  for (auto* w : m.get_weights()) {
```
timmoon10, line 97: I don't think we need to iterate through the weights since we've already found `m_output` in `setup`. Alternatively, we could remove `m_output` as a member variable and search for the weights every time this callback is activated. I prefer this approach since it makes the callback stateless.

src/callbacks/perturb_weights.cpp
```diff
@@ -0,0 +1,166 @@
...
+			auto perturbed_val = val;
+			
+			if(val >= lower && val <= upper){	
+				if (dist(gen) > thres){
+					perturbed_val = val * 1.1;
+				}
+				else {
+					perturbed_val = val * 0.9;
+				}
+			}
```
timmoon10, line 127: If I am understanding this correctly, you want a 50% chance of either increasing the weights by 1.1x or decreasing by 0.9x? In that case, the distribution should be uniform instead of normal.

This works, but it forces the weights to take discrete values. If we want continuous values, we might try something like:
```suggestion
			perturbed_val += dist(gen); // dist is a std::normal_distribution
			perturbed_val = std::min(std::max(perturbed_val, lower), upper);
```

src/callbacks/perturb_weights.cpp
```diff
@@ -0,0 +1,166 @@
...
+	auto& values = w->get_values();
+	auto& new_values = dynamic_cast<El::AbstractDistMatrix<DataType>&>(values);
+
+	auto& local_values = new_values.Matrix();
+	El::Matrix<DataType,El::Device::CPU> temp;
+	El::Copy(local_values, temp);
+
+	// Perturb weights on master process		
+	if (comm->am_trainer_master()) {
+		for (auto i = 0; i < range; i++){
```
timmoon10, line 114: We can accommodate arbitrarily-sized weights with:
```suggestion
		for (auto i = 0; i < temp.Height(); i++){		

```

timmoon10 (changes requested):

src/callbacks/perturb_weights.cpp
tnat410: Updated.

timmoon10 (approved):
LGTM. If we don't get any new errors in [Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM383-1), we can merge.

PR #1913, opened 2021-07-08
Fix compile error in channel-wise mean layer

timmoon10:


include/lbann/layers/misc/channelwise_mean.hpp
```diff
@@ -41,7 +41,7 @@ class channelwise_mean_layer : public data_type_layer<TensorDataType> {
                 "data-parallel data layout");
 public:
 
-  channelwise_mean_layer()
+  channelwise_mean_layer(lbann_comm* comm=nullptr)
```
benson31, line 44:
```suggestion
  channelwise_mean_layer(lbann_comm* = nullptr)
```

benson31 (approved):

PR #1879, opened 2021-06-02
Subgraph Parallelism support for Transformer models

aj-prime:
This PR exposes subgraph parallelism as a new strategy to parallelize the DNN training by processing different branches on separate GPUs.

It implements Data Sub-Graph Parallelism (D&SP) and D&SP-cSub designs. 

Initial support is available to extend sub-graph parallelism for common layers in Transformer models to train out-of-core transformer models. 

An example is available in "applications/nlp/transformer/subgraph" directory. 

Note: It does not support multiple trainers and needs "translatebetweengrid" functions in Hydrogen.

PR #1914, opened 2021-07-08
Align layer inputs when needed

timmoon10:
Prior to PR #1879, we made sure that all layer input data was aligned. This was removed in PR #1879 to handle the case where a layer has parents in different sub-grids, but it introduced errors in the case without sub-grid parallelism. This PR restores the old behavior when sub-grid parallelism isn't enabled.

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM381-1).


benson31 (approved):
LGTM

PR #1915, opened 2021-07-09
Skip the DFT test on Corona

benson31:
Since we don't have rocFFT support in LBANN, we need to skip this test on Corona. For now, this is accomplished at the cluster level (since we don't really check for enabled features in Python), which is not the most graceful approach. But hey, it works.

### DIFF
--- a/bamboo/common_python/tools.py
+++ b/bamboo/common_python/tools.py
@@ -699,6 +699,13 @@ def create_tests(setup_func,
                                                'experiments',
                                                test_name)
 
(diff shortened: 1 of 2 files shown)


timmoon10 (changes requested):
It would be easier to add `skip_clusters=[]` as a kwarg to `create_tests`. That way we don't need to manipulate `_kwargs` directly.

timmoon10 (approved):
LGTM

PR #1917, opened 2021-07-13
Make layer copy/move constructors protected in the base classes

benson31:
Should be a no-op change, just for consistency.


timmoon10 (approved):
If it passes the tests, :shipit:

PR #1918, opened 2021-07-13
Make data_type_layer safe for self-assignment

benson31:
I got too excited about `protected`-izing I forgot to do the thing I actually wanted to do...


timmoon10 (approved):
:shipit:

PR #1898, opened 2021-06-25
2D Scatter-Gather

szaman19:
- Adds support for 2-dimensional inputs for the Values into the Scatter and Gather layer
- Updated unit tests to test the new functionality

PR #1921, opened 2021-07-15
Fix invalid comparison in scatter layer

timmoon10:
Building with `-Werror` exposed an incorrect comparison introduced in PR #1898.

Pinging @szaman19.


benson31 (approved):
LGTM

PR #1922, opened 2021-07-15
Remove redundant check in "perturb weights" callback

timmoon10:
Building with `-Werror`, GCC complained about a comparison that always returns `true`: https://github.com/LLNL/lbann/blob/ecad0fe7ac9ae84cfd4ac834fc54bb08cd4938ce/src/callbacks/perturb_weights.cpp#L86
We can either remove it or make it skip the first step.

Pinging @tnat410.


samadejacobs (approved):
LGTM

PR #1924, opened 2021-07-17
Basic sub-grid parallelism for FC and Conv

aj-prime:
This PR introduces the concept of sub-grid parallelism (primary and secondary grid) for 2nd order optimization (KFAC). KFAC code is refactored to enable the use of sub-grid parallelism.  

**Features**
1. Splits trainer comm into the primary and secondary grid
2. Use the secondary grid to compute KFAC 
3. Ranks in primary gird own the tensors in the model and ranks in the secondary grid are observers only 
4. Model can be initialized on both grids to eliminate the need of transferring errors and activations from primary to secondary grid

**New Flags for Trainer**
1. LBANN_TRAINER_PRIMARY_GRID_SIZE
2. LBANN_TRAINER_CREATE_TWO_MODELS

**Minor modifications**
1. Added GRU model for basic testing in kfac.py
2. Refactored KFAC functions
3. Allgather after the computation of inverse matrices. 
4. Conditioned gradients are calculated on each rank (data-parallel instead of model-parallel)


src/execution_algorithms/kfac.cpp
```diff
@@ -377,6 +435,331 @@ void KFAC::do_batch_end_cbs(model& model)
...
+
+  const auto layers = model.get_layers();
+  const El::mpi::Comm & combined_comm = comm->get_combined_grid_comm();
+  const El::mpi::Comm & trainer_comm = comm->get_KFAC_comm();
+  const int comm_size = El::mpi::Size(comm->get_KFAC_comm());
+  const int comm_rank = El::mpi::Rank(comm->get_KFAC_comm());
+}
+
+
+void KFAC::sync_weights_model(model& model, lbann_comm *comm){
```
timmoon10, line 453: The implementation here works, although it would be easier to iterate through the list of weights rather than to iterate through the list of layers. It would also avoid redundant communication if multiple layers share the same weights.

src/execution_algorithms/kfac.cpp
```diff
@@ -377,6 +435,331 @@ void KFAC::do_batch_end_cbs(model& model)
...
+       global_buffer_size,
+       primary_grid_ranks[recv_index],
+       combined_comm,
+       sync_info);
+  }
+
+
+
+
+  //copy weights to global buffers
```
timmoon10, line 565:
```suggestion
  //copy weights from global buffers
```

src/execution_algorithms/kfac/kfac_util.cpp
```diff
@@ -430,6 +430,66 @@ void allgather_blocks(
   }
 }
 
+template <El::Device Device>
+void allgather_inverse_matrices_sizes(
```
timmoon10, line 434: Can we compute the inverse matrix sizes independently on each proc during the setup stage? This approach incurs an extra blocking communication at every KFAC step.

src/comm.cpp
```diff
@@ -138,6 +138,122 @@ void lbann_comm::split_trainers(
...
+
+
+  El::mpi::Dup(m_trainer_comm, m_combined_grid_comm);
+  if(m_create_two_models){
+    if(m_grid_type==PRIMARY_GRID){
+      El::mpi::Dup(m_primary_grid_comm, m_trainer_comm);
+    }
+    else{
+      El::mpi::Dup(m_secondary_grid_comm, m_trainer_comm);
+    }
```
timmoon10, line 231: I feel uneasy that the trainer comm no longer covers the entire trainer, but only the current subgrid. I understand that this minimizes the impact on the rest of the code. We should document `m_trainer_comm` and mention that the name is misleading.

timmoon10 (approved):
Overall looks good. [Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM387-1).

timmoon10 (changes requested):
I see compile errors on CPU systems, mostly `-Werror` failures related to unused variables.

aj-prime:
The latest commit adds support for sub-grid parallelism in GRU layers and compute-interval for inverse matrices to parallelize computation between grids. 

@timmoon10 - Thanks, I will work on the above suggestions and warnings.

include/lbann/execution_algorithms/kfac/kfac_block.hpp
```diff
@@ -94,12 +94,74 @@ class kfac_block {
...
   }
 
+  /** @brief Compute the inverse of the average Kronecker factors. */
+  virtual void compute_preconditioned_gradients(
+      lbann_comm* comm,
+      DataType learning_rate_factor,
+      bool print_matrix,
+      bool print_matrix_summary,
+      bool print_time) {
+    LBANN_ERROR("this function should be called via a sub-class.");
```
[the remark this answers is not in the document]
aj-prime: We don't have sub-grid support for BN layers; therefore, I did not make them pure virtual functions

src/execution_algorithms/kfac/kfac_block_fc_conv.cpp
```diff
@@ -398,6 +495,179 @@ kfac_block_fc_conv<Device>::get_preconditioned_grad_buffers() {
...
+  this->m_Ainv_height = my_height_A;
+  this->m_Ainv_width = my_height_A;
+  this->m_Ginv_height = my_height_G;
+  this->m_Ginv_width = my_height_G;
+
+  return my_height_A*my_height_A + my_height_G*my_height_G;
+}
+
+template <El::Device Device>
+int kfac_block_fc_conv<Device>::set_inverse_matrices(El::Matrix<DataType, Device>& output,
```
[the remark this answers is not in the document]
aj-prime: Sure

src/execution_algorithms/kfac/kfac_util.cpp
aj-prime: We are not communicating inverse matrix sizes in the current pipeline.  This function will be used when we cannot calculate matrix sizes and will be called only in the first step.

src/comm.cpp
aj-prime: Sure, I will do it.

include/lbann/execution_algorithms/kfac/kfac_block.hpp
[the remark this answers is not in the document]
benson31: This is an anti-pattern that's overly popular in Python. Following @timmoon10's suggestion, I'd recommend moving the exception to the classes you've not implemented and leave the interface pure-virtual.

timmoon10 (approved):

PR #1926, opened 2021-07-21
Add image layer for rotation operation

tnat410:
Rotate an image with an angle (in degree) around its center clockwise.
Using nearest neighbor pixel.


include/lbann/layers/image/rotation.hpp
```diff
@@ -0,0 +1,130 @@
...
+    std::stringstream err;
+    if (num_dims < 1) {
+      err << get_type() << " layer \"" << this->get_name() << "\" "
+          << "expects input with at least one dimensions, "
+          << "but input dimensions are ";
+      for (size_t i = 0; i < num_dims; ++i) {
+        err << (i > 0 ? " x " : "") << dims[i];
+      }
+      LBANN_ERROR(err.str());
+    }
```
timmoon10, line 107:
```suggestion
    if (num_dims != 3) {
      err << get_type() << " layer \"" << this->get_name() << "\" "
          << "expects 3D input in CHW format, "
          << "but input dimensions are ";
      for (size_t i = 0; i < num_dims; ++i) {
        err << (i > 0 ? " x " : "") << dims[i];
      }
      LBANN_ERROR(err.str());
    } 
```

include/lbann/layers/image/rotation.hpp
```diff
@@ -0,0 +1,130 @@
...
+    } 
+	
+    for (size_t i = 0; i < num_angle_dims; ++i) {
+      if(angle_dims[i] < 0) {
+	    err << get_type() << " layer \"" << this->get_name() << "\" "
+            << "attempted to rotate with "
+            << "negative angle (" << angle_dims[i] << ")";
+        LBANN_ERROR(err.str());
+      }		  
+    }
```
timmoon10, line 116: This isn't the right check. We expect `angle_dims` to be exactly equal to `{1}`.

src/layers/image/rotation.cpp
```diff
@@ -0,0 +1,119 @@
...
+  // Tensor dimensions
+  const auto& input_dims = this->get_input_dims(0);
+  const El::Int num_dims = input_dims.size();
+  const auto& num_samples = local_input.Width();
+  const El::Int num_channels = std::accumulate(input_dims.begin(),
+                                               input_dims.end()-2,
+                                               1,
+                                               std::multiplies<int>());
+  const El::Int input_height = input_dims[num_dims-2];
+  const El::Int input_width = input_dims[num_dims-1];
```
timmoon10, line 56: We can simplify if we assume the data is in CHW format:
```suggestion
  const El::Int num_channels = input_dims[0]
  const El::Int input_height = input_dims[1];
  const El::Int input_width = input_dims[2];
```

src/layers/image/rotation.cpp
```diff
@@ -0,0 +1,119 @@
...
+          auto input_col = static_cast<El::Int>(std::floor(rotated_col - half));
+          auto input_row = static_cast<El::Int>(std::floor(rotated_row - half));
+
+          // Input and output pixels
+          auto& pixel_output = local_output(channel * input_height * input_width
+                                                + output_row * input_width
+                                                + output_col,
+                                                sample);
+
+	  if((input_row > 0 && input_row < input_height) && (input_col > 0 && input_col < input_width)){
```
timmoon10, line 90:
```suggestion
	  if((input_row >= 0 && input_row < input_height) && (input_col >= 0 && input_col < input_width)){
```

include/lbann/layers/image/rotation.hpp
```diff
@@ -0,0 +1,130 @@
...
+  void serialize(ArchiveT& ar);
+
+  ///@}
+
+  std::string get_type() const override { return "rotation"; }
+  data_layout get_data_layout() const override { return Layout; }
+  El::Device get_device_allocation() const override { return Device; }
+
+  void fp_compute() override;
+  void bp_compute() override;
```
timmoon10, line 78: The base class implementation will automatically zero out the error signals:
```suggestion
```

timmoon10 (approved):
Overall looks good. [No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM389-1).

We should figure out the LeNet hang before merging this. Also, next steps would be adding options to choose the output size, performing bilinear interpolation to reduce aliasing, and adding a GPU implementation.

include/lbann/layers/image/rotation.hpp
tnat410: I think we should remove it. At first I thought we should not rotate with negative angle, but it should work with negative angle.

timmoon10 (changes requested):
I'm holding this PR until we figure out the hang with multiple procs per trainer.

timmoon10 (approved):

PR #1929, opened 2021-07-27
Fix compile errors in rotation layer

timmoon10:
The last few commits in PR #1926 had some correctness issues. Also, `-Werror` didn't like some unused variables.

Pinging @tnat410.


benson31 (approved):
If it passes Bamboo, good to go.

timmoon10:
[I don't see any new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM390-1).

PR #1930, opened 2021-07-27
Mutation Strategy of Replacing Random Activation

soumyadipghosh:
- Adds MutationStrategy as an argument to RandomPairwiseExchange in LTFB for Neural Architecture Search
- Adds the mutation strategy of replacing random activation layers in the winning model after a tournament in LTFB

### DIFF
--- a/include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
+++ b/include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
@@ -1,0 +1,156 @@
+////////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2014-2021, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory.
+// Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+// the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+//
+// LLNL-CODE-697807.
+// All rights reserved.
+//
+// This file is part of LBANN: Livermore Big Artificial Neural Network
+// Toolkit. For details, see http://software.llnl.gov/LBANN or
+// https://github.com/LLNL/LBANN.
+//
+// Licensed under the Apache License, Version 2.0 (the "Licensee"); you
+// may not use this file except in compliance with the License.  You may
+// obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the license.
+////////////////////////////////////////////////////////////////////////////////
+#ifndef LBANN_EXECUTION_ALGORITHMS_LTFB_MUTATION_STRATEGY_HPP_INCLUDED
+#define LBANN_EXECUTION_ALGORITHMS_LTFB_MUTATION_STRATEGY_HPP_INCLUDED
+
+#include "lbann/utils/random.hpp"
+#include "lbann/models/model.hpp"
+#include "lbann/layers/math/unary.hpp"
+#include "lbann/layers/activations/relu.hpp"
+#include "lbann/layers/activations/softmax.hpp"
+#include "lbann/layers/activations/elu.hpp"
+#include "lbann/layers/activations/leaky_relu.hpp"
+#include "lbann/layers/activations/log_softmax.hpp"
+
+namespace lbann {
+namespace ltfb {
+
+class MutationStrategy : public Cloneable<HasAbstractFunction<MutationStrategy>>
+{
+public:
+  MutationStrategy() {};
+  virtual ~MutationStrategy() = default;
+
+public:
+  /** @brief Apply a change to the model.
+   *  @param[in,out] m The model to change.
+   *  @param[in] step The current execution step in LTFB
+   */
+  virtual void mutate(model& m, const int& step) = 0;
+};
+
+// No Mutation
+class NullMutation : public Cloneable<NullMutation, MutationStrategy>
+{
+
+public:
+  NullMutation() = default; 
+  void mutate(model&, const int&) override {}
+};
+
+// Replace activation layers
+class ReplaceActivation : public Cloneable<ReplaceActivation, MutationStrategy>
+{
+private:
+  std::vector<std::string> m_activation_types = {"relu", "tanh", "softmax", "elu", "leaky relu", "log softmax"};
+
+public:
+  ReplaceActivation() = default;
+
+  std::unique_ptr<lbann::Layer> make_new_activation_layer(lbann::lbann_comm& comm,
+                                                          std::string const& new_type,
+                                                          std::string const& new_name)
+  {
+    using DataType = float;
+    #ifdef LBANN_HAS_GPU
+      constexpr El::Device Dev = El::Device::GPU;
+    #else
+      constexpr El::Device Dev = El::Device::CPU;
+    #endif
+
+    std::unique_ptr<lbann::Layer> layer;
+
+    if(new_type == "relu") {
+       layer = std::make_unique<
+           lbann::relu_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm); 
+    } else if (new_type == "tanh") {
+       layer = std::make_unique<
+            lbann::tanh_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm);
+    } else if (new_type == "softmax") {
+       layer = std::make_unique<
+            lbann::softmax_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm, softmax_mode::INSTANCE);
+    } else if (new_type == "elu") {
+       layer = std::make_unique<
+            lbann::elu_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm, 1);
+    } else if (new_type == "leaky relu") {
+       layer = std::make_unique<
+            lbann::leaky_relu_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm, 0.01);
+    } else if (new_type == "log softmax") {
+       layer = std::make_unique<
+            lbann::log_softmax_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm);
+    } else {
+       LBANN_ERROR("Unknown new layer type: ", new_type);
+    }
+    layer->set_name(new_name);
+    return layer;
+  }
+
+  void mutate(model& m, const int& step)
+  {
+    std::vector<std::string> activation_layer_names;
+
+    auto& comm = *m.get_comm();
+    
+    for (int i = 0; i < m.get_num_layers(); ++i)
+    {
+       // Creating a temp string with lower case representation
+       std::string temp_type = m.get_layer(i).get_type();
+       std::transform(temp_type.begin(), temp_type.end(),
+                                         temp_type.begin(), ::tolower);    
+
+       if (std::find(m_activation_types.cbegin(), m_activation_types.cend(), temp_type)
+                                                              != m_activation_types.cend()) {
+         activation_layer_names.push_back(m.get_layer(i).get_name());
+       }
+    }
+
+    // Generate two random numbers - one for new layer type and one for old layer name
+    int type_index = fast_rand_int(get_fast_generator(), m_activation_types.size());
+    int name_index = fast_rand_int(get_fast_generator(), activation_layer_names.size());
+
+    // Name of new layer
+    std::string new_layer_name = "new_" + m_activation_types[type_index] 
+                                        + std::to_string(step) + std::to_string(name_index);
+      
+    // Print mutation
+    std::cout << "Replacing " << activation_layer_names[name_index] << " with " << new_layer_name << std::endl;
+ 
+    // Replace layer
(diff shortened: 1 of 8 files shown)


include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
```diff
@@ -0,0 +1,156 @@
...
+public:
+  NullMutation() = default; 
+  void mutate(model&, const int&) override {}
+};
+
+// Replace activation layers
+class ReplaceActivation : public Cloneable<ReplaceActivation, MutationStrategy>
+{
+private:
+  std::vector<std::string> m_activation_types = {"relu", "tanh", "softmax", "elu", "leaky relu", "log softmax"};
```
timmoon10, line 68: Softmax and log-softmax are questionable to use as activations, except at the end of the network.

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
```diff
@@ -0,0 +1,156 @@
...
+       }
+    }
+
+    // Generate two random numbers - one for new layer type and one for old layer name
+    int type_index = fast_rand_int(get_fast_generator(), m_activation_types.size());
+    int name_index = fast_rand_int(get_fast_generator(), activation_layer_names.size());
+
+    // Name of new layer
+    std::string new_layer_name = "new_" + m_activation_types[type_index] 
+                                        + std::to_string(step) + std::to_string(name_index);
```
timmoon10, line 136: If you are replacing the layer, it should be fine to reuse the old name. Right now we'll get layer names like `new_new_new_layer123`.

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
```diff
@@ -0,0 +1,156 @@
...
+ 
+    // Replace layer
+    m.replace_layer(make_new_activation_layer(comm, m_activation_types[type_index], new_layer_name), 
+                    activation_layer_names[name_index]);
+
+    // Erase old layer from activation layer list
+    activation_layer_names.erase(activation_layer_names.cbegin() + name_index);
+
+    // Add new layer to activation layer list
+    activation_layer_names.push_back(new_layer_name);
```
timmoon10, line 149: It isn't necessary to update this if we aren't caching it:
```suggestion
```

include/lbann/execution_algorithms/ltfb/random_pairwise_exchange.hpp
```diff
@@ -194,6 +200,10 @@ class RandomPairwiseExchange final
    */
   std::unique_ptr<ExchangeStrategy> m_comm_algo;
 
+  /** @brief The strategy for mutation of a model - insert more details later
```
timmoon10, line 203:
```suggestion
  /** @brief The strategy for mutation of a model
```

src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
```diff
@@ -307,15 +312,15 @@ void RandomPairwiseExchange::select_next(model& m,
     // FIXME (trb 03/18/21): This is ... not great. We need to
     // unravel the "fake" polymorphism in the model non-hierarchy
     // soon.
-    using DAGModel = directed_acyclic_graph_model;
-    auto& local_model = dynamic_cast<DAGModel&>(m);
-    auto& partner_dag_model = dynamic_cast<DAGModel&>(*partner_model);
-    local_model = std::move(partner_dag_model);
+
```
timmoon10, line 315: We are currently always keeping the local model:
```suggestion
    using DAGModel = directed_acyclic_graph_model;
    auto& local_model = dynamic_cast<DAGModel&>(m);
    auto& partner_dag_model = dynamic_cast<DAGModel&>(*partner_model);
    local_model = std::move(partner_dag_model);
    
```

timmoon10 (changes requested):
Overall looks good. I do see we always choose the local model in LTFB, I presume for debugging. We should revert that change.

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
```diff
@@ -0,0 +1,156 @@
...
+    std::vector<std::string> activation_layer_names;
+
+    auto& comm = *m.get_comm();
+    
+    for (int i = 0; i < m.get_num_layers(); ++i)
+    {
+       // Creating a temp string with lower case representation
+       std::string temp_type = m.get_layer(i).get_type();
+       std::transform(temp_type.begin(), temp_type.end(),
+                                         temp_type.begin(), ::tolower);
```
benson31, line 122:
```suggestion
       std::transform(begin(temp_type),
                      end(temp_type),
                      begin(temp_type),
                      [](unsigned char c){ return ::tolower(c); });    
```

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
```diff
@@ -0,0 +1,156 @@
...
+            lbann::elu_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm, 1);
+    } else if (new_type == "leaky relu") {
+       layer = std::make_unique<
+            lbann::leaky_relu_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm, 0.01);
+    } else if (new_type == "log softmax") {
+       layer = std::make_unique<
+            lbann::log_softmax_layer<DataType, data_layout::DATA_PARALLEL, Dev>>(&comm);
+    } else {
+       LBANN_ERROR("Unknown new layer type: ", new_type);
+    }
```
benson31, line 106: Shouldn't this inherit the type, layout, and device allocation of the layer it's replacing?

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+       if (std::find(m_activation_types.cbegin(), m_activation_types.cend(), temp_type)
+                                                              != m_activation_types.cend()) {
+         activation_layer_names.push_back(m.get_layer(i).get_name());
+       }
+    }
+
+    // Generate two random numbers - one for new layer type and one for old layer name
+    int type_index = fast_rand_int(get_fast_generator(), m_activation_types.size());
+    int name_index = fast_rand_int(get_fast_generator(), activation_layer_names.size());
```
benson31, line 132:
```suggestion
    int const type_index = fast_rand_int(get_fast_generator(), m_activation_types.size());
    int const name_index = fast_rand_int(get_fast_generator(), activation_layer_names.size());
```

src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
benson31: This isn't right. Restoring this means that `local_model`, which is an alias for `m`, will now be set to the `partner_model` (i.e., the winning model in this case). But the documentation says that it's the *losing* model that gets mutated.

In summary: the way that it's written currently, without committing @timmoon10's suggestion, the losing model, which is the local model, gets mutated and returns. This breaks the current LTFB implementation in a fairly obvious way. Merging @timmoon10's suggestion restores the normal LTFB behavior when using the `NullMutation` strategy, but breaks the documented behavior when using `ReplaceActivation`. There's clearly a deeper logic error that needs to be resolved here.

benson31 (changes requested):
1. The `mutation_strategy.hpp` header should not contain implementations. There are no templates, so please move everything to a suitably located `.cpp` file.
2. Run `clang-format` on everything.
3. There are some severe algorithmic issues that need to be addressed beyond the simple C++ issues.

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
soumyadipghosh: Removed softmax, log-softmax; Added sigmoid

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
soumyadipghosh: Keeping layer name same.

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
soumyadipghosh: Removed

src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
soumyadipghosh: Yes, fixed this issue. The logic now is that if a model loses in a tournament, the winning model is copied over to it and then mutated.

python/lbann/core/training_algorithm.py
```diff
@@ -186,6 +186,29 @@ def do_export_proto(self):
         params.local_training_algorithm.CopyFrom(self.local_algo.export_proto())
         return params
 
+class MutationStrategy:
+    """The strategy for mutation after a tournament in LTFB.
+       INSERT FURTHER DESCRIPTION HERE.
```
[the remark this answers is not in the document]
soumyadipghosh: Added description.

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
soumyadipghosh: done.

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
soumyadipghosh: Has to be resolved at compile time, so cant use getter functions.

src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
```diff
@@ -134,23 +136,26 @@ bool local_wins(EvalType local,
...
 
 RandomPairwiseExchange::RandomPairwiseExchange(
   std::string metric_name,
   metric_strategy winner_strategy,
-  std::unique_ptr<ExchangeStrategy> comm_algo)
+  std::unique_ptr<ExchangeStrategy> comm_algo,
+  std::unique_ptr<MutationStrategy> mutate_algo)
   : RandomPairwiseExchange({{metric_name, winner_strategy}},
-                           std::move(comm_algo))
+                           std::move(comm_algo), std::move(mutate_algo))
```
benson31, line 152:
```suggestion
                           std::move(comm_algo),
                           std::move(mutate_algo))
```

src/execution_algorithms/ltfb/random_pairwise_exchange.cpp
```diff
@@ -134,23 +136,26 @@ bool local_wins(EvalType local,
...
 {}
 
 RandomPairwiseExchange::RandomPairwiseExchange(
   RandomPairwiseExchange const& other)
-  : m_metrics{other.m_metrics}, m_comm_algo{other.m_comm_algo->clone()}
+  : m_metrics{other.m_metrics}, m_comm_algo{other.m_comm_algo->clone()},
```
benson31, line 157:
```suggestion
  : m_metrics{other.m_metrics},
    m_comm_algo{other.m_comm_algo->clone()},
```

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
```diff
@@ -0,0 +1,79 @@
...
+   *  @param[in] new_type The type of the new activation layer (ReLU or tanh
+   * etc)
+   *  @param[in] new_name The name of the new activation layer
+   */
+
+  std::unique_ptr<lbann::Layer>
+  make_new_activation_layer(lbann::lbann_comm& comm,
+                            std::string const& new_type,
+                            std::string const& new_name);
+  void mutate(model& m, const int& step);
```
benson31, line 74:
```suggestion
  void mutate(model& m, const int& step) final;
```

include/lbann/execution_algorithms/ltfb/mutation_strategy.hpp
```diff
@@ -0,0 +1,79 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing
+// permissions and limitations under the license.
+////////////////////////////////////////////////////////////////////////////////
+#ifndef LBANN_EXECUTION_ALGORITHMS_LTFB_MUTATION_STRATEGY_HPP_INCLUDED
+#define LBANN_EXECUTION_ALGORITHMS_LTFB_MUTATION_STRATEGY_HPP_INCLUDED
+
+#include "lbann/layers/layer.hpp"
```
benson31, line 29:
```suggestion
#include "lbann/models/model.hpp"
#include "lbann/utils/cloneable.hpp"
```
Once `make_new_activation_layer` is moved out of this file, it won't (explicitly) depend on `std::string`, `lbann::lbann_comm` or `lbann::Layer` anymore. This is the correct include set; however, the `model.hpp` could be replaced with a forward declaration.

benson31 (changes requested):

benson31 (approved):

timmoon10 (approved):
If this passes Bamboo, then it's good to go.

timmoon10 (changes requested):
We've identified a bug where different procs in a trainer can choose to replace different layers.

soumyadipghosh:
Addressed the bug mentioned above. Pinging @timmoon10

timmoon10 (approved):
Looks good. If [the Bamboo build](https://lc.llnl.gov/bamboo/browse/LBANN-TIM392-2) has no new failures, it's ready to go.

PR #1932, opened 2021-07-30
Refactored SMILES data reader to use the new data_reader_sample_list

bvanessen:
class.


benson31 (approved):
I couldn't suggest the edit I made (GH doesn't let you make suggestions on deleted lines, apparently... Seems like "undelete this line" should be a valid suggestion, but what do I know?). For that change, structured bindings fit the use-case that you needed, and I'd prefer to use language features rather than needlessly bloating the API.

Assuming this passes Bamboo, I'm good with it.

PR #1935, opened 2021-08-02
Disable K-FAC GRU implementation in GPU builds without cuDNN

timmoon10:
PR #1924 broke the AMD build since it has GPUs but doesn't support GRUs. I've put preprocessor macros so that the K-FAC GRU implementations are only compiled for GPU when cuDNN support is provided. On AMD systems, attempting to use GRUs in K-FAC will lead to a run-time error (which would be expected since using the GRU layer would also lead to a run-time error).

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM393-1).

### DIFF
--- a/src/execution_algorithms/kfac.cpp
+++ b/src/execution_algorithms/kfac.cpp
@@ -261,7 +261,7 @@ void KFAC::train(
   }
 
   sgd_context.stop_timer();
-  
+
 
   // Reset the model back to the training execution context prior to
   // end of training callbacks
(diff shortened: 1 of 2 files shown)


timmoon10:
I see the same Bamboo errors as [before the weekend](https://lc.llnl.gov/bamboo/browse/LBANN-NIGHTD-X86CPUC-1133/test), so I think this is ready to merge. Note that we have been experiencing many errors on Corona, I think from PR #1927.

mrwyattii (approved):
LGTM

PR #1938, opened 2021-08-03
Update the perturb weight callback

tnat410:
Update the perturb weight callback to support better range of weight values.
Add upper bound parameter for weight values.
Add lower bound parameter for weight values.
Add scale parameter for scaling normal distribution N(1,0) for the perturb values.
Add perturbation probability parameter.
Make the callback only perturbs weight during training.

### DIFF
--- a/include/lbann/callbacks/perturb_weights.hpp
+++ b/include/lbann/callbacks/perturb_weights.hpp
@@ -45,8 +45,9 @@ public:
    *  @param batch_interval Number of training mini-batch steps 
    *  @param output_name   Name of weight layer
    */
-  perturb_weights(std::string output_name,
-               El::Int batch_interval = 1);
+  perturb_weights(EvalType upper, EvalType lower, EvalType scale, 
+		  std::string output_name,
+                  El::Int batch_interval = 1);
 
   perturb_weights* copy() const override { return new perturb_weights(*this); }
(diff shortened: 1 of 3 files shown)


include/lbann/callbacks/perturb_weights.hpp
```diff
@@ -71,6 +76,18 @@ class perturb_weights : public callback_base {
...
+   /*
+   *  upper.
+   *  lower.
+   *  scale.
+   *  perturb_probability.
+   */	
+  EvalType m_upper;
+  EvalType m_lower;
+  EvalType m_scale;
+  EvalType m_perturb_probability;
```
timmoon10, line 89: Let's properly document these variables. If you comment with `///` or `/**`, the comments can be parsed by Doxygen to generate user-friendly documentation.
```suggestion
  /// Upper bound of the perturbed weight
  EvalType m_upper;
  /// Lower bound of the perturbed weight
  EvalType m_lower;
  /// Scale of normal distribution N(1,0) for the perturbed weight
  EvalType m_scale;
  /// The probability of perturbing the weight
  EvalType m_perturb_probability;
```

timmoon10 (approved):
Overall looks good. If we merge my comment and if the [continuous integration tests](https://lc.llnl.gov/bamboo/browse/LBANN-TIM397-1) pass, we can merge this.

PR #1939, opened 2021-08-04
Run ExaGAN discriminators fully within distconv

timmoon10:
Changes:
- distconv has poor support for strided convolution. I've worked around this by making the 5x5x5 convolutions have unit stride, and following them up with 2x2x2 max pooling.
- Fixed bug where distconv convolutions assume the weights have optimizers. This was causing some problems since one copy of the discriminator doesn't have any optimizers.

This depends on https://github.com/LLNL/DiHydrogen/pull/59. [Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM395-1).


samadejacobs (approved):
LGTM

PR #1940, opened 2021-08-04
Add MIOpen Workspace Size Query for Convolution Layer

mrwyattii:
Added the functions in `dnn_lib` namespace for MIOpen corresponding to `miopenConvolution*GetWorkSpaceSize` functions, so that workspaces are properly sizes.  Added equivalent functions for cuDNN that return `1 << 30` (same as prior behavior).


benson31 (approved):
LGTM.

PR #1941, opened 2021-08-05
Remove Catch2 dependency if unit tests are not enabled

timmoon10:
PR #1925 introduced a bug where the HDF5 data utilities were always built with their unit tests, which broke the default build.

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM396-2).


benson31 (approved):

bvanessen (approved):
I ran into the same problems yesterday and was working with a local patch.

PR #1942, opened 2021-08-06
Update the rotation layer 

tnat410:
Performing bilinear interpolation instead of using nearest neighbor.

### DIFF
--- a/src/layers/image/rotation.cpp
+++ b/src/layers/image/rotation.cpp
@@ -37,8 +37,8 @@ void rotation_layer<TensorDataType, Layout, Device>::fp_compute() {
   // Useful constants
   constexpr DataType Pi = M_PI;
   constexpr DataType degree = 180;
-  constexpr DataType half = 0.5;
   constexpr DataType zero = 0;
+  constexpr DataType one = 1;
 
   // Input and output tensors
   const auto& local_input = this->get_local_prev_activations();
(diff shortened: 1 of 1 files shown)


src/layers/image/rotation.cpp
```diff
@@ -66,30 +66,62 @@ void rotation_layer<TensorDataType, Layout, Device>::fp_compute() {
...
                                                 + output_row * input_width
                                                 + output_col,
                                                 sample);
 
-	  if((input_row >= 0 && input_row < input_height) && (input_col >= 0 && input_col < input_width)){
-          	auto& pixel_input = local_input(channel * input_height * input_width
-                                       	     	  + input_row * input_width
-                                           	  + input_col,
-                                                  sample);
+	  if((input_col >= 0 && input_col < input_width) && (input_row >= 0 && input_row < input_height)){
```
timmoon10, line 86: The behavior is slightly incorrect on the left and bottom edges. If `input_col==input_width-1`, then `input_col0==input_col1==input_width-1`. The effect is a constant region on the edge.
```suggestion
	  if((input_col >= 0 && input_col < input_width-1) && (input_row >= 0 && input_row < input_height-1)){
```

timmoon10 (changes requested):
One small fix for the right and bottom edge, but otherwise looks good. [Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM399-1).

timmoon10 (approved):

PR #1943, opened 2021-08-06
Fix initialization order warning in "perturb weights" callback

timmoon10:
@benson31 reports a compilation warning since the initialization order doesn't match the order of class members. I've also taken the opportunity to tweak the documentation.

[Running Bamboo on Corona](https://lc.llnl.gov/bamboo/browse/LBANN-TIM398-X86CPUC-1). Pinging @tnat410.


benson31 (approved):
Fixes the issue! Thanks!

PR #1944, opened 2021-08-06
Add an image layer for multiple operation

tnat410:
Rotate a image clockwise around its center
Then shear it based on set of (X,Y ) shear factor.
Then translate it based on set of (X,Y) translate factor.
Pinging @timmoon10 , @samadejacobs

### DIFF
--- a/include/lbann/layers/image/CMakeLists.txt
+++ b/include/lbann/layers/image/CMakeLists.txt
@@ -2,6 +2,7 @@
 set_full_path(THIS_DIR_HEADERS
   bilinear_resize.hpp
   rotation.hpp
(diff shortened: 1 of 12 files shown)


timmoon10 (changes requested):
Overall looks good. The major problem is that the name "composite layer" is very vague. How about "composite image transform layer"?

[Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM401-1).

timmoon10 (approved):
Looks good to merge, pending no new errors in [Bamboo](https://lc.llnl.gov/bamboo/browse/LBANN-TIM401-3).

PR #1946, opened 2021-08-10
Accommodate large channel dim in CUDA kernels for channel-wise layers

timmoon10:
I've been seeing CUDA errors since the maximum CUDA grid size is 65K in the y and z dimensions. All of these kernels were written to accommodate arbitrarily large dimensions (up to the limits of `int`), but the grid dims hadn't been properly configured. [Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM400-1).

This is a quick hack, but I wonder what's the best way to handle the grid dim limits. It may be useful to put a named constant instead of hard-coding 65535, especially if AMD systems have different constraints.


mrwyattii (approved):
LGTM

timmoon10:
I've generalized by adding a `gpu_lib::clip_grid_dims` function that will put the correct max grid dims for CUDA and HIP. I see [no new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM400-4).

I've also fixed the grid dims for the gather and scatter layers. Pinging @szaman19.

PR #1947, opened 2021-08-10
Remove the explicit finding of HDF5

benson31:
Now, as it should have been all along, this is Conduit's task.

In versions past of Conduit, certain issues could arise when finding
the HDF5 package if it was built differently from how Conduit expected
it. This changed a few versions back, and the Conduit mechanisms for
finding HDF5 now seem (I hope) robust to these issues.


mrwyattii (approved):
LGTM

bvanessen (approved):

PR #1948, opened 2021-08-10
Mutation Strategy of Replacing Random Convolution

soumyadipghosh:
- Adds the mutation strategy of replacing random convolution layers by changing the kernel size and number of channels
- The padding is appropriately adjusted to account for changing kernel size so that height and width dimensions remain the same
- A shim convolution layer (with kernel size 1) is inserted after changing the number of channels in a convolution layer for matching the channel dimensions.

### DIFF
--- a/src/execution_algorithms/ltfb/mutation_strategy.cpp
+++ b/src/execution_algorithms/ltfb/mutation_strategy.cpp
@@ -42,9 +42,9 @@
 #include "lbann/utils/random.hpp"
 
 #ifdef LBANN_HAS_GPU
-  constexpr El::Device Dev = El::Device::GPU;
+constexpr El::Device Dev = El::Device::GPU;
 #else
-  constexpr El::Device Dev = El::Device::CPU;
+constexpr El::Device Dev = El::Device::CPU;
 #endif
 
 namespace lbann {
@@ -202,7 +202,7 @@ void ReplaceActivation::mutate(model& m, const int& step)
 
 void ReplaceConvolution::mutate(model& m, const int& step)
 {
-  static std::vector<int> const kernels = {5, 7, 9};
+  static std::vector<int> const kernels = {3, 5, 7, 9};
   static std::vector<int> const channels = {6, 16, 32};
 
   std::vector<int> conv_layer_indices; // Indices of all convolution layers
@@ -292,15 +292,9 @@ void ReplaceConvolution::mutate(model& m, const int& step)
     auto shim_channels = child.get_output_dims(0)[0];
 
     // Replace shim layer: O/P channels stay the same but I/P channels change
-    m.remove_layer(shim_layer_name);
-    m.insert_layer(make_new_convolution_layer(1,
-                                              0,
-                                              1,
-                                              shim_channels,
-                                              1,
-                                              1,
-                                              shim_layer_name),
-                   name);
+    m.replace_layer(
+      make_new_convolution_layer(1, 0, 1, shim_channels, 1, 1, shim_layer_name),
+      shim_layer_name);
   }
   else {
     std::cout << "Creating shim layer for layer " << name << std::endl;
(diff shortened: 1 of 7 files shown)


src/models/model.cpp
```diff
@@ -1879,6 +1895,12 @@ void model::remove_layer(std::string const& removable_layer_name) {
   child.replace_parent_layer(l.get_parent_layer_pointer(0), child.find_parent_layer_index(l));
   parent.replace_child_layer(l.get_child_layer_pointer(0), parent.find_child_layer_index(l));
 
+  // Remove weights for the old layer
+  auto old_weights_ptrs = m_layers[removable_layer_index]->get_weights_pointers();
+  for (const auto w : old_weights_ptrs) {
+     this->remove_weights(std::shared_ptr<weights>(w)->get_name());
+  }
```
timmoon10, line 1902: This behavior will cause problems if the layer is sharing weights. It's fine since that's not our current use-case, but it would be good to mention that behavior in the documentation.

src/models/model.cpp
```diff
@@ -1920,6 +1942,12 @@ void model::replace_layer(OwningLayerPtr&& new_layer, std::string const& old_lay
   parent.replace_child_layer(new_layer, parent.find_child_layer_index(l));
   new_layer->add_parent_layer(m_layers[old_layer_index-1]);
 
+  // Remove weights for the old layer
+  auto old_weights_ptrs = m_layers[old_layer_index]->get_weights_pointers();
+  for (const auto w : old_weights_ptrs) {
+     this->remove_weights(std::shared_ptr<weights>(w)->get_name());
+  }
```
timmoon10, line 1949: We should mention this behavior in the documentation.

timmoon10 (approved):
Overall looks good. [Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM403-1).

For next steps, it may be a good idea if the mutations were incremental. Like there's a certain probability the kernel size increases by 2 or decreases by 2, or a probability that the number of channels doubles or halves. Also, taking advantage of the block structure in ResNets could be useful, instead of manually inserting shim layers.

timmoon10 (changes requested):
@soumyadipghosh reports that this causes an error.

src/models/model.cpp
```diff
@@ -1920,6 +1942,12 @@ void model::replace_layer(OwningLayerPtr&& new_layer, std::string const& old_lay
   parent.replace_child_layer(new_layer, parent.find_child_layer_index(l));
   new_layer->add_parent_layer(m_layers[old_layer_index-1]);
```
timmoon10, line 1943: This may do something weird if the layer graph isn't sequential:
```suggestion
  new_layer->add_parent_layer(l.get_parent_layer_pointer(0));
```

src/execution_algorithms/ltfb/mutation_strategy.cpp
```diff
@@ -160,5 +200,120 @@ void ReplaceActivation::mutate(model& m, const int& step)
                   activation_layer_names[name_index]);
 }
 
+void ReplaceConvolution::mutate(model& m, const int& step)
+{
+  static std::vector<int> const kernels = {5, 7, 9};
```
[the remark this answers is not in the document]
soumyadipghosh: Done.

src/models/model.cpp
soumyadipghosh: Added a comment for both

src/models/model.cpp
soumyadipghosh: This is an important fix!

soumyadipghosh:
Without regularization, this PR works fine. Pinging @timmoon10

timmoon10 (approved):
LGTM. If we see no new errors in the [Bamboo build](https://lc.llnl.gov/bamboo/browse/LBANN-TIM403-2), we can merge.

PR #1957, opened 2021-08-30
Bugfix for sample list

bvanessen:
Fixed a bug where the front of the priority queue was accessed when it was empty.


benson31 (approved):
This fixes one of the issues I'm observing.

PR #1961, opened 2021-09-07
Data augmentation in training mode

samadejacobs:
This PR extends previous implementations of data transformation/augmentation to work in training mode only.

### DIFF
--- a/src/layers/image/bilinear_resize.cpp
+++ b/src/layers/image/bilinear_resize.cpp
@@ -26,6 +26,7 @@
 
 #define LBANN_BILINEAR_RESIZE_LAYER_INSTANTIATE
 #include "lbann/layers/image/bilinear_resize.hpp"
+#include "lbann/models/model.hpp"
 
 namespace lbann {
 
@@ -40,6 +41,11 @@ void bilinear_resize_layer<TensorDataType, Layout, Device>::fp_compute() {
   const auto& local_input = this->get_local_prev_activations();
   auto& local_output = this->get_local_activations();
 
+  const auto& mode = this->m_model->get_execution_context().get_execution_mode();
+  if (mode != execution_mode::training) {
+    El::Copy(local_input, local_output);
+    return;
+  }
   // Dimensions
   const auto& input_dims = this->get_input_dims();
   const auto& num_dims = input_dims.size();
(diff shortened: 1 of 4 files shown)


src/layers/image/bilinear_resize.cpp
```diff
@@ -40,6 +41,11 @@ void bilinear_resize_layer<TensorDataType, Layout, Device>::fp_compute() {
   const auto& local_input = this->get_local_prev_activations();
   auto& local_output = this->get_local_activations();
 
+  const auto& mode = this->m_model->get_execution_context().get_execution_mode();
+  if (mode != execution_mode::training) {
+    El::Copy(local_input, local_output);
+    return;
+  }
```
timmoon10, line 48: The bilinear resize layer takes a fixed-size input and resizes it to a fixed-size output. I don't see how we can disable it at evaluation time.
```suggestion
```

src/layers/image/bilinear_resize.cu
```diff
@@ -122,6 +123,11 @@ void bilinear_resize_layer<TensorDataType, Layout, Device>::fp_compute() {
   const auto& local_input = this->get_local_prev_activations();
   auto& local_output = this->get_local_activations();
 
+  const auto& mode = this->m_model->get_execution_context().get_execution_mode();
+  if (mode != execution_mode::training) {
+    El::Copy(local_input, local_output);
+    return;
+  }
```
timmoon10, line 130: I don't think it makes sense to disable this during evaluation:
```suggestion
```

timmoon10 (changes requested):
Can't we already implement this behavior by using random layers to control the transformations (at least for the crop, rotation, and composite layers)? The random layers have an option to always output the mean value during evaluation time.

```python
img = lbann.Input()
angle = lbann.Uniform(min=-45, max=45, training_only=True)
rot_img = lbann.Rotation(img, angle)
```

This PR's approach adds complexity by making these layers have to keep track of the execution mode.

Other thoughts:
- It doesn't make sense to disable the bilinear resize layer during evaluation time. The input and output dims will always be different, so copying the values isn't valid.
- If we do want this behavior, it should be clearly documented.
- This approach is not valid if we modify the rotation layer and composite layer to have different input and output dimensions.
- If you're not careful, this approach might lead to different data distributions during training and evaluation, which could result in wonky validation metrics.

PR #1963, opened 2021-09-07
Add a bunch of f(x,C)-type operators

benson31:
A common pattern in LBANN models, especially objective function computations, is something like:
```python
x = lbann.SomeBinaryOp(y, lbann.Constant(value=some_value))
```
At a minimum, this will allocate a `minibatch_size x sample_size` (basically `dims(y.Matrix())`) matrix with each entry set to `some_value`. While this may not be the most pressing of issues, this is clearly a pessimization. So this PR introduces some operators that translate these into unary operators with an additional constant parameter associated with them.

The following operations are implemented:

- `f(x; C) = x + C` (tested), "add constant"
- `f(x; C) = x - C` (tested), "subtract constant"
- `f(x; C) = C - x` (tested), "subtract from constant"
- `f(x; C) = C*x` (tested), "scale"
- `f(x; C) = max(C, x)`, "max constant"
- `f(x; C) = min(C, x)`, "min constant"
- `f(x; C) = x == C` (tested), "equal constant"
- `f(x; C) = x != C` (tested), "not equal constant"
- `f(x; C) = x < C`, "less constant"
- `f(x; C) = x > C`, "greater constant"
- `f(x; C) = x <= C`, "less-equal constant"
- `f(x; C) = x >= C`, "greater-equal constant"

(Feedback on the names is welcome...)

TODO:

- [x] GPU implementations/instantiations.


bvanessen (approved):

benson31:
Bamboo matches develop.

timmoon10 (approved):
These are going to be really convenient. :shipit:

PR #1964, opened 2021-09-10
Use default-allocated GPU memory for long-lived buffers

timmoon10:
@benson31 and @tnat410 have complained about high memory usage in Resnet-50, such that we can no longer run with a mini-batch of 256 on a Lassen node. In an ideal world, I estimate the memory usage on a single GPU would look like:
```
(300 MB) + (150 MB)*mbsize
```
However, I see:
```
(2.5 GB) + (200 MB)*mbsize
```
Nsight suggests that the excess in the mini-batch term is from over-allocation in the CUB memory pool. When I allocate the activation and weight buffers with `cudaMalloc` instead of the memory pool:
```
(2.4 GB) + (170 MB)*mbsize
```
This memory reduction allows for a 256 mini-batch on a Lassen node. I don't see any effect on performance and [Bamboo has no new failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM410-2).

Next steps are to reduce the excess in the constant term. At least 1 GB is from the hard-coded workspace size for cuDNN convolutions. Nsight suggests that the remaining 1 GB is never actually accessed, but I'm not sure where it's allocated.

### DIFF
--- a/src/layers/data_type_layer.cpp
+++ b/src/layers/data_type_layer.cpp
@@ -680,12 +680,11 @@ setup_matrices(const El::Grid& grid) {
   }
 
 #ifdef LBANN_HAS_GPU
-  // Use default-allocated device memory for forward prop matrices
-  // Note: Using the GPU memory pool uses more memory than directly
-  // allocating GPU buffers. Since forward prop buffers are rarely
-  // reallocated, the memory pool also has no performance benefit.
+  // Use default-allocated GPU memory for forward prop matrices
+  // Note: GPU memory pool uses more memory and these buffers are
+  // rarely reallocated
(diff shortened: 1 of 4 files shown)


bvanessen (changes requested):
@timmoon10 Please add some command line option to set these flags.

timmoon10:
I'm working on adding an environment variable to disable this optimization, but I wonder if it is worth the extra complexity. I'm hard-pressed to think of situations where we would not want to use this. Allocating with a GPU memory pool may somewhat improve performance in cases where we are frequently constructing and destroying models, but I suspect it would be minor compared to the cost of things like serialization.

PR #1965, opened 2021-09-16
Add binary-with-constant layers to the Python Front-End

benson31:
These were accidentally omitted from #1963.


bvanessen (approved):

timmoon10 (approved):
:shipit:

PR #1970, opened 2021-09-18
Optionally use default-allocated GPU memory for long-lived buffers

timmoon10:
This is a variant of PR #1964 that provides an option to disable that memory optimization.

[Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM412-1). Closes #1964.


bvanessen (approved):

bvanessen:
@timmoon10 So just to be clear, the defaults on this will be to use the direct allocation.  This seems reasonable, but having the flag lets us easily do performance evaluation.

PR #1977, opened 2021-09-28
Set the operator data type to match the layer's data type

benson31:
This addresses an issue raised by @timmoon10 that arises when changing the layer precision in the python front-end. The `OperatorLayer` will only support a single operator (that change coming in a future PR; adding an ASSERT to hold us over until then), and it doesn't make sense to allow the types to be different.


benson31:
#1978 is better. Using that instead.

PR #1979, opened 2021-09-30
Make the learning rate in the optimizer not a dependent type.

benson31:
Optimizers were storing the learning rate at their native precision. I don't think there's an advantage vs universally using `double`. So I did that. There's also some code movement and formatting. Finally I cleaned up some issues to get things compiling cleanly again.

This does not address the systemic issues in any depth, and we probably shouldn't bother until we revamp the callbacks.


timmoon10 (approved):
LGTM

PR #1981, opened 2021-10-06
PROBIESNet

samadejacobs:
A regression model for predicting laser proton beam acceleration parameters (i.e., divergence angle, temperature, energies etc) from proton beam images, essentially a 6 conv + fc layers. This PR works fine without new data reader/input layer refactor, a placeholder is there to use the input layer API pinging @bvanessen.


samadejacobs:
> @samadejacobs I have fixed these issues and tested locally.

Tested local too, LGTM.

PR #1983, opened 2021-10-13
Rename some training algorithm symbols

benson31:
This PR renames the following symbols. There are no changes in behavior or functionality; it is merely cosmetic.

- `training_algorithm` -> `TrainingAlgorithm`
- `exectution_context` -> `ExecutionContext`
- `termination_criteria` -> `TerminationCriteria`
- `sgd_training_algorithm` -> `SGDTrainingAlgorithm`
- `sgd_execution_context` -> `SGDExecutionContext`
- `sgd_termination_criteria` -> `SGDTerminationCriteria`
- `ltfb::ExecutionContext` -> `ltfb::LTFBExecutionContext`
- `ltfb::TerminationCriteria` -> `ltfb::LTFBTerminationCriteria`
- `kfac::ExecutionContext` -> `kfac::KFACExecutionContext`

The `kfac::` and `ltfb::` symbols were renamed so that the base class names match the SGD pattern. Note that KFAC just uses `SGDTerminationCriteria`, so there is not a separate symbol to rename.

I thought about renaming `sgd_training_algorithm` to just `SGD`, but then we'd have two classes `lbann::sgd` and `lbann::SGD`, which seems unnecessarily confusing for humans reading and discussing the code (I wouldn't complain if one were, say, a function that created instances of the other class; it's the fact that they're semi-related classes that suggests undue confusion...).

I also thought about renaming `sgd_training_algorithm` to be `BatchedIterativeTrainingAlgorithm`, but I felt that `SGDTrainingAlgorithm` was a more minimal change that accomplished my goal. I'm happy to discuss this and just rename to `BatchedIterativeTrainingAlgorithm` if that's the group consensus.

Bamboo is running.


benson31:
Update: execution contexts belong with the execution algorithms. Since I was touching all these files anyway, I have added one more commit that relocates the files and adjusts the `#include` paths to them. Either this added no new files to the diff, or someone was `#include`-ing a file they ought not to have been.

bvanessen (approved):

PR #1994, opened 2021-10-29
Minor cleanup of sub-grid parallel layers

timmoon10:
This PR consists of some cleanup I did while debugging a race condition with sub-grid parallelism (https://github.com/LLNL/Elemental/pull/126). Most importantly, it adds a unit test to test various combinations of layers to fork/join sub-graph regions. This seems to fix my issue, but we will need to do some significant refactoring before sub-grid parallelism is ready for general use.

Depends on https://github.com/LLNL/Elemental/pull/126.


timmoon10:
The old infrastructure for sub-graph parallelism should be replaced by the infrastructure introduced in https://github.com/LLNL/lbann/pull/2025.

PR #1997, opened 2021-11-03
Refactor deconvolution layer API to match PyTorch

timmoon10:
One of the annoyances with our deconvolution layers is that we can't make its dimensions match the corresponding convolution layer if there is a stride. The output dimensions are ambiguous and we always output the smallest valid tensor. To fix this, I am copying [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html) and adding a `output_padding` option. If we have a stride-2 convolution, we can recover the original dimensions by applying a stride-2 deconvolution with `output_padding=1`.

While I was tinkering around, I've cleared out a lot of the cruft in our  API, expanded documentation, and added a unit test. The option names match PyTorch exactly (except for `has_bias`). String lists have been replaced with `list` s of `int` s. The old API still works (the only change is that `has_bias` defaults to `True`), but it has been marked as deprecated. If we don't experience too many problems, we can apply a similar approach to clean up the convolution layers.

[Bamboo is running](https://lc.llnl.gov/bamboo/browse/LBANN-TIM425-2).


timmoon10:
[No new CI errors](https://lc.llnl.gov/gitlab/moon13/lbann/-/pipelines/152498). As of 3/9/22, this is ready to merge.

benson31 (approved):
LGTM

benson31:
From the "software development" side of things, an important contribution of this PR is the start of something I've wanted to do for a while: using `repeated` fields instead of space-delimited strings. The latter may have made sense (in the sense that they're easier on human typers) when we were manually writing prototext, but in the PFE era, these add overhead in parsing the protobuf messages inside LBANN. Thanks, @timmoon10, for starting us down this path and I hope we can migrate the other layers/etc. in this direction (I have a feeling that the PFE is flexible enough to allow us to do this with no user impact).

samadejacobs (approved):

PR #2001, opened 2021-11-08
Use tournament set in LTFB with truncation selection exchange

timmoon10:
It's confusing that we are using different validation sets for the two LTFB communication algorithms, truncation selection exchange and random pairwise exchange. TSE uses the test set and RPE uses the tournament set. To make apples-to-apples comparisons easier, I propose we use the tournament set for both techniques.

[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM428-2). Pinging @tnat410.


src/execution_algorithms/ltfb/truncation_selection_exchange.cpp
```diff
@@ -102,18 +102,18 @@ EvalType TruncationSelectionExchange::evaluate_model(model& m,
...
 
   // Mark the data store as loading - Note that this is a temporary fix
   // for the current use of the tournament
-  m.mark_data_store_explicitly_loading(execution_mode::testing);
+  m.mark_data_store_explicitly_loading(execution_mode::tournament);
 
-  // Evaluate model on test (or validation?) set
-  get_trainer().evaluate(&m, execution_mode::testing);
+  // Evaluate model on validation set
+  get_trainer().evaluate(&m, execution_mode::tournament);
```
samadejacobs, line 116: One potential problem I see here is that tournament set are typically partitioned between trainers, TSE expect a global dataset for tournament (validation) to rank all trainers.

timmoon10: The user should be responsible for configuring the dataset correctly. Within LBANN, there is no functional difference between the validation, test, and tournament sets. They're just labels to indicate usage. So we should use the validation set for validation, the test set for testing, and the tournament set for tournaments. It's confusing if tournament set isn't used for tournaments.

timmoon10:
As of 12/1, this is ready to merge.

samadejacobs (approved):
LGTM

PR #2002, opened 2021-11-09
Remove deprecated Python scripts for ONNX conversion and plotting

timmoon10:
The ONNX conversion scripts are very old and broken. We have had a user run into problems when trying to use them, so I think it would be safer to remove them entirely. I've also removed some one-off plotting scripts. This also allows us to remove some of our Python dependencies. We now only depend on Protobuf and NumPy (and NumPy isn't strictly required since it's only used in the sub-graph-parallel implementation of Transformers).


timmoon10:
As of 12/1, this is ready to merge.

bvanessen (approved):

PR #2005, opened 2021-11-11
Optimize GPU entry-wise operator functions with contiguous data

timmoon10:
This is an optimization of `apply_entrywise_unary_operator` and `apply_entrywise_binary_operator ` on GPU. In particular, it avoids unnecessary index logic when the data buffers are contiguous. It is intended to address a slowdown when upgrading from CUDA 11.0 to 11.4. @szaman19 reports an 80% speedup in ReLU layers on Pascal, while I don't observe any effect on Lassen. I speculate V100s have optimized integer operations in ways that P100s have not.

[No new Bamboo errors](https://lc.llnl.gov/bamboo/browse/LBANN-TIM430-1).


szaman19:
I'm seeing 80% speedup on Pascal on 11.4 with this update

benson31 (approved):
LGTM

PR #2006, opened 2021-11-11
Fix grid dimensions in GPU entry-wise operator functions

timmoon10:
This is a quick followup bugfix for PR #2005. When running `apply_entrywise_unary_operator`/`apply_entrywise_binary_operator` with non-contiguous GPU data, the CUDA block dimensions are excessively large.


benson31 (approved):
LGTM

PR #2007, opened 2021-11-12
Use Aluminum shuffler in Distconv for ExaGAN model

timmoon10:
The hybrid shuffler in Distconv only supports intra-node communication, so it is unsuitable for applications where a single sample does not fit in a node.


samadejacobs (approved):
LGTM

PR #2008, opened 2021-11-18
Remove a redundant symbol from the layer API

benson31:
```
void Layer::set_num_spliting_groups(El::Int);
void Layer::set_num_spliting_groups(int);
```
These symbols collide if `El::Int == int` (which happens if you don't enable 64-bit integers in Hydrogen).


timmoon10 (approved):
:shipit:

PR #2010, opened 2021-11-19
Bugfix for K-FAC with convolution layers

timmoon10:
We missed a transpose in our K-FAC implementation for convolution layers, causing incorrect results. This PR addresses this error and adds a unit test. This is the same fix as https://github.com/aj-prime/lbann/commit/07c1774bc3fe554ae5cd04d493b5c0d61400678b.

[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM433-1). Pinging @aj-prime.


benson31 (approved):
LGTM

PR #2012, opened 2021-11-22
Explicitly set the search name for Cereal to be all lowercase.

### LINKED ISSUE #2011: cereal isn't found because cmake should be looking for 'cereal', not 'CEREAL'
[Here](https://github.com/LLNL/lbann/blob/develop/CMakeLists.txt#L198).

```cereal-1.3.0.10``` installs these cmake files:
```
$ pkg info -l cereal | grep -i cmake
	/usr/local/lib/cmake/cereal/cerealConfig.cmake
	/usr/local/lib/cmake/cereal/cerealConfigVersion.cmake
	/usr/local/lib/cmake/cereal/cerealTargets.cmake
```
They begin with lowercase ```cereal```.

benson31:
I haven't had issues, but this should be correct. Closes #2011.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,11 +196,13 @@ find_package(Threads REQUIRED)
 find_package(Clara REQUIRED)
 
 find_package(CEREAL NO_MODULE
+  NAMES cereal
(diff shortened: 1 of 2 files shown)


mrwyattii (approved):
LGTM

PR #2014, opened 2021-11-22
Fix tensor dimension issues

benson31:
Addresses some issues @bvanessen was seeing with tensor dimensions.

The initial implementation was very open to unintentional abuse (e.g., if the "sample dims" are pre-padded with 1s and the minibatch size is 1, should another 1 be prepended or no? There's no way to differentiate). This resolves that by making it the user's responsibility to pass the full tensor dimension. As a result, the operator layer now prepends the minibatch dimension manually.


timmoon10 (approved):
LGTM

benson31:
@bvanessen has tested and approved this, but is away from his computer at the moment. Merging on his request.

PR #2015, opened 2021-11-23
Add option for distconv height groups in ExaGAN app

timmoon10:
The current ExaGAN implementation only distributes tensors along the sample and depth dimensions. If we run with many GPUs, then the depth dimension becomes overdecomposed and the halo regions limit scaling. Distributing along the height dimension helps us push the scaling further.


samadejacobs (approved):
approve

PR #2016, opened 2021-11-24
Layer documentation

timmoon10:
Documentation for layers is spotty, sometimes incorrect, and mostly limited to the C++ classes. This PR adds documentation to the Protobuf messages, making it more helpful to users who are using the Python front-end. All layers have been documented (except for deconv, which is documented in PR #1997), all mistakes have been fixed (to my knowledge), and deprecated options are marked as such. It is in a shape where we could get reasonable documentation by parsing `layers.proto` with Doxygen. It is not quite complete since many important layers are actually implemented with operators, but those are mostly straightforward.

[Bamboo is in progress](https://lc.llnl.gov/bamboo/browse/LBANN-TIM435-1).

### DIFF
--- a/bamboo/unit_tests/test_unit_algo_kfac.py
+++ b/bamboo/unit_tests/test_unit_algo_kfac.py
@@ -94,6 +94,7 @@ def construct_trainer(lbann, ):
         print_matrix=False,
         print_matrix_summary=True,
         use_pi=True,
+        compute_interval=1,
     )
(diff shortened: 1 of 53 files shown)


benson31 (approved):
LGTM

bvanessen (approved):
So I am guessing that we still need to write a script to parse the protobuf text into documentation.  Is that correct?

PR #2017, opened 2021-12-01
Added support for the Perlmutter system at NERSC.

bvanessen:
Added support in the build script to allow for center specific BLAS libraries.

Temporarily added the --dirty flag to the spack install to handle how
Spack is dealing with Cray modules.


benson31 (approved):
Seems ok.

PR #1999, opened 2021-11-04
Add RoBERTa base model to applications

mrwyattii:
Adds the RoBERTa base model (based on [HuggingFace implementation](https://huggingface.co/docs/transformers/v4.15.0/en/model_doc/roberta#transformers.RobertaModel)). Also added a script for downloading model configuration and pretrained weights from huggingface. An example of how to run the model with LBANN is provided in `main.py`

Includes changes from #2019

PR #2018, opened 2021-12-07
Add dimension checks to the binary operators

benson31:
Addresses a debugging issue.

While debugging this issue, I also realized we had an issue where `LockedView`s were being used instead of `View`s. This fixes that, too, mostly to have the hassle of a separate PR.


mrwyattii (approved):
LGTM, addresses error in #1999

PR #2019, opened 2021-12-09
Add new PFE functions

mrwyattii:
- Implement commonly used transformations
- Add new activation functions
- Mimic pytorch layers to help users move models from pytorch -> LBANN

### DIFF
--- a/python/lbann/modules/__init__.py
+++ b/python/lbann/modules/__init__.py
@@ -10,3 +10,7 @@ from lbann.modules.base import Module, FullyConnectedModule, ChannelwiseFullyCon
 from lbann.modules.rnn import LSTMCell, GRU, ChannelwiseGRU
 from lbann.modules.transformer import MultiheadAttention
 from lbann.modules.graph import *
+from lbann.modules.subgraph import *
+from lbann.modules.activations import *
+from lbann.modules.pytorch import *
+from lbann.modules.transformations import *
--- a/python/lbann/modules/pytorch.py
+++ b/python/lbann/modules/pytorch.py
@@ -6,11 +6,13 @@ from lbann.utils import str_list
 # Mimics torch.matmul in LBANN
 def PytorchMatmul(x, x_shape, y, y_shape, return_dims=False):
     if len(x_shape) != len(y_shape):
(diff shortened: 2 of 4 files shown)


bvanessen (approved):

python/lbann/modules/__init__.py
```diff
@@ -10,3 +10,7 @@
 from lbann.modules.rnn import LSTMCell, GRU, ChannelwiseGRU
 from lbann.modules.transformer import MultiheadAttention
 from lbann.modules.graph import *
+from lbann.modules.subgraph import *
+from lbann.modules.activations import *
+from lbann.modules.pytorch import *
+from lbann.modules.transformations import *
```
timmoon10, line 16: I prefer being explicit about what we're importing:
```suggestion
from lbann.modules.activations import Gelu, Gelu_approx, Silu
from lbann.modules.pytorch import PytorchMatmul, PytorchLinear, PytorchLayerNorm
from lbann.modules.transformations import Permute, Cumsum
```
We ran into some trouble from blindly importing `lbann.modules.subgraph` (https://github.com/LLNL/lbann/pull/2026).

python/lbann/modules/pytorch.py
```diff
@@ -0,0 +1,69 @@
...
+        z = lbann.Reshape(z, dims=str_list(z_shape))
+
+    if return_dims:
+        return z, z_shape
+    return z
+
+
+# Mimics torch.nn.Linear in LBANN
+def PytorchLinear(x, input_shape, hidden_size, weights=[], name="", return_dims=False):
+    need_reshape = len(input_shape) > 2
```
timmoon10, line 36: I think we need a reshape if the data is 1D, or we should use the basic fully-connected layer.

timmoon10 (changes requested):
I've just copy-pasted my comments from https://github.com/LLNL/lbann/pull/1999.

python/lbann/modules/pytorch.py
mrwyattii: Added a check for 1D inputs with recent commit.

timmoon10 (approved):
:shipit:

PR #2023, opened 2021-12-21
Fix changes that should not have been in #2020

benson31:
I don't know how or why these changes happened? But basically, the trailing return type (whose namespace resolution depends on being AFTER the function name opens the scope) moved to the front of the prototype but did not have namespaces added. I like it the other way better, so I just restored that.


mrwyattii (approved):
LGTM

PR #2025, opened 2022-01-05
Refactor sub-graph parallelism

timmoon10:
This is work to clean up and generalize #1897. 

Tasks:
- [x] Layers initialize distributed buffers on sub-grid
- [x] Weights initialize distributed buffers on sub-grid
- [x] Naive communication pattern between layers on different sub-grids
- [x] Basic option to construct uniform sub-grids
- [x] ~Remove old functionality~ Deferred to a future PR
- [x] Check correctness

Future tasks:
- Restore optimized communication pattern for split, sum, slice, concat layers
- Support running full model on a sub-grid (trickiness: input and eval layers, metrics, objective functions, callbacks)
- K-FAC with secondary grid


timmoon10:
This PR cleans up and generalizes the API to specify which sub-grid to run specific layers on:
1. The first step is to configure the trainers with multiple sub-grids. Grid 0 corresponds to the trainer grid, and the user can add additional sub-grids in whatever order they see fit. I've added the `LBANN_NUM_SUBGRIDS` environment variable, which divides the trainer into equally-sized sub-grids. I envision more options for strided sub-grids (used to minimize inter-node communication) or unequal sub-grids.
2. Similar to the current implementation, you can manually execute a layer on a specific sub-grid by setting its `ParallelStrategy`:
```
y1 = lbann.Cos(x, parallel_strategy={'grid_tag' : 1})
y2 = lbann.Cos(x, parallel_strategy={'grid_tag' : 2})
z = lbann.Sum(y1, y2, parallel_strategy={'grid_tag' : 0})
```
4. If the user does not explicitly set the grid, then the layer will automatically choose one (typically the same grid as its parent layers).
5. This PR just uses basic inter-grid communication using `El::copy::TranslateBetweenGrid`. Optimized communication for the split, sum, slice, and concat layers will be the next step.
6. At the moment, this approach is very sensitive to the layer ordering. It will often serialize the compute on different sub-grids, so it takes some finessing to get it to parallelize properly. I anticipate that things will improve once we have sub-grid-specific layers, at least for compute graphs with basic branching patterns. In the future it may be worthwhile tinkering with the layer graph's topological sort to maximize parallelism between sub-graphs.

timmoon10:
[CI results](https://lc.llnl.gov/gitlab/moon13/lbann/-/pipelines/140711). I see no new failures on Lassen and Pascal. The sub-grid-related fail on Lassen due to segfaults, I suspect because we need to merge https://github.com/LLNL/Elemental/pull/133. I didn't get an allocation on Pascal.

timmoon10:
[No new CI failures on Lassen and Ray, didn't get an allocation on Pascal](https://lc.llnl.gov/gitlab/moon13/lbann/-/pipelines/145640).

include/lbann/trainers/trainer.hpp
```diff
@@ -237,6 +237,12 @@ class trainer
                 El::Int num_batches = 0);
 
   ///@}
+  /** @name Sub-grid management */
+
+  std::vector<El::Grid*> get_grids() const;
+  void add_grid(std::unique_ptr<El::Grid> g);
+
+  ///@{
```
benson31, line 245:
```suggestion
  /** @name Sub-grid management */
  ///@{
  std::vector<El::Grid*> get_grids() const;
  void add_grid(std::unique_ptr<El::Grid> g);
  ///@}
```

benson31 (approved):
LGTM. Seems inline with the discussions we had as a group.

src/layers/data_type_layer.cpp
```diff
@@ -781,28 +812,24 @@ void data_type_layer<InputTensorDataType, OutputTensorDataType>::
 fp_setup_inputs(El::Int mini_batch_size) {
   if (get_num_parents() < 1) { return; }
 
-  // Determine distributed matrix alignment
-  const auto& alignment_dist = get_parent_layer().get_activations(*this).DistData();
```
bvanessen, line 785: Are new no longer ensuring that the matrices have the same alignment?

src/layers/data_type_layer.cpp
```diff
@@ -781,28 +812,24 @@ void data_type_layer<InputTensorDataType, OutputTensorDataType>::
...
     if (!keep_original_inputs(i)) continue;
 #endif // LBANN_HAS_DISTCONV
+
     // Initialize input tensor
     const auto& parent = get_parent_layer(i);
     const auto& parent_output = parent.get_activations(*this);
     auto& input = *m_inputs[i];
     input.Empty(false);
-    if(!this->is_subgraph_parallelism_enabled()){
-      input.AlignWith(alignment_dist);
```
bvanessen, line 798: Is this done somewhere else or skipped now?

bvanessen, line 798: Is this done somewhere else or skipped now?

bvanessen (changes requested):
Should we maintain alignment between matrices on the same grid.

bvanessen (commented):
I have a couple of questions that I want to get addressed before this merges.

src/layers/data_type_layer.cpp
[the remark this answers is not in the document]
timmoon10: Alignment involves overwriting the matrix's `El::Grid`:

https://github.com/LLNL/Elemental/blob/3b7bfb8a8fa6182107fe22a248e1f19c9692692a/src/core/DistMatrix/ElementMatrix.cpp#L252

I believe the only case where matrices have non-default alignment is with the model-parallel slice layer. That case will still be correct with this PR, but it will pessimize by performing communication to restore default alignment.

include/lbann/layers/learning/entrywise_scale_bias.hpp
```diff
@@ -188,54 +174,11 @@ ::setup_data(size_t max_mini_batch_size) {
...
-    err << this->get_type() << " layer \"" << this->get_name() << "\" "
-        << "has misaligned input and weights matrices";
-    LBANN_ERROR(err.str());
-  }
-
-#endif // 0
-
-}
-
-template <typename TensorDataType, data_layout Layout, El::Device Dev>
```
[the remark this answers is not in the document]
timmoon10: The matrix construction has been moved to `setup_data`, which is in this header.

include/lbann/layers/transform/weights.hpp
```diff
@@ -37,57 +36,28 @@ namespace lbann {
...
-    std::vector<int> dims_;
-    for (const auto& d : dims) { dims_.push_back(d); }
-    this->set_output_dims(dims_);
-    this->m_expected_num_parent_layers = 0;
-  }
-
-  weights_layer(const weights_layer& other)
-    : data_type_layer<TensorDataType>(other),
-      m_gradient(other.m_gradient ? other.m_gradient->Copy() : nullptr) {
-    m_workspace.SetMemoryMode(other.m_workspace.MemoryMode());
```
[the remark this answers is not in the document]
timmoon10: We no longer set the memory mode explicitly and rely on the fact that Hydrogen uses memory pools by default.

include/lbann/layers/transform/weights.hpp
```diff
@@ -97,132 +67,27 @@ class weights_layer : public data_type_layer<TensorDataType> {
...
-
-    // Clean up
-    m_workspace.Empty();
-
-  }
-
- private:
-
-  /** Weights gradient. */
-  std::unique_ptr<AbsDistMatrixType> m_gradient;
```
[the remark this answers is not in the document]
timmoon10: Yep, we now construct workspaces as needed in `fp_compute`/`bp_compute` and we accumulate the gradient directly in the optimizer's gradient buffer.

include/lbann/layers/transform/weights.hpp
```diff
@@ -97,132 +67,27 @@ class weights_layer : public data_type_layer<TensorDataType> {
...
-    : weights_layer(nullptr, { 1 } )
-  {}
-
-  void setup_matrices(const El::Grid& grid) override {
-    data_type_layer<TensorDataType>::setup_matrices(grid);
-
-    // Initialize weights gradient
-    auto dist = this->get_activations().DistData();
-    dist.rowDist = El::STAR;
-    m_gradient.reset(AbsDistMatrixType::Instantiate(dist));
```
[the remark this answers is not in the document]
timmoon10: Yep, we accumulate the gradient directly in the optimizer's gradient buffer.

src/models/model.cpp
```diff
@@ -1548,20 +1552,15 @@ void model::setup_layer_execution_order() {
...
+void model::setup_layers(
+  size_t max_mini_batch_size,
+  DataReaderMetaData& dr_metadata,
+  const std::vector<El::Grid*>& grids_) {
 
   for (El::Int i = 0; i < get_num_layers(); ++i) {
     auto& l = get_layer(i);
     l.set_model(this);
-
-    if(this->is_subgraph_parallelism_enabled())
```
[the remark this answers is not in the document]
timmoon10: Good catch, I think that will be the case. The layers will no longer have a way to access the `El::Grid` s constructed in the old sub-graph parallelism infrastructure.

src/layers/data_type_layer.cpp
timmoon10: Yep, non-aligned matrices will now trigger communication.

src/layers/data_type_layer.cpp
timmoon10: Yep, non-aligned matrices will now trigger communication.

timmoon10: Yep, non-aligned matrices will now trigger communication.

bvanessen (approved):

PR #2026, opened 2022-01-06
Make sure sub-graph-specific code is encapsulated in Python submodules

timmoon10:
By importing sub-graph code directly into our base Python submodules (`lbann.modules`, `lbann.models`), we were overwriting the non-sub-graph implementations of multi-head attention and Transformer. Forcing the user to load `lbann.modules.subgraph` or `lbann.models.subgraph` seems appropriate since this is an advanced, experimental feature.

[No new Bamboo failures](https://lc.llnl.gov/bamboo/browse/LBANN-TIM438-1). Pinging @aj-prime and @tnat410.


benson31 (approved):
LGTM

PR #2035, opened 2022-01-25
Bugfix for the checkpoint lenet unit tests

bvanessen:
Updated the checkpoint logic so that when checkpointing during a
validation phase it will use the current epoch and step counts of the
active trainng execution context.  This will allow the validation
checkpoints to be properly ordered between the normal training
checkponts.  Additionally, under the new SGD execution algorithm, the
validation execution context is reset for every validation phase.  As
a result, two helper functions were added to the checkpoint class that
allow the callback to find the current active training execution
context if it was available, otherwise the active execution context is
used.  This ensures that in an inference only run with checkpoint, the
normal behavior should be preserved.

### DIFF
--- a/include/lbann/callbacks/checkpoint.hpp
+++ b/include/lbann/callbacks/checkpoint.hpp
@@ -179,43 +179,6 @@ public:
     }
   }
 
-  /**
-   * Get the current execution context for the model.  However, if the current
-   * execution context is not for training and there is a valid training
-   * execution context, return the training context instead.
-   *
-   */
-  inline const SGDExecutionContext& get_current_execution_context_with_training_override(model* m) {
-    const auto& c = static_cast<SGDExecutionContext&>(m->get_execution_context());
-    if(c.get_execution_mode() != execution_mode::training &&
-       this->get_active_trainer().execution_context_valid(m, execution_mode::training)) {
-      const auto& training_ctx = static_cast<SGDExecutionContext&>(
-        this->get_active_trainer().get_execution_context(m, execution_mode::training));
-      return training_ctx;
-    }
-    return c;
-  }
-
-  /**
-   * When generating a checkpoint for non-training execution phases, the epoch
-   * number should be pulled from the training context to provide a proper
-   * ordering of the checkpoint.
-   */
-  inline size_t get_epoch_with_training_override(model* m) {
-    const auto& c = get_current_execution_context_with_training_override(m);
-    return c.get_epoch();
-  }
-
-  /**
-   * When generating a checkpoint for non-training execution phases, the step
-   * count should be pulled from the training context to provide a proper
-   * ordering of the checkpoint.
-   */
-  inline size_t get_step_with_training_override(model* m) {
-    const auto& c = get_current_execution_context_with_training_override(m);
-    return c.get_step();
-  }
-
   bool need_checkpoint(model *m, callback_phase phase);
   std::string find_latest_checkpoint(lbann_comm& comm,
                                      const std::string& trainer_name,
--- a/include/lbann/callbacks/checkpoint.hpp
+++ b/include/lbann/callbacks/checkpoint.hpp
@@ -179,6 +179,43 @@ public:
     }
   }
 
+  /**
+   * Get the current execution context for the model.  However, if the current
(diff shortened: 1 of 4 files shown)


bvanessen:
CI is running here https://lc.llnl.gov/gitlab/vanessen/lbann/-/pipelines/140297

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -179,6 +179,43 @@ class checkpoint : public callback_base {
...
+  inline const SGDExecutionContext& get_current_execution_context_with_training_override(model* m) {
+    const auto& c = static_cast<SGDExecutionContext&>(m->get_execution_context());
+    if(c.get_execution_mode() != execution_mode::training &&
+       this->get_active_trainer().execution_context_valid(m, execution_mode::training)) {
+      const auto& training_ctx = static_cast<SGDExecutionContext&>(
+        this->get_active_trainer().get_execution_context(m, execution_mode::training));
+      return training_ctx;
+    }
+    return c;
+  }
```
benson31, line 197:
```suggestion
  inline const SGDExecutionContext& get_current_execution_context_with_training_override(model* m) const {
    const auto& c = dynamic_cast<SGDExecutionContext&>(m->get_execution_context());
    if(c.get_execution_mode() != execution_mode::training &&
       this->get_active_trainer().execution_context_valid(m, execution_mode::training)) {
      return dynamic_cast<SGDExecutionContext&>(
        this->get_active_trainer().get_execution_context(m, execution_mode::training));
    }
    return c;
  }
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -179,6 +179,43 @@ class checkpoint : public callback_base {
...
+
+  /**
+   * When generating a checkpoint for non-training execution phases, the epoch
+   * number should be pulled from the training context to provide a proper
+   * ordering of the checkpoint.
+   */
+  inline size_t get_epoch_with_training_override(model* m) {
```
benson31, line 204:
```suggestion
  inline size_t get_epoch_with_training_override(model* m) const {
```

include/lbann/callbacks/checkpoint.hpp
```diff
@@ -179,6 +179,43 @@ class checkpoint : public callback_base {
...
+    const auto& c = get_current_execution_context_with_training_override(m);
+    return c.get_epoch();
+  }
+
+  /**
+   * When generating a checkpoint for non-training execution phases, the step
+   * count should be pulled from the training context to provide a proper
+   * ordering of the checkpoint.
+   */
+  inline size_t get_step_with_training_override(model* m) {
```
benson31, line 214:
```suggestion
  inline size_t get_step_with_training_override(model* m) const {
```

benson31 (approved):
Overall seems fine. If CI's happy, I'm happy.

benson31:
Actually, these three functions should probably be `private:` too. But so should most of this interface and it isn't so we can clean it all up later (and hopefully promote checkpointing to a first-class feature and thus get rid of the callback).

src/trainers/trainer.cpp
```diff
@@ -191,6 +191,19 @@ trainer::get_execution_context(execution_context_key_pair_t key)
     *(m_model_execution_context[key].get()));
 }
 
+bool trainer::execution_context_valid(observer_ptr<model> model,
+                                      execution_mode mode)
+{
+  auto key = std::make_pair(model, mode);
+  return execution_context_valid(key);
```
benson31, line 198:
```suggestion
  return execution_context_valid(std::make_pair(model, mode));
```
Avoid the copy.

benson31 (changes requested):
Upon further reflection, I think I feel strongly that the definitions added to `checkpoint.hpp` should be moved to the `.cpp`. I'm not _thrilled_ that the name `SGDExecutionContext` is being added to this header file, but that can be hidden with a forward declaration rather than an `#include` as long as it's just used by reference.

It would be better if these functions were written as free functions instead and put in an anonymous namespace in `checkpoint.cpp`.

benson31 (approved):
lGTM

PR #2037, opened 2022-01-26
Update all copyright statements to the current year

benson31:
There seems to have been a variety of copyright statements; this cleans that up. It also adds it to some source files that were missing it.

I haven't addressed any of the Python code. That should probably be done, too.


timmoon10 (approved):
:shipit:

PR #2039, opened 2022-01-27
Add RoBERTa Integration Test

mrwyattii:
- Fixes a problem with sizing attention masks correctly when an attention mask is provided to the model
- Removes unneeded weights layer initializer from original RoBERTa impl.
- Adds an integration test for the RoBERTa model
- RoBERTa can now be imported from `lbann.models` in PFE


timmoon10 (approved):
LGTM. Since we now have a module for RoBERTa in `lbann.models`, we can remove it from the applications directory.

bvanessen (approved):

PR #2040, opened 2022-01-27
Correct the copy semantics of the argument parser

benson31:
When copying the argument parser, the reference held by Clara's `ExeName` was not updated to the new object, thereby creating a dangling reference to the `exe_name_` slot in the "source" argument parser.

This is not only true of the executable name; it would happen with any existing option. Therefore, the only safe argument parser to copy is an empty argument parser. Since this isn't very useful, I have removed the copy constructor/assignment interfaces from the class.

This was discovered in an attempt to clear the argument parser of existing state by just copy-assigning a default-constructed parser to it. To add clarity, I've just added a `clear()` function to the parser that resets it to its default-constructed state (clearing the exe name, as well as any arguments/flags/options that have been added).

### DIFF
--- a/include/lbann/utils/argument_parser.hpp
+++ b/include/lbann/utils/argument_parser.hpp
@@ -243,6 +243,29 @@ public:
   /** @brief Create the parser */
   argument_parser();
 
+  /** @brief Copy construction is disabled.
+   *
+   *  The Clara parser is not actually copyable because the references
+   *  are bound to memory in the parameter map. They are not re-bound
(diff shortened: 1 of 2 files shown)


benson31:
[GitLab](https://lc.llnl.gov/gitlab/benson31/tomslbannfork/-/pipelines/140974).

bvanessen (approved):

bvanessen:
@benson31 This fixed the bugs in my PR.

PR #2041, opened 2022-01-28
Python function to perform naive hyperparameter sweep

timmoon10:
Given a function to configure an LBANN experiment and lists of hyperparameter values, `lbann.contrib.hyperparameter.grid_search` will exhaustively generate all valid configurations and train them with LBANN. In particular, it will use multi-trainer training, possibly in multiple rounds.

For example, suppose you have implemented:
```
def make_lenet(activation, loss, optimizer, learning_rate):
    # <implementation>
    return model, optimizer, data_reader, trainer
```
Then you could do a hyperparameter sweep with:
```
lbann.contrib.hyperparameter.grid_search(
    script,
    make_lenet,
    activation=['relu', 'sigmoid', 'tanh'],
    loss=['cross_entropy', 'mse'],
    optimizer=['sgd', 'adam'],
    learning_rate=[0.1, 0.01, 0.001],
)
```
The log files can be post-processed to find the most promising configurations.

### DIFF
--- a/docs/hyperparameter_tuning.rst
+++ b/docs/hyperparameter_tuning.rst
@@ -1,0 +1,105 @@
+.. role:: python(code)
+          :language: python
+
+.. _hyperparameter_tuning:
+
+============================================================
+Hyperparameter Tuning
+============================================================
+
+.. note:: See the docstring for
+          `lbann.contrib.hyperparameter.grid_search` for the API. This
+          page provides additional guidance on usage.
+
+.. note:: This feature is experimental and its API is still being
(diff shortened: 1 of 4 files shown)


samadejacobs (approved):
LGTM, might need to expand the documentation, may be copy 'conversation' to readthedocs, and expand a bit, say how to replace default training algo with LTFB training (which I believe is a good candidate/consumer for this). This simple but nice approach worth extending to random search as well, not necessarily for this PR.

timmoon10:
I've added a Read the Docs page with some guidance on usage and a simple example.

PR #2042, opened 2022-02-01
Cleanup of the month

benson31:
Just cleaning up some warnings.


timmoon10 (approved):
This is fine, but I'd prefer to revert the changes to `in_top_k.cu` and `tessellate.cu`. The arguments are unused, but make it easier to understand.

benson31:
I don't necessarily agree -- the changes in question are in anonymous-namespace functions that aren't exported to the public. This could be solved equivalently by object orientation (`dims3` or a new `TensorDims` object instead of manually unpacking the dimensions in calling code) or by documenting why only the width or dims>0 are needed. But my goal is no warnings, not elegance nor improved developer documentation, so the changes have been reverted with the formal argument names commented out to silence the warning.

bvanessen (approved):

PR #2043, opened 2022-02-02
Update the utilities for C++17

benson31:
Since LBANN has moved to requiring C++17, these could be cleaned up a bit. For now, we just squash `lbann::utils::any` and `lbann::make_unique`.


timmoon10 (approved):
:shipit:

bvanessen (approved):

PR #2045, opened 2022-02-04
Remove RoBERTa impl from applications directory

mrwyattii:
In #2039 an implementation of RoBERTa was added to `lbann.models`, making the implementation in `applications/nlp/` redundant.

Also modifies the path for pre-trained weights to work with new implementation of the model.


timmoon10 (approved):
:shipit:

PR #2046, opened 2022-02-07
Remove the auto-print "feature" of the LBANN exception ctor.

benson31:
I also took the liberty to clean up a few things in the exception class. In particular, there's no reason to store the stack trace in its own string; it is only used in the thing that gets printed by `print_report()`.

The biggest change is that the `what()` string now just stores the whole report message, and `print_report()` just writes the `what()` string to the given stream. The visible behavior should be mostly the same for users -- `lbann.cpp` for instance only calls `print_report()` to a file stream if the user has requested that stack traces dump to files. `El::ReportException()` will print the `what()` string, which now includes the stack trace. But this is the behavior we want anyway since the stack trace is no longer reported when the exception is constructed.

I also did a little cleanup in `stack_trace.cpp`, where I noticed that `lbann::exception::print_report()` was being used. I only wish I could do more.

### DIFF
--- a/include/lbann/utils/exception.hpp
+++ b/include/lbann/utils/exception.hpp
@@ -37,8 +37,8 @@
 #define LBANN_ERROR(...)                                        \
   do {                                                          \
     const int rank_LBANN_ERROR = lbann::get_rank_in_world();    \
-    throw lbann::exception(                                     \
-      lbann::build_string(                                      \
+    throw ::lbann::exception(                                   \
+      ::lbann::build_string(                                    \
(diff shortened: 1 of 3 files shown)


benson31:
I also cleaned up the docs a bit.

Since CI is a bit wonked right now, I have built locally on Pascal and I ran both the Catch2 and the `ci_test` tests manually -- no (unexpected) issues.

timmoon10 (approved):
LGTM

bvanessen (approved):

PR #2052, opened 2022-02-09
Silence warning message in Catch test for Python wrapper

### LINKED ISSUE #2048: (LOW priority) Catch2 Python unit test output
It'd be ideal if the Python unit tests in Catch2 (`src/utils/unit_test/python_test.cpp`) produced no output on success.

Actual:
```
> ./unit_test/seq-catch-tests "[python]" (cereal-dynamic-init=)[1]
Filters: [python]
  File "<string>", line 3
    this should throw a NameError
              ^
SyntaxError: invalid syntax
===============================================================================
All tests passed (111 assertions in 1 test case)
```

Expected:
```
> ./unit_test/seq-catch-tests "[python]" (cereal-dynamic-init=)[1]
Filters: [python]
===============================================================================
All tests passed (111 assertions in 1 test case)
```

This isn't wrong, per se, but the current output produces somewhat confusing output. It says there was a syntax error, but then it says everything passed. It'd be good if this output weren't generated when running the tests.

timmoon10:
When running the Catch tests, the output is cluttered by a warning from the Python wrapper (see Issue #2048). It turns out Python will print to `stderr` whenever it encounters a syntax error. Fortunately, we can temporarily replace `sys.stderr` so that it prints to `/dev/null` instead. The Catch tests pass for me on Lassen and I no longer see the warning message.

Closes #2048.

### DIFF
--- a/src/utils/unit_test/python_test.cpp
+++ b/src/utils/unit_test/python_test.cpp
@@ -58,13 +58,25 @@ def throw_exception():
 
     SECTION ("Making syntax error") {
       PyObject* main = PyImport_ImportModule("__main__");
-      std::string func_def = R"(
-def make_syntax_error():
-    this should throw a NameError
+      std::string silence_warnings = R"(
+import os, sys
+stderr_orig = sys.stderr
+stderr_devnull = open(os.devnull, 'w')
+sys.stderr = stderr_devnull
 )";
+      std::string func_def = R"(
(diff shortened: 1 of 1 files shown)


benson31 (approved):
I'm sad about how many lines it takes to do this, but I'm glad to see that output gone. Thanks very much!

LGTM

PR #2053, opened 2022-02-09
Cleanup of the SMILES test file.

benson31:
There's currently a sporadic segfault showing up in our CI in this test. However, I have no idea why and it's not reproducible outside of CI. This is some cleanup to make this test more accessible and to ease debugging. If it happens to resolve the segfault in the process, great!


benson31:
This should be good to go. I'm still not seeing a segfault locally, but it looks like #2054 should fix that.

bvanessen (approved):

PR #2055, opened 2022-02-09
Update Transformer app with recent API changes

timmoon10:
The "dump weights" callback changed the way it names files, so this PR update the evaluation script so that we can import an LBANN-trained Transformer into PyTorch.

Pinging @tnat410.


samadejacobs (approved):
LGTM

PR #2060, opened 2022-02-10
Cleanup data reader tests memory leaks

bvanessen:
Switched data reader tests from using raw pointers to unique pointers for instantiating the data reader.

### DIFF
--- a/src/data_readers/unit_test/data_reader_HDF5_hrrl_public_api.cpp
+++ b/src/data_readers/unit_test/data_reader_HDF5_hrrl_public_api.cpp
@@ -152,7 +152,7 @@ TEST_CASE("hdf5 data reader data field fetch tests",
   conduit::Node ref_node;
   ref_node.parse(hdf5_hrrl_data_sample_id, "yaml");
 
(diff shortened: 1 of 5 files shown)


src/data_readers/unit_test/data_reader_HDF5_hrrl_public_api.cpp
```diff
@@ -168,8 +168,8 @@ TEST_CASE("hdf5 data reader data field fetch tests",
 
   El::Int num_samples = 1;
 
-  auto data_store = new lbann::data_store_conduit(hdf5_dr);
-  hdf5_dr->set_data_store(data_store);
+  auto data_store = std::make_unique<lbann::data_store_conduit>(hdf5_dr.get());
+  hdf5_dr->set_data_store(data_store.get());
```
benson31, line 172: This is tricky. The data reader is actually taking ownership (not that that's documented anywhere):
```suggestion
  {
    auto data_store = std::make_unique<lbann::data_store_conduit>(hdf5_dr.get());
    hdf5_dr->set_data_store(data_store.release());
  }
```

benson31 (approved):

PR #2063, opened 2022-02-15
Improved ATOM / SMILES tests

bvanessen:
Refactored the ATOM application code so that the model definition is
in a separate file so that it is easier to add an explicit test of
that exact model.

Added an integration test for the ATOM application model.

Changed the SMILES data reader so that the metadata file is specified
via a metadata_file field rather than piggybacking on the label_file,
which was causing confusion.

Enabled Ray to run the SMILES / ATOM integration tests.

Removed a number of deprecated command line arguments: no_header,
num_train_samples, num_test_samples, num_validate_samples


ci_test/integration_tests/test_integration_atom_wae.py
```diff
@@ -25,15 +25,16 @@
 num_decoder_layers = 3
 
 # Reconstruction loss
-expected_train_recon_range = (0.500, 0.525)
+#expected_train_recon_range = (0.500, 0.525)
```
benson31, line 28:
```suggestion
```
If they aren't used, just get rid of them? If these are left without a comment, nobody will remember why they're there and commented out.

include/lbann/data_readers/data_reader_smiles.hpp
```diff
@@ -119,6 +119,16 @@ class smiles_data_reader : public data_reader_sample_list<sample_list_ifstream<l
   void load_list_of_samples(const std::string sample_list_file);
 
 
+  /** @brief Sets the name of the metadata file */
+  void set_metadata_filename(std::string fn) { m_metadata_filename = fn; }
```
benson31, line 123:
```suggestion
  void set_metadata_filename(std::string fn) { m_metadata_filename = std::move(fn); }
```

benson31 (approved):
LGTM

PR #2065, opened 2022-02-15
Remove the Cloneable interface from the training algorithms.

benson31:
There's no practical reason to clone a training algorithm. Namely, *users* will never clone a training algorithm as there's no way to implement that operation in the front-end. Moreover, it complicates semantics. Where training algorithms are given unique identifiers, which is valuable for user-friendly output, how should that be propagated to the clone? If there are two instances of the algorithm with the same name, it is no longer unique.

Additionally, it's telling that I changed literally no other code outside the class definitions (and the SGD test)...


benson31:
[Pipeline is Green](https://lc.llnl.gov/gitlab/benson31/tomslbannfork/-/pipelines/146607).

timmoon10 (approved):
Seems reasonable to me.

PR #2066, opened 2022-02-16
Add include guard for dist_embedding layer cerealization 

szaman19:
To stop instantiation of both CPU and GPU layers if only one of LBANN_HAS_SHMEM or LBANN_HAS_NVSHMEM is defined. Otherwise, there is a linker error when building with +distconv +nvshmem via Spack


timmoon10:
I've copy-pasted the Cereal registration logic from the "DFT abs" layer. This should help avoid registering multiple times when both OpenSHMEM and NVSHMEM are detected.

benson31 (approved):
LGTM

PR #2068, opened 2022-02-17
Update ATOM WAE app driver to use LTFB training algorithm

timmoon10:
This replaces the LTFB callback with the LTFB training algorithm. Some implications:
- When training with LTFB, it is no longer straightforward to just specify a number of training epochs. The user must specify the tournament interval and the number of tournament steps. If the user wants to train for a certain number of epochs, they must manually figure out the dataset size and set the right number of LTFB tournaments.
- I've switched to "checkpoint binary" communication in LTFB. This means that it is no longer meaningful to communicate a subset of weights.


samadejacobs (approved):
LGTM

PR #2069, opened 2022-02-18
Training algorithm timers

benson31:
The main thing that hasn't been done is making the display optional.


timmoon10 (approved):
Overall looks good. Making the output optional would be useful in cases where timing isn't so important, like the CI tests.

I wonder about how we can make the format more configurable. We have a lot of reporting for the callbacks, but in practice it doesn't have a big impact in most experiments. Perhaps the default should only report the forward and backward prop times (or maybe just `TrainingAlgorithm::apply`), and then we can add options to report the time spent in callbacks, optimizers, etc.

samadejacobs (approved):
Ditto @timmoon10's comments

bvanessen (approved):

PR #2072, opened 2022-03-02
Fix Cereal error for distributed embedding

szaman19:
PR #2066 didn't solve the compilation error with nvshmem and distconv enabled. The dist_embedding layer required the layout template parameter. This fix just instantiates with lbann::data_layout::DATA_PARALLEL 

Pinging @timmoon10


timmoon10 (approved):
:shipit:

PR #2073, opened 2022-03-07
Avoid constructing unnecessary weights in Python modules

timmoon10:
We always create two weights in the convolution Python module: one for the kernel and one for the biases. This PR adds logic so that we only create the kernel if the bias is disabled. A similar change is made to the fully-connected module and the channel-wise fully-connected module was already using this approach.

[CI is in progress](https://lc.llnl.gov/gitlab/moon13/lbann/-/pipelines/152013).

### DIFF
--- a/python/lbann/modules/base.py
+++ b/python/lbann/modules/base.py
@@ -97,7 +97,7 @@ class FullyConnectedModule(Module):
             self.weights.append(
                 lbann.Weights(initializer=lbann.HeNormalInitializer(),
                               name=self.name+'_matrix'))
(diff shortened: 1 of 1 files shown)


bvanessen (approved):
LGTM if it works for @samadejacobs

benson31 (approved):
LGTM, pending CI and @samadejacobs

samadejacobs (approved):
WFM

timmoon10:
No new CI failures on Lassen, Pascal, and Ray. Catalyst failed to build, which suggests an issue in `develop`.

PR #2079, opened 2022-03-17
Bump CUDA to version 11 and push the C++17 standard to CUDA

benson31:
Pinging @jonesholger.


bvanessen (approved):
LGTM.  I haven't tested this but if it passes CI, it is good.

PR #2081, opened 2022-03-18
Enable on demand (nightly) integration tests

bvanessen:
Updating the CI testing framework so that the weekly flag is passed
into the setup_experiment function so that tests, especially
integration, tests can be run in both a quick mode (i.e. nightly) and
a more extensive tests (i.e. weekly).

Updated the ATOM integration test to use both nightly and weekly modes.

### DIFF
--- a/ci_test/common_python/tools.py
+++ b/ci_test/common_python/tools.py
@@ -679,7 +679,7 @@ def create_tests(setup_func,
         # Make sure test name is prefixed with 'test_'
         test_name_base = 'test_' + test_name_base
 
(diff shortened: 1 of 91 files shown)


benson31 (approved):
Looks good

benson31 (commented):
I'm wondering if the setup function should instead take the LBANN module being tested and the PyTest config object, instead of just the `weekly` flag. That, coupled with the "features" capability that @graham63 is working on, would give us access to a lot of information about _what_ we're testing (the `lbann` module) and _how_ we're testing (the PyTest config).

PR #2084, opened 2022-03-22
Fix Catalyst CI build

bvanessen:
Switched the Catalyst build back to using GCC to avoid some weird Clang issues with Spack and Python packges.


benson31 (approved):
I don't understand why these changes worked for you when a nearly identical set failed for me, but 🤷‍♂️.

PR #2087, opened 2022-03-24
Move data store exchange to primary thread

bvanessen:
Moved the data store exchange out of the fetch_data call to allow it
to start and finish in the primary thread.  This should alleviate
issues with MPI_THREAD_MULTIPLE and background I/O.  Furthermore,
there should be no loss in performance since it is being executed in
the same sequence as before.  Additionally, split the data store calls
into two phases that are split between initiating the non-blocking
sends and receives and the subsequent wait statements.  Currently,
these calls are still executed back-to-back to provide equivalent
semantics as the prior implementation.

### DIFF
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -342,7 +342,6 @@ class generic_data_reader {
   void start_data_store_mini_batch_exchange();
   void finish_data_store_mini_batch_exchange();
 
-
   /**
    * During the network's update phase, the data reader will
    * advanced the current position pointer.  If the pointer wraps
--- a/include/lbann/data_readers/data_reader.hpp
+++ b/include/lbann/data_readers/data_reader.hpp
@@ -339,6 +339,10 @@ class generic_data_reader {
     m_supported_input_types[INPUT_DATA_TYPE_RESPONSES] = b;
   }
 
+  void start_data_store_mini_batch_exchange();
+  void finish_data_store_mini_batch_exchange();
+
+
   /**
    * During the network's update phase, the data reader will
    * advanced the current position pointer.  If the pointer wraps
--- a/src/data_readers/data_reader.cpp
+++ b/src/data_readers/data_reader.cpp
@@ -205,6 +205,29 @@ int lbann::generic_data_reader::fetch(
   return mb_size;
 }
 
+void lbann::generic_data_reader::start_data_store_mini_batch_exchange() {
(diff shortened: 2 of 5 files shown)


bvanessen:
https://lc.llnl.gov/gitlab/vanessen/lbann/-/pipelines/157560

src/data_readers/data_reader.cpp
```diff
@@ -205,6 +205,28 @@ int lbann::generic_data_reader::fetch(
   return mb_size;
 }
 
+void lbann::generic_data_reader::start_data_store_mini_batch_exchange() {
+  int loaded_batch_size = get_loaded_mini_batch_size();
+
+  /// Make sure that every rank participates in the data store prior
+  /// to seeing if the local rank's position is valid.  Note that
+  /// every rank will hold data that may be used in the last mini-batch
```
benson31, line 213:
```suggestion
  // Make sure that every rank participates in the data store prior
  // to seeing if the local rank's position is valid.  Note that
  // every rank will hold data that may be used in the last mini-batch
```

src/data_readers/data_reader.cpp
```diff
@@ -205,6 +205,28 @@ int lbann::generic_data_reader::fetch(
...
+  if (data_store_active()) {
+    m_data_store->start_exchange_mini_batch_data(m_current_pos-m_base_offset-m_model_offset, loaded_batch_size);
+  }
+  return;
+}
+
+void lbann::generic_data_reader::finish_data_store_mini_batch_exchange() {
+  /// Make sure that every rank participates in the data store prior
+  /// to seeing if the local rank's position is valid.  Note that
+  /// every rank will hold data that may be used in the last mini-batch
```
benson31, line 223:
```suggestion
  // Make sure that every rank participates in the data store prior
  // to seeing if the local rank's position is valid.  Note that
  // every rank will hold data that may be used in the last mini-batch
```

include/lbann/data_readers/data_reader.hpp
```diff
@@ -339,6 +339,10 @@ class generic_data_reader {
     m_supported_input_types[INPUT_DATA_TYPE_RESPONSES] = b;
   }
 
+  void start_data_store_mini_batch_exchange();
+  void finish_data_store_mini_batch_exchange();
+
+
```
benson31, line 345:
```suggestion
  void start_data_store_mini_batch_exchange();
  void finish_data_store_mini_batch_exchange();

```

benson31 (approved):
Just some style tweaks.

PR #2088, opened 2022-03-28
Reduce LTFB startup overhead and reporting

bvanessen:
Removed three global all gather collectives from LBANN's start up
unless LTFB global statistics and verbose reporting are on.  Also
limited the number of LTFB partner pairings that are reported unless
LTFB verbose reporting is enabled.


benson31 (approved):
LGTM.

PR #2089, opened 2022-03-28
Remove the LTFB callback

benson31:
The LTFB training algorithm has been featured for a while now, and projects have adopted it. The training algorithm is the driving force of new development. The callback is now just baggage, and when bugs do pop up in LTFB, it's nice to just fix them once.


benson31:
This is currently on top of #2088 to avoid the conflict of deleting a file that that PR changed. When that merges, I'll rebase this.

bvanessen (changes requested):
We need to review the use cases that are not supported in the new execution algorithm and if there are sufficient replacement test cases.

ndryden (approved):
It's been 18 months, I think we can merge this.

PR #2104, opened 2022-04-27
Bugfix in Spack for compiler specification

bvanessen:
Separated out the spack compiler flag from the rest of the variants
and dependent packages.  This ensures that the compiler is propagated
to the top level LBANN package as well as any co-concretized
packages. Pulled out the high speed linker into a conditional block to
make sure it is only set if the compiler is compatible.  Added the
option to override the compiler from the command line.

Added a spack variant of the host architecture to make sure that spack
builds for the intended system and not some other microarchitecture.

### DIFF
--- a/scripts/build_lbann.sh
+++ b/scripts/build_lbann.sh
@@ -498,7 +498,6 @@ if [[ "${CENTER_COMPILER}" =~ .*"%gcc".* ]]; then
     CENTER_LINKER_FLAGS="+gold"
 fi
 
-
 GPU_VARIANTS_ARRAY=('+cuda' '+rocm')
 DEPENDENT_PACKAGES_GPU_VARIANTS=
 for GPU_VARIANTS in ${GPU_VARIANTS_ARRAY[@]}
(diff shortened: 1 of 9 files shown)


JaeseungYeom (approved):
This PR fixed the issue of compiler choice. However, the spack concretization target arch is still not accurate for me. I suppose this is a spack issue.

scripts/build_lbann.sh
```diff
@@ -624,6 +643,12 @@ if [[ -n "${INSTALL_DEPS:-}" ]]; then
     echo ${CMD} | tee -a ${LOG}
     [[ -z "${DRY_RUN:-}" ]] && { ${CMD} || exit_on_failure "${CMD}"; }
 
+    # Set the environment to vaoid concretizing for microarchitectures that are
```
benson31, line 646:
```suggestion
    # Set the environment to avoid concretizing for microarchitectures that are
```

benson31 (approved):
This looks fine. I don't like the large blocks of commented-out code, but I assume that's just preserving a thought so it's fine.

PR #2109, opened 2022-05-12
NASNet

samadejacobs:
LBANN implementation of NASNet search space with random (baseline) and LTFB search strategies. This would be the first in the series of extending LBANN for neural architecture search. Additional (future) search space would be a directory within the upper level nas directory.


bvanessen (approved):
@samadejacobs Can we get a small integration test associated with this.

samadejacobs:
> @samadejacobs Can we get a small integration test associated with this.

@bvanessen , as discussed, integration test added.

PR #2110, opened 2022-05-16
Added additional includes to enable builds on OS X.

bvanessen:


benson31 (approved):
This is fine and gets people building on OSX.

Some additional background on the issue: The source of the problem seems to be that implicitly instantiated templates get added to the library with type `t` (i.e., a local symbol), at least on Darwin. However, recent changes to LBANN have started calling the `El::copy::Translate` interface directly -- it _should_ be equivalent to just calling `El::Copy` but without all the indirection to finally dispatch to `El::copy::Translate`. Now, though, the linker won't use the local-to-libHydrogen symbol, so link error! (A symmetric analysis applies to `El::copy::TranslateBetweenGrids`.)

However, the reason I didn't PR this in the first place is that it could be unstable in the sense that _technically_ Hydrogen is separate from LBANN and this is a private interface. (I generally accept the public interface to be the top-level `El::` namespace, though we haven't formally declared it to be so.) There are three possible fixes:

1. Just do what this PR does, include the right Hydrogen files (they're installed and available, after all), let LBANN reinstantiate the templates it needs, and move on.
2. Deal with the cost of the indirection of `El::Copy` and use that undeniably public function instead. This will incur some performance hit, but it's not clear to me that it's significant in the context of a tensor copy or serialization.
3. Add explicit instantiation for the `El::copy::` functions so they become global symbols. I like this idea the least because, again, I view these as implementation details rather than first-class interfaces.

PR #2112, opened 2022-05-25
Fix typo on distconv channelwise FC

szaman19:
Minor typo introduced during clean up but will cause failure for distconv-enabled builds. 

Pinging @benson31


benson31 (approved):
Oh come on -- it couldn't figure _that_ out?? They should add autocorrect to compilers... :/ (Mostly because, if it performs similarly to autocorrect for text messages, the resulting code could be hilarious.)

PR #2115, opened 2022-06-20
Add a permute layer utilizing cuTENSOR for the permute implementation.

benson31:
This is an implementation of Permute using cuTENSOR for CUDA GPUs. It's meant to be relatively modular, so other backends could be used in the future.

To access this, cuTENSOR must be found by CMake (using standard means). If it's found, it will be used. If it's not found, the permute-via-gather implementation will be used. For those not using the [LBANN transformations module](https://github.com/LLNL/lbann/blob/develop/python/lbann/modules/transformations.py), the layer name in the Python front-end is `TensorPermute`. If this layer is used and cuTENSOR has not been found, there will be a C++ runtime error.

I observe a 10-15% speedup (vs permute-via-gather) in **total** runtime of Roberta on v100s, and 20-25% on p100s.

The tests pass locally; running CI now.

TODO:

- [ ] Cereal serialization of the permute layer.


include/lbann/layers/transform/permute.hpp
```diff
@@ -0,0 +1,107 @@
...
+  void setup_dims(DataReaderMetaData& dr_metadata) final;
+  void fp_compute() final;
+  void bp_compute() final;
+
+private:
+  class PermuteImpl;
+  std::unique_ptr<PermuteImpl> m_impl;
+};
+
+#ifdef LBANN_HAS_CUDA
```
bvanessen, line 93: Should this be `LBANN_HAS_CUDA` or `LBANN_HAS_CUTENSOR`

bvanessen (approved):
Passes local tests.  LGTM

include/lbann/layers/transform/permute.hpp
benson31: CUTENSOR. Changed. Thanks.

PR #2116, opened 2022-06-20
Cleanup Convolution and Deconvolution Protobuf messages

benson31:
This removes the deprecated fields from Deconvolution, and changes the Convolution message to better align with both Deconvolution and PyTorch's nomenclature.


samadejacobs (approved):
LGTM, confirmed with yet to be checked in (WIP) exagan models.

bvanessen (approved):

PR #2118, opened 2022-06-22
Nersc multi gan

samadejacobs:
LBANN implementation of multi generative model (multi GAN) for 3D NERSC Cosmology application. 
Subgraph parallelism of multi discriminators network is optional (can be set by enable_subgraph flag)
This PR depends on [PR2116](https://github.com/LLNL/lbann/pull/2116)


bvanessen (approved):
The code looks good. Note that the conditional GAN training is not tuned for hyperparameters.  This work also could take better use of sub-grid parallelism, but does not at the moment.

PR #2123, opened 2022-06-27
Allow Catch2 version 3 to be used in addition to version 2.

benson31:
This will hold us over for a transition period. I'd like to get to Catch2 v3 in general, though, as it's better factored. (I can do that now, but I know @bvanessen just put work in to make sure we only use Catch2 v2...)


bvanessen (approved):
LGTM.  Haven't tested yet.

benson31:
CI shows a build problem that I thought I had fixed. I might have an uncommitted fix lying around. I'll fix this and ask for a follow-up review when all's good with CI.

PR #2128, opened 2022-07-19
Add MIOpen profiler annotations

benson31:
Adding profiling regions to MIOpen interface.

MIOpen either isn't annotated or I can't figure out how to turn them on. This adds profiling regions to functions that directly call MIOpen functions to provide this insight.

### DIFF
--- a/include/lbann/utils/dnn_lib/miopen/convolution.hpp
+++ b/include/lbann/utils/dnn_lib/miopen/convolution.hpp
@@ -29,6 +29,7 @@
 #include "lbann/utils/dnn_enums.hpp"
 #include "lbann/utils/dnn_lib/helpers.hpp"
 #include "lbann/utils/gpu/helpers.hpp"
+#include "lbann/utils/profiling.hpp"
 
 #include "utils.hpp"
 
@@ -48,6 +49,7 @@ get_fwd_conv_workspace_size(TensorDescriptor const& wDesc,
                             TensorDescriptor const& yDesc,
                             El::SyncInfo<El::Device::GPU> const& si)
 {
+  BASIC_PROF_REGION("miopen:get_fwd_conv_workspace_size");
   size_t size;
   auto handle_manager = internal::make_default_handle_manager(si);
   CHECK_MIOPEN(miopenConvolutionForwardGetWorkSpaceSize(handle_manager.get(),
@@ -66,6 +68,7 @@ get_bwd_data_conv_workspace_size(TensorDescriptor const& dyDesc,
                                  TensorDescriptor const& dxDesc,
                                  El::SyncInfo<El::Device::GPU> const& si)
 {
+  BASIC_PROF_REGION("miopen:get_bwd_data_conv_workspace_size");
   size_t size;
   auto handle_manager = internal::make_default_handle_manager(si);
   CHECK_MIOPEN(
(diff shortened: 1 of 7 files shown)


include/lbann/utils/dnn_lib/miopen/dropout.hpp
```diff
@@ -43,13 +44,15 @@ using namespace miopen;
 
 inline size_t get_dropout_states_size()
 {
+  BASIC_PROF_REGION("miopen:get_dropout_states_size");
   size_t size;
   CHECK_MIOPEN(miopenDropoutGetStatesSize(get_handle(), &size));
```
bvanessen, line 49: @benson31 I am wondering if this type of call on the dropout also has a similar cost to the get convolution workspace size call.

include/lbann/utils/dnn_lib/miopen/local_response_normalization.hpp
```diff
@@ -43,6 +44,7 @@ using namespace miopen;
 
 inline size_t get_lrn_ws_size(TensorDescriptor const& yDesc)
 {
+  BASIC_PROF_REGION("miopen:get_lrn_ws_size");
   size_t size;
   CHECK_MIOPEN(miopenLRNGetWorkSpaceSize(yDesc,
```
bvanessen, line 49: Same thought here.  Are these calls performance killers.

include/lbann/utils/dnn_lib/miopen/convolution.hpp
```diff
@@ -48,15 +49,7 @@ get_fwd_conv_workspace_size(TensorDescriptor const& wDesc,
...
-  size_t size;
-  auto handle_manager = internal::make_default_handle_manager(si);
-  CHECK_MIOPEN(miopenConvolutionForwardGetWorkSpaceSize(handle_manager.get(),
-                                                        wDesc,
-                                                        xDesc,
-                                                        convDesc,
-                                                        yDesc,
-                                                        &size));
-  return size;
+  return size_t{1 << 30};
```
bvanessen, line 52: So at the moment this is a performance killer that may only offer a small benefit in memory savings.  I am wondering if we should make this a runtime configuration option so that we can test to see if the MIOPEN stack fixes this bug in the future without code change on our part.  So maybe we add a runtime flag (LBANN_USE_MIOPEN_DYNAMIC_WORKSPACE_CALCULATION) or something like that in a conditional guard.

bvanessen: That way we can easily test this in the future.

bvanessen (changes requested):
Overall looks good to me.  Wondering if we should make some small tweaks for future testing of MIOPEN.

include/lbann/utils/dnn_lib/miopen/convolution.hpp
benson31: As I think we've discussed offline, this implementation matches the cuDNN implementation. If the device memory allocator is caching blocks (e.g., CUB), the workspace will be reused in subsequent layers anyway. If it's not caching, this will be the least of our performance concerns.

include/lbann/utils/dnn_lib/miopen/dropout.hpp
benson31: I think it's worth investigating deeper, but not in this PR. The main focus here was adding the annotations. Adding these annotations highlighted the severe cost of `miopenConvolution*GetWorkSpaceSize`. Comparing to the cuDNN code path suggested a simple and acceptable fix. 

Regarding this specific function, the cuDNN code path will just call `cudnnDropoutGetStatesSize`. This matches that, and I don't see these as a hotspot (or even mildly concerning) in the context of other performance issues.

include/lbann/utils/dnn_lib/miopen/local_response_normalization.hpp
benson31: These aren't really causing concern in the profiles. Alexnet is the only model committed to LBANN that uses this. The cuDNN codepath returns `0UL`, but MIOpen will fail in this case.

bvanessen (approved):

PR #2131, opened 2022-08-09
Add padding transformation for vision.

tnat410:
Adding padding for vision in the transformation pipeline of the data reader.
Padding the given image on all sides with 0 (black) value.


bvanessen (approved):

benson31 (approved):
LGTM

PR #2134, opened 2022-08-09
Add image layer for cutout operation.

tnat410:
The cutout layer applies a fixed-size zero mask to a random location of each input image.
The cutout layer selects a pixel coordinate within the image as a center point and then places the cutout mask around that location.
The cutout layer only has one square patch as the cutout region.


benson31 (approved):
LGTM

PR #2138, opened 2022-08-30
Adding support for the Flux batch scheduler.  

bvanessen:
Updating the Corona CI pipeline to use flux commands.  Adding a Flux launcher script to the
PFE.

Temporarily disabling all of the other pipelines.

### DIFF
--- a/.gitlab/corona/pipeline.yml
+++ b/.gitlab/corona/pipeline.yml
@@ -119,7 +119,7 @@ integration tests:
     - cd ci_test/integration_tests
     - export WEEKLY_FLAG=${WITH_WEEKLY:+--weekly}
     - echo "python3 -m pytest -s -vv --durations=0 ${WEEKLY_FLAG} --junitxml=results.xml"
-    - python3 -m pytest -s -vv --durations=0 ${WEEKLY_FLAG} --junitxml=results.xml
+#    - python3 -m pytest -s -vv --durations=0 ${WEEKLY_FLAG} --junitxml=results.xml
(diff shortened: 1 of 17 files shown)


.gitlab/corona/pipeline.yml
```diff
@@ -190,6 +201,9 @@ release allocation:
     # problems if it's not set.
     LBANN_CI_RUN_WEEKLY: ${LBANN_CI_RUN_WEEKLY}
 
+    # Preserve any spack-build-* directories to allow for faster testing
+    GIT_CLEAN_FLAGS: -ffdx -e spack-build-*
+
```
benson31, line 206: I don't understand why these are needed? The catch tests run with the build, the relevant logs are stored as artifacts, and the `ci_tests/` directory doesn't use the build tree at all.

benson31: Oh nevermind. This is to accelerate successive pipelines in which one happens to have gotten the same runner (and Spack hashes LBANN the same way). My concern, though, is that, as hashes accumulate, we will just eat up our disk space with stale build directories. We may want to augment this approach with a cron job to clean that waste up (e.g., directories with mtime older than a week or something).

bvanessen: @benson31 I agree that we should have a cron job to clean these directories and spack environments.

benson31 (approved):
General code looks great. There's some extra debugging output in there still, but it's just Gitlab logging, so that's fine.

Just a note that the Corona unit/integration tests are commented out.

benson31 (approved):
LGTM

PR #2140, opened 2022-09-23
Enable libfabric plugin support with (N|R)CCL

bvanessen:
Adding support for building LBANN with libfabric support in RCCL and aluminum.

### DIFF
--- a/python/lbann/contrib/lc/launcher.py
+++ b/python/lbann/contrib/lc/launcher.py
@@ -79,6 +79,14 @@ def make_batch_script(
     # present in MVAPICH2-2.3rc2.
     set_environment('MV2_USE_RDMA_CM', 0)
 
+    # Optimizations for Tioga
+    if system in ('tioga'):
+        #set_environment('NCCL_SOCKET_IFNAME', 'hsi')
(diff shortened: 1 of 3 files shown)


python/lbann/contrib/lc/launcher.py
```diff
@@ -79,6 +79,15 @@ def set_environment(key, default):
     # present in MVAPICH2-2.3rc2.
     set_environment('MV2_USE_RDMA_CM', 0)
 
+    # Optimizations for Tioga
+    if system in ('tioga'):
+        #set_environment('NCCL_SOCKET_IFNAME', 'hsi')
+        set_environment('MIOPEN_DEBUG_DISABLE_FIND_DB', '1')
+        set_environment('MIOPEN_DISABLE_CACHE', '1')
+        set_environment('NCCL_DEBUG', 'INFO')
+        set_environment('NCCL_DEBUG_SUBSYS', 'INIT')
```
benson31, line 88: I don't think we want to keep these long-term.

benson31 (approved):

python/lbann/contrib/lc/launcher.py
bvanessen: Good catch.

PR #2153, opened 2022-11-08
Fixed bad convolution parameters producing incorrect layer shapes.

fiedorowicz1:


bvanessen (commented):
@fiedorowicz1 Are these changes in line with the appropriate publication about the ExaGAN model?

fiedorowicz1:
I'm not sure if these changes are consistent with the original publication, but they appear to be consistent with the Multi-GAN setup.

bvanessen (approved):

PR #2154, opened 2022-11-09
Updated Corona to 5.3.0

bvanessen:


scripts/customize_build_env.sh
```diff
@@ -107,7 +107,7 @@ set_center_specific_modules()
                 MODULE_CMD="module --force unload StdEnv; module load gcc/10.2.1 mvapich2/2.3 python/3.7.2"
                 ;;
             "zen" | "zen2") # Corona
-                MODULE_CMD="module load gcc-tce/10.3.1 rocm/5.2.0 mvapich2-tce/2.3.6"
+                MODULE_CMD="export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/lib64; module load StdEnv gcc-tce/10.3.1 cmake/3.23.1 openmpi-tce/4.1.2 rocm/5.3.0"
```
benson31, line 110: IMNSHO, if a library in `/usr/lib64` cannot be loaded (i.e., a library that's actually `/usr/lib64/libFOO.so` and not something like `/usr/lib64/foo/libFOO.so`), this is a bug in the OS, the shell, the loader, or (most likely) Spack. Things installed to this location are supposed to "just work" and `LD_LIBRARY_PATH` should *only* be used for user-installed software.

benson31: That said, this is _fine_ as a workaround, but we should not make it permanent.

benson31 (approved):

scripts/customize_build_env.sh
bvanessen: I have fixed this in spack.

PR #2157, opened 2022-11-21
Add alternating optimizer updates

fiedorowicz1:
Added a callback to support GAN training and alternate generator/discriminator updates by freezing/unfreezing a set of weights. Also added a layer which acts as the zero/identity function depending on whether the layer is frozen or not. This is useful for reusing the same compute graph to train the generator and the discriminator by simply switching which loss function to use.


include/lbann/layers/transform/identity_zero.hpp
```diff
@@ -0,0 +1,124 @@
...
+  void setup_dims(DataReaderMetaData& dr_metadata) override {
+    data_type_layer<TensorDataType>::setup_dims(dr_metadata);
+    this->set_output_dims(this->get_input_dims());
+
+    // Check that input dimensions match
+    const auto& output_dims = this->get_output_dims();
+    for (int i = 0; i < this->get_num_parents(); ++i) {
+      if (this->get_input_dims(i) != output_dims) {
+        const auto& parents = this->get_parent_layers();
+        std::stringstream err;
```
benson31, line 77:
```suggestion
        std::ostringstream err;
```

bvanessen (approved):

PR #2163, opened 2022-12-15
Identified that there are multiple unnecessary calls to EL::Empty

bvanessen:
function when resetting matrices at the start of each forward and backward prop phases.  Currently this seems to be triggering a memory leak in Hydrogen, but they should also be unnecessary becasue there are already calls to resize the matrix and they will be overwritten.

Enable temporary performance testing by providing command line and environment options to enable application of the Empty function to intermediate state, zero'ing out that state, or just reusing dirty buffers with the understanding that the matrices will be overwritten.

### DIFF
--- a/include/lbann/layers/transform/concatenate.hpp
+++ b/include/lbann/layers/transform/concatenate.hpp
@@ -29,6 +29,8 @@
 
 #include "lbann/layers/data_type_layer.hpp"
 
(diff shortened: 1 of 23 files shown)


bvanessen:
@ fiedorowicz1 Please checkout this PR to see if it fixes the performance bug.

PR #2166, opened 2022-12-22
Fixed a bug in the GPU-based computation of the batchnorm statistics

bvanessen:
when working with either serialized GPU kernels or very large grid sizes that clip.  Note that this should not be a commonly observed bug.


bvanessen:
Note that this was identified by @szaman19

szaman19 (approved):

benson31 (approved):
LGTM

PR #2173, opened 2023-01-26
Disable local spack mirrors by default

bvanessen:
Updated build script to opt out of using local center mirrors by default.


benson31 (approved):
LGTM

PR #2175, opened 2023-01-27
Fix include path and remove Python.h inclusion from headers

tbennun:
This PR moves the inclusion of Python.h from the headers to the two .cpp files that use it. It also fixes an include path that did not include a top-level directory.


tbennun:
@benson31 @bvanessen forgot to tag

benson31 (approved):
Thanks for cleaning this up.

bvanessen (approved):

PR #2176, opened 2023-01-27
Move protobuf headers to `lbann/proto`

tbennun:
As part of installation, the `*.pb.h` files are found in the top-level directory, which might cause name clashes. This PR moves those files into the lbann/proto include subdirectory.

@benson31 @bvanessen

### DIFF
--- a/src/utils/unit_test/protobuf_utils_test.cpp
+++ b/src/utils/unit_test/protobuf_utils_test.cpp
@@ -30,7 +30,7 @@
 #include "lbann/utils/protobuf/impl.hpp"
 
 #include <google/protobuf/descriptor.h>
-#include "lbann/proto/protobuf_utils_test_messages.pb.h"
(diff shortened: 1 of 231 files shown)


bvanessen (approved):

src/utils/unit_test/protobuf_utils_test.cpp
```diff
@@ -30,7 +30,7 @@
 #include "lbann/utils/protobuf/impl.hpp"
 
 #include <google/protobuf/descriptor.h>
-#include <protobuf_utils_test_messages.pb.h>
+#include "lbann/proto/protobuf_utils_test_messages.pb.h"
```
benson31, line 33:
```suggestion
#include "protobuf_utils_test_messages.pb.h"
```
This is private-to-the-testing, so it does not need the same treatment.

PR #2177, opened 2023-01-28
Fix typo in dist_embedding_layer

szaman19:
This breaks the build only when nvshmem is enabled, so may be missed by CI.


benson31 (approved):
LGTM

bvanessen (approved):

PR #2180, opened 2023-01-30
Remove references to Aluminum's MPI_CUDA backend

### LINKED ISSUE #2179: Use Aluminum's Host-Transfer backend rather than MPI-CUDA
`lbann_comm` [appears](https://github.com/LLNL/lbann/blob/develop/src/comm.cpp#L460-L466) to still attempt to call allreduces on the MPI-CUDA backend. Those were moved out of the host-transfer backend in Aluminum v0.5.

benson31:
These are vestigial traces of a bygone era. Most of the `MPICUDABackend` was split out into the `HostTransferBackend`. LBANN was only using codepaths that are now in the `HostTransferBackend`. This removes some lingering references to the `MPICUDABackend` that are now erroneous after this split.

Closes #2179.

### DIFF
--- a/include/lbann/comm.hpp
+++ b/include/lbann/comm.hpp
@@ -50,10 +50,10 @@ namespace lbann {
 #endif
 
 /** Grid types in sub-grid parallelism (2nd order) */
-enum class GridType 
-{   
-  NO_GRID = 0, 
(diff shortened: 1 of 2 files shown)


ndryden (approved):
Looks good to me.

PR #2186, opened 2023-02-02
Add a command line argument to optionally hang the LBANN app

benson31:
This hang occurs after library initialization but before any model/training setup. This is superior to the hang callback because it allows us to debug almost everything that LBANN does, rather than just things during or after the training loop. (If we're ok with the inherent fragility of the "environment variable to MPI rank" method, we could even move this before the library initialization function.)

Because this is a property of the driver and not the LBANN library, the command line argument is only added in the driver and not in "options.cpp" or any other such place, and its key to the argument parser map is not made public via the macro idiom in "options.hpp". It would be an abuse for library code to reference this.

### DIFF
--- a/model_zoo/lbann.cpp
+++ b/model_zoo/lbann.cpp
@@ -40,6 +40,12 @@
 
 #include <cstdlib>
 
+#if __has_include(<unistd.h>) && __has_include(<sys/types.h>)
+#define LBANN_HAS_UNISTD_H
+#include <sys/types.h>
+#include <unistd.h>
+#endif
+
 using namespace lbann;
 
 namespace {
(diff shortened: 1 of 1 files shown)


model_zoo/lbann.cpp
```diff
@@ -102,6 +113,24 @@ int main(int argc, char* argv[])
...
 
+    auto const hang_rank = arg_parser.get<int>("hang on start");
+    if (hang_rank >= 0)
+    {
+      auto const my_rank = comm->get_rank_in_world();
+      if (hang_rank == my_rank) {
+#ifdef LBANN_HAS_UNISTD_H
+        char hostname[1024];
+        gethostname(hostname, 1024);
+        std::cout << "LBANN [hang]: (hostname: " << hostname << ", pid: " << getpid()
```
benson31, line 124: Just as a note: The use of `std::cout` instead of `std::cerr` here is intentional. Due to the different buffering semantics between the two streams, I encountered situations in which this would it the console prior to a bunch of `std::cout` output. So I add it to that stream so it should be guaranteed to be the last thing to print prior to actually hanging, making it easily found by users. An alternative would be to flush `std::cout` and then push this to `std::cerr`.

ndryden (approved):
Looks good to me.

bvanessen (approved):

PR #2188, opened 2023-02-03
Clean up compiler warnings

benson31:
This covers Clang builds on Pascal and GCC builds on Lassen. There are still some `-Wdeprecated-declarations` warnings, but those aren't as trivial to squash.

### DIFF
--- a/src/data_readers/data_reader_hdf5_legacy.cpp
+++ b/src/data_readers/data_reader_hdf5_legacy.cpp
@@ -281,7 +281,7 @@ bool hdf5_reader<TensorDataType>::fetch_label(Mat& Y, int data_id, int mb_idx) {
   assert_always(m_hyperslab_labels);
   assert_always(m_use_data_store);
   TensorDataType *buf = nullptr;
-  assert_eq(Y.Height(), m_num_features);
+  assert_eq((unsigned long) Y.Height(), m_num_features);
   conduit::Node node;
   const conduit::Node& ds_node = m_data_store->get_conduit_node(data_id);
   node.set_external(ds_node);
@@ -303,7 +303,7 @@ bool hdf5_reader<TensorDataType>::fetch_data_field(data_field_type data_field, C
   assert_always(m_hyperslab_labels);
   assert_always(m_use_data_store);
   TensorDataType *buf = nullptr;
-  assert_eq(Y.Height(), m_num_features);
+  assert_eq((unsigned long) Y.Height(), m_num_features);
   conduit::Node node;
   if (data_store_active() || m_data_store->has_conduit_node(data_id)) {
     prof_region_begin("get_conduit_node", prof_colors[0], false);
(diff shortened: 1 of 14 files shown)


src/data_readers/data_reader_hdf5_legacy.cpp
```diff
@@ -328,10 +328,13 @@ template <typename TensorDataType>
...
-  assert_eq(X.Height(),
-            m_num_features / dc::get_number_of_io_partitions()
-            / (sizeof(DataType) / sizeof(TensorDataType)));
+  assert_eq(sizeof(DataType) % sizeof(TensorDataType), 0ul);
+  assert_eq((unsigned long) X.Height(),
+            m_num_features / dc::get_number_of_io_partitions());
+  //        / (sizeof(DataType) / sizeof(TensorDataType)));
+  // FIXME (trb 02/03/2023) -- when DataType is 'float' and
+  // TensorDataType is 'double', this last operand will be 0, which is
+  // invalid.
```
ndryden, line 337: Is this assert still valid with the last bit commented out?

ndryden (approved):
Looks good to me.

src/data_readers/data_reader_hdf5_legacy.cpp
benson31: My thinking was that when `DataType` is `float` and `TensorDataType` is `float`, the commented-out term is 1, so that's the same. Then when `DataType` is `float` and `TensorDataType` is `double, the commented-out term is 0 and everything explodes. My decision to comment it out and leave the `FIXME` is due to the fact that `TensorDataType` should only actually be `float` in real LBANN usage as it stands today. I'll need to actually look at the surrounding code to see if it actually makes sense to leave it this way long-term.

bvanessen: Why don't we add an explicit assert that `sizeof(DataType) == sizeof(TensorDataType)` so that this weird case is actually caught?

bvanessen (changes requested):
Can / should we add an asset to actually make the data reader fail when in this weird case?

benson31:
We don't actually need to. We need to just tell the compiler this cannot be zero because of the previous assertion.

1. If `sizeof(DataType) >= sizeof(TensorDataType)`, then `sizeof(DataType) / sizeof(TensorDataType) >= 1`.
2. If `sizeof(DataType) < sizeof(TensorDataType)`, then `sizeof(DataType) % sizeof(TensorDataType) = sizeof(DataType) >= 1 > 0`, so the assertion on the previous line will have failed.

I'll clean this up to clarify this in the code.

src/data_readers/data_reader_hdf5_legacy.cpp
benson31: Please have a look at my revised changes and the associated note. In particular, it's good to observe that `sizeof(DataType) % sizeof(TensorDataType)` is actually `constexpr`, so the first `assert_eq` can be replaced by a `constexpr` assertion. This seems to have been sufficient for the compiler to determine that the potentially-dangerous `<num> / (sizeof(DataType) / sizeof(TensorDataType))` operation could never be reached in a case in which `sizeof(DataType) / sizeof(TensorDataType) == 0` (as one would have hoped).

bvanessen (approved):

PR #2190, opened 2023-02-07
Fix an explicit path in Python Front-End

benson31:
This shouldn't code a specific ROCm version. If the user wants this, they can load the right module or set this path explicitly. Or Spack should be able to do that for them.

As a separate issue, we should talk about the `CRAY_LD_LIBRARY_PATH` addition as well, since that _also_ seems to be a Spack issue. It's not needed in any sane build.


ndryden (approved):
:shipit:

bvanessen (approved):

PR #2199, opened 2023-02-21
Add pytorch reference implementation of cosmoflow

fiedorowicz1:
Adds a pytorch reference implementation of cosmoflow utilizing distributed data parallel and automatic mixed precision.

### DIFF
--- a/applications/physics/cosmology/cosmoflow/ref/README.md
+++ b/applications/physics/cosmology/cosmoflow/ref/README.md
@@ -1,3 +1,3 @@
 # PyTorch Reference Implementation of CosmoFlow
 
-This directory contains a reference implementation of CosmoFlow in PyTorch as it is implemented for LBANN in the parent directory. This implementation uses `DistributedDataParallel` (DDP) to run on multiple GPUs. The included `batch.sh` script provides an example for training this model with DDP on Lassen.
+This directory contains a reference implementation of CosmoFlow in PyTorch as it is implemented for LBANN in the parent directory. This implementation uses `DistributedDataParallel` (DDP) to run on multiple GPUs. The included `batch.sh` script provides an example for training this model with DDP on Lassen. `main.py` assumes some OpenMPI environment variables have been set, therefore, it may need to be modified to run on other systems.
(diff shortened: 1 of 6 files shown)


applications/physics/cosmology/cosmoflow/ref/batch.sh
```diff
@@ -0,0 +1,17 @@
...
+#BSUB -J cosmoflow
+#BSUB -q pbatch
+#BSUB -alloc_flags ipisolate
+
+firsthost=`jsrun --nrs 1 -r 1 /bin/hostname`
+export MASTER_ADDR=$firsthost
+export MASTER_PORT=23456
+
+echo "Started at $(date)"
+jsrun -r 4 --bind none -c 10 python main.py --preload --enable-amp --use-batchnorm
```
ndryden, line 16: Minor thing, but I recall that jsrun/SMPI can sometimes complain if you never initialize MPI (which we don't, due to using NCCL). Passing `--smpiargs="off"` avoids this.

applications/physics/cosmology/cosmoflow/ref/README.md
```diff
@@ -0,0 +1,3 @@
+# PyTorch Reference Implementation of CosmoFlow
+
+This directory contains a reference implementation of CosmoFlow in PyTorch as it is implemented for LBANN in the parent directory. This implementation uses `DistributedDataParallel` (DDP) to run on multiple GPUs. The included `batch.sh` script provides an example for training this model with DDP on Lassen.
```
ndryden, line 3: You might also note that the script hardcodes OpenMPI env variables, so even if you write a new batch script, it still may not run on other systems.

ndryden (approved):
Looks reasonable to me. Just a couple suggestions to clarify things.

applications/physics/cosmology/cosmoflow/ref/README.md
fiedorowicz1: Fixed in https://github.com/LLNL/lbann/pull/2199/commits/05265916be3ce2adafa5bf68da350d6e3d680b14

PR #2204, opened 2023-02-23
Put a guard around a GPU test for external layer

bvanessen:


src/layers/unit_test/external_layer_test.cpp
```diff
@@ -54,13 +54,15 @@ TEST_CASE("External Layer test", "[layer][externallayer]")
...
 
+#ifdef LBANN_HAS_GPU
     // Configuration that does not exist
     lbann::Layer* nonexistent = nullptr;
     REQUIRE_NOTHROW(nonexistent = setupfunc(lbann_data::COMPLEX_DOUBLE,
                                             lbann::data_layout::DATA_PARALLEL,
                                             El::Device::GPU,
                                             &world_comm));
     CHECK(!IsValidPtr(nonexistent));
+#endif // LBANN_HAS_GPU
```
benson31, line 65:
```suggestion
    // Configuration that does not exist
    lbann::Layer* nonexistent = nullptr;
    REQUIRE_NOTHROW(nonexistent = setupfunc(lbann_data::COMPLEX_DOUBLE,
                                            lbann::data_layout::DATA_PARALLEL,
                                            El::Device::CPU,
                                            &world_comm));
    CHECK(!IsValidPtr(nonexistent));
```

benson31 (changes requested):

src/layers/unit_test/external_layer_test.cpp
tbennun: I agree that this change captures what we want to test and doesn’t require the ifdef

tbennun (approved):

benson31 (approved):

PR #2205, opened 2023-02-23
Replace Relu with LeakyRelu in Unet3d app.

KIwabuchi:
This PR replaces Relu() with LeakyRelu() in the Unet 3d model because the current Relu implementation in distconv does not work in this context.
Specifically, the distconv's ReLU activation function uses cuDNN's ReLU implementation, and it requires the same stride sizes in the tensors.


bvanessen (approved):

benson31 (approved):
LGTM

PR #2211, opened 2023-03-02
Added documentation for operator layers

graham63:
Add documentation for the Operators supported in LBANN.


docs/operators.rst
```diff
@@ -0,0 +1,172 @@
...
+
+------------------------------------------------
+Constant Subtract
+------------------------------------------------
+
+Subtract each input value from a constant.
+
+.. math::
+
+   x = (c - x)
```
benson31, line 87:
```suggestion
   y = (c - x)
```

docs/operators.rst
```diff
@@ -0,0 +1,172 @@
...
+
+------------------------------------------------
+Scale
+------------------------------------------------
+
+Scale each input value by a constant.
+
+.. math::
+
+   x = c * x
```
benson31, line 136:
```suggestion
   y = c * x
```

docs/operators.rst
```diff
@@ -0,0 +1,172 @@
...
+
+------------------------------------------------
+Subtract Constant
+------------------------------------------------
+
+Subtract a constant from each input value.
+
+.. math::
+
+   x = (x - c)
```
benson31, line 172:
```suggestion
   y = (x - c)
```

benson31 (changes requested):
Overall this is good, just a few suggestions so it doesn't look like
they're in-place operators (they aren't).

benson31 (commented):
Also, do we want to maybe put the Python Front-End names for these in
there? (For both the operator and the OperaorLayer shortcut?)

bvanessen (approved):

PR #2214, opened 2023-03-03
Use Aluminum's NCCLBackend instead of HostTransferBackend

benson31:
This must be merged before LLNL/DiHydrogen#68, but it does not require that PR to be merged simultaneously.


ndryden (approved):
Looks good to me.

PR #2218, opened 2023-03-07
Add guards on distconv CI tests

szaman19:
Adds guards on the Distconv-enabled layer CI tests to check for LBANN Distconv feature or skip the tests.


bvanessen (approved):
LGTM

PR #2221, opened 2023-03-10
Bugfix CI and build infrastructure

bvanessen:


scripts/build_lbann.sh
```diff
@@ -38,7 +38,7 @@ CMD_LINE_VARIANTS=
 
 # Default versions of Hydrogen, DiHydrogen, and Aluminum - use head of repo
 HYDROGEN_VER="@develop"
-ALUMINUM_VER="@1.2.1"
+ALUMINUM_VER="@1.2.3:"
```
tbennun, line 41: what does the trailing colon do?

ndryden: "Any version newer than". (I had to ask the same question. 😉 )

tbennun (approved):

scripts/build_lbann.sh
bvanessen: Yes that is correct.

PR #2241, opened 2023-03-31
Fix an issue in which makefile wouldn't create needed directories

### LINKED ISSUE #2240: LBANN cannot be built with "make"
Nobody should build LBANN with `make` anyway (`pip install ninja` or however you might want to acquire it), but if someone were to try to use the `Unix Makefiles` generator with CMake, they would get a message similar to the following:
```
> make
[  0%] Running protoc on LBANN protobuf sources.
/path/to/lbann/build/src/proto/lbann/proto/: No such file or directory
make[2]: *** [src/proto/CMakeFiles/LbannProto_genSrc.dir/build.make:117: src/proto/lbann/proto/lbann.pb.cc] Error 1
make[1]: *** [CMakeFiles/Makefile2:3308: src/proto/CMakeFiles/LbannProto_genSrc.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

benson31:
Closes #2240.

### DIFF
--- a/src/proto/CMakeLists.txt
+++ b/src/proto/CMakeLists.txt
@@ -58,6 +58,7 @@ if (LBANN_HAS_PYTHON_FRONTEND)
 endif (LBANN_HAS_PYTHON_FRONTEND)
   endforeach ()
   add_custom_command(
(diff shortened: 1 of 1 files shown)


ndryden (approved):
Nice catch.

PR #2245, opened 2023-04-06
Improve HDF5 data reader robustness

bvanessen:
Added support for ingesting all integer and real format numbers in the
HDF5 data reader.  Switched to using the conduit type conversion routines for HDF5 to Conduit coercion.

Added warnings if integer formats are not coerced into real formats.

Added a guard to throw an error if character / string based formats are in the data scheme of an HDF5 sample.

Refactored the HDF5 data reader to separate the load_sample and load_sample_from_sample_list.  This allows for a simple white box catch2 test to be added.  Added catch2 tests for loading HRRL HDF5 and C3FD HDF5 files from "disk".


KIwabuchi:
@bvanessen Just in case, what is the difference between src/data_readers/unit_test/test_data/hdf5_unet3d_test_data_and_schemas.yaml and src/data_readers/unit_test/test_data/hdf5_3D_UNet_test_data_and_schemas.yaml?
Maybe we don't need one of them?

benson31 (approved):
Seems fine. Others should comment on whether it's doing "the right thing".

PR #2250, opened 2023-04-17
Synthetic ImageNet support in ResNet/AlexNet

ndryden:
Adds a `--synthetic` flag to the ResNet and AlexNet applications to use LBANN's synthetic data reader rather than real ImageNet data.


benson31 (approved):
LGTM. Thank you for doing the type annotations.

PR #2252, opened 2023-04-18
Fix error in NASNet integration test

ndryden:
The NASNet integration test would error out with an undefined variable when attempting to print how its tolerance was violated. This appears to have been introduced in #2141. (It is a separate issue that the tolerance is being violated.)

I fixed the variable naming to be consistent with our other tests.


benson31 (approved):
LGTM

PR #2254, opened 2023-04-21
Support synthetic data in CosmoFlow

ndryden:
Does what the title says.

Note that the number of samples is based on the full CosmoFlow dataset.

### DIFF
--- a/applications/physics/cosmology/cosmoflow/train_cosmoflow.py
+++ b/applications/physics/cosmology/cosmoflow/train_cosmoflow.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 import cosmoflow_model
 
 import argparse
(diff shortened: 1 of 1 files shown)


fiedorowicz1 (approved):
Looks good. On a single Lassen node with a mini-batch size less than or equal to four, it works. However, a mini-batch size greater than four segfaults. Also, trying to use distconv throws an error. These are likely synthetic data reader issues not related to this PR.

ndryden:
I'm not seeing segfaults or distconv issues, but it may be due to differences in how it's being run -- what are your exact arguments to `train_cosmoflow.py`? I wouldn't be surprised by issues, since the synthetic data reader hasn't been well-tested of late.

fiedorowicz1:
I ran `python train_cosmoflow.py --nodes=1 --procs-per-node=4 --depth-groups=1 --synthetic --mini-batch-size=8`.

bvanessen (approved):

PR #2257, opened 2023-04-28
In-Place Layers

tbennun:
This PR introduces in-place operations to LBANN. The work should reduce memory consumption for activations.
The decision of whether to make a layer in-place depends on two factors:
1. The layer specifies (via a method) that it can be computed in-place
2. All parents of this layer do not need their original forward-pass output activations in their backpropagation computation (specified via another new method)


src/layers/layer.cpp
```diff
@@ -525,6 +529,29 @@ void Layer::setup_pointers()
                 get_child_names(),
                 ")");
   }
+
+  // Set whether this layer will run in-place
+  if (!this->can_run_inplace() || this->distconv_enabled()) {
```
bvanessen, line 534:
```suggestion
  if (!this->can_run_inplace()
#ifdef LBANN_HAS_DISTCONV
      || this->distconv_enabled()
#endif // LBANN_HAS_DISTCONV
      ) {
```

benson31: While I agree with this change in the current state of the code, a better solution might be to have `layer::distconv_enabled()` always defined, just return `false` if no distconv support. I'm testing this now, and I'll PR if everything is ok in CI.

src/layers/layer.cpp
```diff
@@ -525,6 +529,44 @@ void Layer::setup_pointers()
                 get_child_names(),
                 ")");
   }
+
+  // Set whether this layer will run in-place
+
+  // Check for environment variable that disables this behavior
+  bool envvar_disable_inplace = false;
+  char* no_inplace = getenv("LBANN_NO_INPLACE");
```
bvanessen, line 537: @tbennun This should use the options class.

tbennun: Updated, thanks!

include/lbann/layers/activations/relu.hpp
```diff
@@ -75,6 +75,11 @@ class relu_layer : public data_type_layer<TensorDataType>
   std::string get_type() const override { return "ReLU"; }
   data_layout get_data_layout() const override { return T_layout; }
   El::Device get_device_allocation() const override { return Dev; }
+  bool can_run_inplace() const override { return true; }
+  int get_backprop_requirements() const override
+  {
+    return ERROR_SIGNALS | ACTIVATIONS;
```
bvanessen, line 81: @tbennun Am I missing something, but I think that this should only rely on `PREV_ACTIVATIONS`

bvanessen:
```suggestion
    return ERROR_SIGNALS | PREV_ACTIVATIONS;
```

tbennun: ReLU is a special case: if the output activations are zero, then the input error signals should be zero as well. Otherwise everything is kept equal. This is the property that we use to run the layer in-place.

include/lbann/layers/transform/constant.hpp
```diff
@@ -58,6 +58,8 @@ class constant_layer : public data_type_layer<TensorDataType>
   std::string get_type() const override { return "constant"; }
   data_layout get_data_layout() const override { return T_layout; }
   El::Device get_device_allocation() const override { return Dev; }
+  bool can_run_inplace() const override { return false; }
+  int get_backprop_requirements() const override { return ERROR_SIGNALS; }
```
benson31, line 62:
```suggestion
  int get_backprop_requirements() const override { return NO_REQUIREMENTS; }
```
?

include/lbann/layers/transform/stop_gradient.hpp
```diff
@@ -61,6 +61,8 @@ class stop_gradient_layer : public data_type_layer<TensorDataType>
   std::string get_type() const override { return "stop_gradient"; }
   data_layout get_data_layout() const override { return T_layout; }
   El::Device get_device_allocation() const override { return Dev; }
+  bool can_run_inplace() const override { return false; }
+  int get_backprop_requirements() const override { return ERROR_SIGNALS; }
```
benson31, line 65:
```suggestion
  int get_backprop_requirements() const override { return NO_REQUIREMENTS; }
```
?

benson31 (approved):
LGTM

benson31 (approved):
LGTM

ci_test/integration_tests/test_integration_roberta.py
```diff
@@ -51,7 +51,7 @@
     'num_nodes': 1,
     'num_epochs': 0,
     'mini_batch_size': 16,
-    'expected_test_loss_range': (6.8, 7.1),
+    'expected_test_loss_range': (6.1, 7.1),
```
bvanessen, line 54:
```suggestion
    'expected_test_loss_range': (6.1, 7.1),    # Test loss range expanded from 6.8 to 6.1 5/4/23
```

ci_test/integration_tests/test_integration_roberta.py
```diff
@@ -68,7 +68,7 @@
     'num_nodes': 1,
     'num_epochs': 0,
     'mini_batch_size': 16,
-    'expected_test_loss_range': (6.75, 7.1),
+    'expected_test_loss_range': (6.1, 7.1),
```
bvanessen, line 71:
```suggestion
    'expected_test_loss_range': (6.1, 7.1),    # Test loss range expanded from 6.75 to 6.1 5/4/23
```

include/lbann/layers/loss/top_k_categorical_accuracy.hpp
```diff
@@ -69,6 +69,11 @@ class top_k_categorical_accuracy_layer : public data_type_layer<TensorDataType>
   std::string get_type() const override { return "top-k accuracy"; }
   data_layout get_data_layout() const override { return T_layout; }
   El::Device get_device_allocation() const override { return Dev; }
+  bool can_run_inplace() const override { return false; }
+  int get_backprop_requirements() const override
+  {
+    return ERROR_SIGNALS | PREV_ACTIVATIONS;
```
bvanessen, line 75: Maybe I am missing something, but I don't see a backprop method, nor do I see where it requires `PREV_ACTIVATIONS`

src/layers/data_type_layer.cpp
```diff
@@ -1122,8 +1215,13 @@ void data_type_layer<InputTensorDataType,
                      OutputTensorDataType>::clear_prev_error_signals_()
 {
   if (!m_persistent_error_signals) {
-    for (auto& es : m_gradient_wrt_outputs)
-      es->Empty(true);
+    for (int i = 0; i < (int)m_gradient_wrt_outputs.size(); ++i) {
+      // Error signal is reused in input layer
```
bvanessen, line 1219: I don't understand this comment.

tbennun: After local backprop is complete, the error signals are propagated to the parent layers and then the input error signals are cleared. Since in-place layers share the memory, clearing it would clear the reused error signals as well, so we must skip this step.

include/lbann/layers/regularizers/batch_normalization.hpp
```diff
@@ -246,6 +246,11 @@ class batch_normalization_layer : public data_type_layer<TensorDataType>
   std::string get_type() const override { return "batch normalization"; }
   data_layout get_data_layout() const override { return T_layout; }
   El::Device get_device_allocation() const override { return Dev; }
+  bool can_run_inplace() const override { return false; }
+  int get_backprop_requirements() const override
+  {
+    return ERROR_SIGNALS | PREV_ACTIVATIONS;
```
bvanessen, line 252:
```suggestion
    return ERROR_SIGNALS | WEIGHTS | PREV_ACTIVATIONS;
```

include/lbann/layers/regularizers/dropout.hpp
```diff
@@ -114,6 +114,8 @@ class dropout : public data_type_layer<TensorDataType>
   std::string get_type() const override { return "dropout"; }
   data_layout get_data_layout() const override { return T_layout; }
   El::Device get_device_allocation() const override { return Dev; }
+  bool can_run_inplace() const override { return true; }
+  int get_backprop_requirements() const override { return ERROR_SIGNALS; }
```
bvanessen, line 118:
```suggestion
  int get_backprop_requirements() const override { return ERROR_SIGNALS | PREV_ACTIVATIONS; }
```

bvanessen: Hmmm, That is a mistake on my recommendation.

ci_test/integration_tests/test_integration_roberta.py
tbennun: Updated.

include/lbann/layers/loss/top_k_categorical_accuracy.hpp
tbennun: Added comment.

src/layers/data_type_layer.cpp
tbennun: Updated comment.

include/lbann/layers/transform/weights.hpp
```diff
@@ -70,6 +70,8 @@ class weights_layer : public data_type_layer<TensorDataType>
   std::string get_type() const override;
   data_layout get_data_layout() const override;
   El::Device get_device_allocation() const override;
+  bool can_run_inplace() const override { return false; }
+  int get_backprop_requirements() const override { return ERROR_SIGNALS; }
```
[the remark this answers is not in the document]
bvanessen:
```suggestion
  int get_backprop_requirements() const override { return ERROR_SIGNALS | WEIGHTS; }
```

bvanessen (approved):

PR #2258, opened 2023-04-28
Always enable the check for distconv

benson31:
This will just return `false` when DistConv support is not enabled.


tbennun (approved):
LGTM

bvanessen (approved):

PR #2262, opened 2023-05-08
Refactor Leaky ReLU to Use Activations in Backward

fiedorowicz1:
Replace `prev_activations` with `activations` in backprop to support in-place. Passing standard and distconv `leaky_relu` unit tests on Lassen.


tbennun (approved):
LGTM, leaky ReLU is bijective and the `activation > 0` property is preserved from `prev_activation` so this works as-is.

src/layers/activations/leaky_relu.cpp
```diff
@@ -84,7 +84,7 @@ template <typename TensorDataType, data_layout Layout, El::Device Device>
 void leaky_relu_layer<TensorDataType, Layout, Device>::bp_compute()
 {
   local_bp<TensorDataType>(this->m_negative_slope,
-                           this->get_local_prev_activations(),
+                           this->get_local_activations(),
```
bvanessen, line 87: @tbennun @fiedorowicz1 Shouldn't this instigate a change in the header file about the backprop requirements.

fiedorowicz1: Yes, I was expecting this PR would be merged before in-place. Fixed now.

bvanessen (approved):

PR #2264, opened 2023-05-16
Add an LBANN Python unit test wrapper and utilities

tbennun:
An example of the new unit test wrapper syntax:

```python
import lbann
import numpy as np
import test_util
import pytest

@pytest.mark.parametrize('constant', [0, 1])
@test_util.lbann_test()
def test_simple(constant):
    np.random.seed(20230515)
    x = np.random.rand(2, 2).astype(np.float32)
    ref = x + constant

    tester = test_util.ModelTester()
    x = tester.inputs(x)
    ref = tester.make_reference(ref)

    # Test layer
    y = lbann.AddConstant(x, constant=constant)
    tester.set_loss(lbann.MeanSquaredError(y, ref))
    return tester
```

### DIFF
--- a/ci_test/common_python/single_tensor_data_reader.py
+++ b/ci_test/common_python/single_tensor_data_reader.py
@@ -1,0 +1,31 @@
+"""
+Simple data reader that opens one file with one tensor. Used for unit testing.
+"""
+import numpy as np
+
+# Lazy-load tensor
+tensor = None
+
+
+def lazy_load():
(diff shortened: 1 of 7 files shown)


benson31 (approved):
I really like this. Thank you!

fiedorowicz1 (approved):
Looks very nice to use and will make new tests super easy to implement!

bvanessen (approved):

PR #2266, opened 2023-05-18
Remove the ROCm prefix from the implicit directories

benson31:
This should prevent CMake from removing `${ROCM_PATH}/lib` from the RPATH of all HIP-linked binaries (it also removes it from the CXX implicit link directories just to be safe).


tbennun (approved):
LGTM

bvanessen (approved):

PR #2268, opened 2023-05-30
Memory Optimizations for Convolution and Batch Normalization Layers

fiedorowicz1:
Allow the convolution and batch normalization layers to reuse their inputs for their error signals if the parent layer does not need its activations for backprop.


tbennun (approved):

benson31 (approved):
LGTM

bvanessen (approved):

PR #2269, opened 2023-05-30
Transformer optimizations

tbennun:
This PR includes two major modifications to transformers:
* Batching heads in multi-head attention into single operations instead of on a per-head basis
* Stacking the weights and biases for queries/keys/values in self-attention

The PR also includes fixes for the weighted sum operator (w.r.t. in-place layers) and path fixes for the transformer test

### DIFF
--- a/python/lbann/models/transformer.py
+++ b/python/lbann/models/transformer.py
@@ -397,6 +397,8 @@ class Transformer(lbann.modules.Module):
             for i in range(num_decoder_layers)
         ]
         self.separate_heads = self.decoder[0].attention1.separate_heads
+        assert all(dec.attention1.separate_heads == self.separate_heads
+                   for dec in self.decoder)
 
     def _positional_encoding(self, sequence_length):
         """Positional encodings corresponding to a sequence length.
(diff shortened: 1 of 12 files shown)


python/lbann/models/transformer.py
```diff
@@ -393,6 +396,7 @@ def __init__(
             )
             for i in range(num_decoder_layers)
         ]
+        self.separate_heads = self.decoder[0].attention1.separate_heads
```
ndryden, line 399: Is it worth sanity-checking that all layers have the same `separate_heads` setting?

python/lbann/modules/transformer.py
```diff
@@ -91,26 +108,118 @@ def forward(self, queries, keys, values, mask=None):
...
+            # Unstack
+            qkv_slice = lbann.Slice(qkv_fc,
+                                    axis=1,
+                                    slice_points=[
+                                        0, self.embed_dim, 2 * self.embed_dim,
+                                        3 * self.embed_dim
+                                    ])
+            queries_fc = lbann.Identity(qkv_slice)
+            keys_fc = lbann.Identity(qkv_slice)
+            values_fc = lbann.Identity(qkv_slice)
```
ndryden, line 133: Not a criticism, but we should make a note that this notation for slicing is quite unintuitive. Maybe we should revisit it at some point.

ndryden (approved):
This all looks correct to me. Nice work!

python/lbann/models/transformer.py
tbennun: Done

bvanessen (approved):

PR #2270, opened 2023-06-06
Fix get_distconv_environment

bvanessen:
Updated the get_distconv_environment function to be able to initialize
NVSHMEM if requested and to set LBANN_KEEP_ERROR_SIGNALS to 1, 
since that is always required for distconv right now.

Updated two nvshmem unit tests to initialize NVSHMEM.

### DIFF
--- a/ci_test/common_python/tools.py
+++ b/ci_test/common_python/tools.py
@@ -1021,8 +1021,11 @@ def gpus_per_node(lbann):
 # Get the environment variables for Distconv.
 def get_distconv_environment(init_nvshmem=False):
     # TODO: Use the default halo exchange and shuffle method. See https://github.com/LLNL/lbann/issues/1659
-    return {"LBANN_DISTCONV_HALO_EXCHANGE": "AL",
+    environment = {"LBANN_DISTCONV_HALO_EXCHANGE": "AL",
             "LBANN_DISTCONV_TENSOR_SHUFFLER": "AL",
             "LBANN_KEEP_ERROR_SIGNALS": "1",
-            "LBANN_INIT_NVSHMEM": (1 if init_nvshmem else 0),
         }
+    if init_nvshmem:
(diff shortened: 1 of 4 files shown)


ci_test/common_python/tools.py
```diff
@@ -1019,7 +1019,10 @@ def gpus_per_node(lbann):
...
 
 # Get the environment variables for Distconv.
-def get_distconv_environment():
+def get_distconv_environment(init_nvshmem=False):
     # TODO: Use the default halo exchange and shuffle method. See https://github.com/LLNL/lbann/issues/1659
     return {"LBANN_DISTCONV_HALO_EXCHANGE": "AL",
-            "LBANN_DISTCONV_TENSOR_SHUFFLER": "AL"}
+            "LBANN_DISTCONV_TENSOR_SHUFFLER": "AL",
+            "LBANN_KEEP_ERROR_SIGNALS": "1",
+            "LBANN_INIT_NVSHMEM": (1 if init_nvshmem else 0),
```
benson31, line 1027: I might actually suggest just always setting this to 1. If NVSHMEM is enabled, it should be `1`. If NVSHMEM is not enabled, its value will be ignored. You can then get rid of all the `init_nvshmem` cluttter.

benson31 (approved):

ci_test/common_python/tools.py
szaman19: Yeah, `LBANN_INIT_NVSHMEM` is ignored if `LBANN_HAS_NVSHMEM` is not set.

szaman19 (approved):

tbennun (approved):

ci_test/common_python/tools.py
bvanessen: @benson31 Here is a fix for that.

benson31 (approved):

PR #2283, opened 2023-06-28
Bugfix channelwise softmax

bvanessen:
The channelwise softmax header requires `exception.hpp`. To prevent extraneous transitive inclusion, the affected functions have been moved to an `impl.hpp` file, which is then included in the instantiation functions.


bvanessen:
@benson31 This passes CI.

benson31:
Could you please update the description to give at least some notion of the bug being fixed? I see you moving code; it's not clear why this movement was necessary.

benson31 (approved):

PR #2287, opened 2023-07-17
Add a simple extension-based switch to read protobuf in binary

benson31:
This is a quick hack. Pass a file to any of the prototext flags (`--prototext`, `--model`, etc) that contains the string "protobin" and it will be read as a binary file. Any other file will be read as "prototext".


tbennun (approved):
LGTM!

PR #2289, opened 2023-07-21
Training progress bar callback

tbennun:
Simple callback that prints progress during training epochs, and an ETA for epoch completion.

### DIFF
--- a/docs/callbacks.rst
+++ b/docs/callbacks.rst
@@ -84,6 +84,7 @@ Available Callbacks
 
    Export Onnx <callbacks/export_onnx>
    Summarize images <callbacks/summarize_images>
+   Print progress bar <callback/progress_bar>
 
 ..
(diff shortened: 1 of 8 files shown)


benson31 (approved):
My concern with something like this is mostly that the output would be exploded when dumped to a file and that another callback that writes output to stdout could interfere with the progress bar. It might be good to document this somewhere since it lands on the user either to ensure this won't happen or at least to understand that it might.

ndryden (approved):
Looks quite nice for me, thanks!

There's a couple minor comments about maybe not using OpenMP for this, but if it's too annoying to use the C++ STL timing stuff, that's fine.

Another suggestion is to consider checking whether `cout` is printing to a terminal or not with `isatty` to avoid potential footguns if output is going to a file.

PR #2292, opened 2023-07-26
Use a pinned workspace in Slice layer

benson31:
@tbennun identified a costly memcpy that might be improved by using pinned memory. This attempts to implement this.


tbennun (approved):
LGTM if tests pass

benson31:
CI matches develop.

PR #2294, opened 2023-07-27
Enable async memcpy

fiedorowicz1:
Re-enables asynchronous memcpy and adds a synchronization event to prevent a race condition. Also moves a data pre-fetching call to allow backwards kernels to be queued on the GPU first.


ndryden (approved):
Looks okay to me.

benson31 (approved):
LGTM. I'm good if it "passes" the "CI" "tests".

fiedorowicz1:
It's "passing" all the same "tests" as before on "Lassen."

PR #2295, opened 2023-07-31
Implement the Select operator

tbennun:
Implements the Select operator for setting a value based on a predicate tensor (and some tests). This has shown to improve performance on some models (second PR that uses this will follow).


fiedorowicz1 (approved):
Other than the one comment, LGTM.

tbennun:
@benson31 I'm awaiting your input since this is a new operator but otherwise ready to merge.

benson31 (approved):
LGTM

PR #2296, opened 2023-08-01
Rename percent_of_data_to_use

fiedorowicz1:
`percent_of_data_to_use` is inaccurately named and causes confusion as a value of 1.0 actually corresponds to 100%. This PR renames this to `fraction_of_data_to_use`. It is passing all of the same tests as before on Lassen.


tbennun (approved):

benson31 (approved):
LGTM. If CI's (as) happy (as ever), I'm (as) happy (as ever).

PR #2304, opened 2023-08-16
Add affine transformation weights to Layer Normalization

tbennun:
Please also double check that I did not make any mistakes adding those weights (regarding zeroing them out etc.); this is my first time adding weights to layers.


ndryden (approved):
This looks okay to me if it passes testing.

benson31 (approved):

PR #2312, opened 2023-08-30
Improve argparse error reporting

ndryden:
This now explicitly checks whether an option exists and throws a descriptive LBANN exception if not. Previously, the `std::unordered_map::at` would throw a very opaque error.

### DIFF
--- a/include/lbann/utils/argument_parser.hpp
+++ b/include/lbann/utils/argument_parser.hpp
@@ -28,6 +28,7 @@
 #define LBANN_UTILS_ARGUMENT_PARSER_HPP_INCLUDED
 
 #include "lbann/utils/environment_variable.hpp"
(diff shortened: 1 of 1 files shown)


include/lbann/utils/argument_parser.hpp
```diff
@@ -652,6 +653,9 @@ template <typename T>
 inline T const&
 argument_parser<ErrorHandler>::get(std::string const& option_name) const
 {
+  if (!option_is_defined(option_name)) {
+    LBANN_ERROR("Invalid option", option_name);
```
benson31, line 657:
```suggestion
    LBANN_ERROR("Invalid option: ", option_name);
```

benson31 (approved):

PR #2313, opened 2023-08-30
Improve Caliper/Profiling support

ndryden:
This improves our support for Caliper, and for profiling in general:
* The `prof_stop` function is now actually called by the profiler callback, so profilers can clean up.
* Adiak takes a dup of our world comm so it can gather MPI information.
* Caliper is now configurable via a new `--caliper_config` command-line option.
* The Python frontend now has support for `--profile`, `--profile-init`, and `--caliper-config` options, which do what you think they do.
* The ResNet, CosmoFlow, and Transformer applications have been updated to make use of these.


benson31 (approved):
:shipit:

applications/nlp/transformer/train.py
```diff
@@ -226,8 +226,9 @@ def make_batch_script(trainer_params, model_params, script_params, args):
             lbann.CallbackProgressBar()
         )
 
-    if args.profiling:
-        model.callbacks.append(lbann.CallbackProfiler())
+    profiler = create_profile_callback(args)
```
tbennun, line 229: where is this imported from?

tbennun (changes requested):
One comment on the Python side (please use some module path for that function)

fiedorowicz1 (approved):
Looks good other than @tbennun's comment

applications/nlp/transformer/train.py
ndryden: Fixed. Thanks!

tbennun (approved):

PR #2315, opened 2023-08-31
Move some annotations for Caliper around

benson31:
This is the start of using loop annotations in Caliper.

This approach is a bit clunky, but keeps us in Caliper "macro-land" -- on another branch I'm playing with using the Caliper C++ API directly, which requires less restructuring (which I might still do since I think it's a bit cleaner anyway) but more book-keeping. Either approach could be done by tomorrow, it's just a matter of which approach we prefer.

### DIFF
--- a/python/lbann/contrib/args.py
+++ b/python/lbann/contrib/args.py
@@ -191,12 +191,11 @@ def add_profiling_arguments(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         '--profile-init', action='store_true', default=False,
         help='enable profiling initialization')
-    if lbann.has_feature('CALIPER'):
-        parser.add_argument('--caliper', action='store_true', default=False,
-                            help='enable Caliper')
-        parser.add_argument(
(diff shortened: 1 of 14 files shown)


python/lbann/contrib/args.py
```diff
@@ -235,11 +241,15 @@ def get_profile_args(args: argparse.Namespace) -> list[str]:
...
     try:
+        caliper = args.caliper
         caliper_config = args.caliper_config
     except AttributeError:
         raise ValueError('passed arguments have not been processed by '
                          '`add_profiling_arguments`')
 
-    if caliper_config:
-        return ['--caliper_config', f'"{caliper_config}"']
+    if lbann.has_feature("CALIPER"):
```
ndryden, line 250: Can we print a warning if the user requests Caliper options but LBANN doesn't support them (to avoid silently dropping them)?

ndryden (approved):

bvanessen (approved):

PR #2317, opened 2023-09-07
Use Caliper's default-generated output name

ndryden:
We were using `lbann.cali` as the Caliper output filename (when not overridden). However, this means multiple runs will cause Caliper to overwrite its prior output. Using the filename it generates by default (which is a timestamp) will avoid this.


benson31 (approved):
🚢 🇮🇹

PR #2321, opened 2023-09-11
Stop error signal backpropagation

tbennun:
Stops error signal backpropagation if layer is tagged with `PROPAGATE_NOTHING` backpropagation requirements. Disabling computations can be disabled via the `LBANN_NO_BACKPROP_DISABLE`.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -1642,9 +1642,16 @@ void model::backward_prop(bool compute_weight_grads_only)
     }
 
     // Add parents to disabled layers if this layer is disabled
-    if (!enable_layer) {
-      for (auto& parent : l.get_parent_layers()) {
-        disabled_layers.insert(parent);
+    if (envvar_disable_layers) {
+      // Start propagating disabled layers up
+      if (l.get_backprop_requirements() == PROPAGATE_NOTHING)
+        enable_layer = false;
+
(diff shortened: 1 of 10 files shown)


src/models/model.cpp
```diff
@@ -1625,10 +1636,25 @@ void model::backward_prop(bool compute_weight_grads_only)
...
+      // Start propagating disabled layers up
+      if (l.get_backprop_requirements() == PROPAGATE_NOTHING)
+        enable_layer = false;
+
+      // Either start or continue propagating disabled layers
+      if (!enable_layer) {
+        for (auto& parent : l.get_parent_layers()) {
+          disabled_layers.insert(parent);
+        }
+      }
```
fiedorowicz1, line 1655: I think this would disable all parent layers even if the parent has two children one who has backprop disabled and one who does not.

fiedorowicz1 (changes requested):

src/models/model.cpp
tbennun: @fiedorowicz1 Fixed.

fiedorowicz1 (approved):

tbennun:
@benson31 @bvanessen ready for review, added a test to verify the behavior.

bvanessen (approved):

PR #2322, opened 2023-09-15
Clean up Caliper annotations and Adiak data

benson31:
This moves Caliper annotations to group by layer type in the tree. It also adds the total batch and epoch counts to Adiak.

The Adiak bits should be considered a hack. At the very least, there's an obvious flaw if SGD is invoked multiple times per run. However, I'm not going to add any new callbacks until I reassociate them with the training algorithm rather than the model. This will be dealt with then.

### DIFF
--- a/src/execution_algorithms/sgd_training_algorithm.cpp
+++ b/src/execution_algorithms/sgd_training_algorithm.cpp
@@ -40,6 +40,11 @@
 
 #include "lbann/proto/training_algorithm.pb.h"
 
+// FIXME (trb): This is a quick hack for a short-term deliverable.
+#ifdef LBANN_HAS_CALIPER
+#include <adiak.hpp>
+#endif
+
 #include <iostream>
 #include <memory>
 #include <string>
@@ -114,6 +119,7 @@ void SGDTrainingAlgorithm::train(SGDExecutionContext& c,
   do_train_begin_cbs(model, ScopeTimer{train_timer, "train_begin callbacks"});
 
   // Start iterating
+  size_t total_batches = 0;
   c.start_timer();
   LBANN_CALIPER_LOOP_BEGIN(train_epoch, "train_epoch");
   for (size_t epoch = 0; !term(c); ++epoch) {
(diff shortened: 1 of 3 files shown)


applications/vision/resnet.py
```diff
@@ -166,6 +166,7 @@
 else:
     lbann_args = ['--use_data_store', '--preload_data_store', '--node_sizes_vary']
 lbann_args += lbann.contrib.args.get_profile_args(args)
+lbann_args += ['--data_reader_fraction=0.01']
```
ndryden, line 169: This debugging aid should probably be removed?

model_zoo/lbann.cpp
```diff
@@ -96,6 +96,15 @@ int main(int argc, char* argv[])
...
 
+#ifdef LBANN_HAS_ROCM
+  {
+    int device, pci;
+    static_cast<void>(hipGetDevice(&device));
+    static_cast<void>(hipDeviceGetAttribute(&pci, hipDeviceAttributePciBusId, device));
+    std::cout << "RANK " << comm->get_rank_in_world() << ": gpu 0x" << std::hex << pci << std::dec << std::endl;
+  }
+#endif
+
```
ndryden, line 107: Is this a debugging aid or something we want to keep?

python/lbann/contrib/lc/launcher.py
```diff
@@ -91,8 +91,10 @@ def prepend_environment_path(key, prefix):
     # Optimizations for Tioga
     if system in ('tioga', 'rzvernal'):
         #set_environment('NCCL_SOCKET_IFNAME', 'hsi')
-        set_environment('MIOPEN_DEBUG_DISABLE_FIND_DB', '1')
-        set_environment('MIOPEN_DISABLE_CACHE', '1')
+        #set_environment('MIOPEN_DEBUG_DISABLE_FIND_DB', '1')
+        #set_environment('MIOPEN_DISABLE_CACHE', '1')
+        set_environment('MIOPEN_CUSTOM_CACHE_DIR', '${TMPDIR}/miopen')
+        set_environment('MIOPEN_USER_DB_PATH', '${TMPDIR}/miopen')
```
ndryden, line 97: Ditto.

src/models/model.cpp
```diff
@@ -2091,26 +2091,26 @@ void model::save_model()
...
+//   if (m_comm->am_world_master()) {
+//     std::cout << "\nDistconv-enabled layers:\n\t" << dc_enabled.str() << "\n\n"
+//               << "Distconv-disabled layers:\n\t" << dc_disabled.str() << "\n\n";
+
+// #ifdef LBANN_DEBUG
+//     std::cout << "Parallel Strategy description - ";
+//     print_parallel_strategy_header(std::cout);
+// #endif
+//     endl(std::cout);
+//   }
```
ndryden, line 2113: Ditto, though it might be more fine to remove this. ;)

ndryden (changes requested):
Looks like some debugging stuff got committed too.

benson31:
@ndryden This was a `git commit -a` error; the commit has been revised.

ndryden (approved):

PR #2325, opened 2023-09-21
Fixed multiple bugs in data store

bvanessen:
Fixed a bug in the data store where the call to get_empty_node was not
using a proper mutex for the m_data structure.

Added additional instances of locking to avoid thread race conditions.

Fixed shared memory segment so that only rank 0 on a node will try to unlink it.

### DIFF
--- a/src/data_store/data_store_conduit.cpp
+++ b/src/data_store/data_store_conduit.cpp
@@ -131,11 +131,11 @@ data_store_conduit::~data_store_conduit()
   if (m_profile) {
     m_profile->close();
   }
-  if (m_is_local_cache && m_mem_seg) {
+  if (m_is_local_cache && m_mem_seg && m_comm->get_rank_in_node() == 0) {
     int sanity = shm_unlink(m_seg_name.c_str());
     if (sanity != 0) {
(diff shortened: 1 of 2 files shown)


src/data_store/data_store_conduit.cpp
```diff
@@ -131,11 +131,11 @@ data_store_conduit::~data_store_conduit()
...
   }
-  if (m_is_local_cache && m_mem_seg) {
+  if (m_is_local_cache && m_mem_seg && m_comm->get_rank_in_node() == 0) {
     int sanity = shm_unlink(m_seg_name.c_str());
     if (sanity != 0) {
       std::cout << "\nWARNING: shm_unlink failed in "
-                   "data_store_conduit::~data_store_conduit()\n";
+	"data_store_conduit::~data_store_conduit() for name " << m_seg_name.c_str() << "\n";
     }
     sanity = munmap(reinterpret_cast<void*>(m_mem_seg), m_mem_seg_length);
```
tbennun, line 140: should this be done on all ranks? also, before shm_unlink?

tbennun (changes requested):
Unlink process might be buggy

tbennun (approved):

PR #2328, opened 2023-09-26
HDF5 Data reader enhancements

bvanessen:
Added a check in the HDF5 data reader to check that the metadata for
each field actually matches the dimensions of the data fields.

Added a helper function for conduit to allow the calculation of a product of a data array's elements.

### DIFF
--- a/include/lbann/utils/conduit_extensions.hpp
+++ b/include/lbann/utils/conduit_extensions.hpp
@@ -25,6 +25,8 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include "conduit/conduit_data_array.hpp"
+#include "conduit_relay_io_hdf5.hpp"
+#include "hdf5.h"
 
 namespace conduit {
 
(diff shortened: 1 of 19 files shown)


src/data_readers/data_reader_HDF5.cpp
```diff
@@ -357,7 +358,23 @@ void hdf5_data_reader::load_sample(conduit::Node& node,
                                       original_path,
                                       node[new_pathname]);
       }
-
+      // Check that the dimensions of each node matches its metadata
+      if (metadata.has_child(HDF5_METADATA_KEY_DIMS)) {
+        int n_elts = node[pathname].dtype().number_of_elements();
+        conduit::int64_array data_array_dims = metadata[HDF5_METADATA_KEY_DIMS].value();
+        auto expected_n_elts = data_array_prod(data_array_dims);
```
benson31, line 365:
```suggestion
        int const n_elts = node[pathname].dtype().number_of_elements();
        conduit::int64_array const data_array_dims = metadata[HDF5_METADATA_KEY_DIMS].value();
        auto const expected_n_elts = data_array_prod(data_array_dims);
```

include/lbann/utils/conduit_extensions.hpp
```diff
@@ -0,0 +1,45 @@
...
+    for(index_t i = 0; i < a.number_of_elements(); i++)
+    {
+        const T &val = a.element(i);
+        res *= val;
+    }
+
+    return res;
+}
+
+} // conduit
```
benson31, line 45:
```suggestion
namespace lbann {

template <typename T>
T
data_array_prod(DataArray<T> const& a)
{
    T res = 1;
    for(index_t i = 0; i < a.number_of_elements(); i++)
    {
        const T &val = a.element(i);
        res *= val;
    }

    return res;
}

} // namespace lbann
```
IMO we shouldn't be extending their namespace; just put it in lbann. Also, conduit::DataArray doesn't own any data, so it's just a "shallow copy" (pointer and size), but the "expected" API would probably be pass-by-const-reference. (I'm moderately surprised that they don't expose begin() and end() functions (and suitable iterators) for this container!)

src/data_readers/data_reader_HDF5.cpp
```diff
@@ -357,7 +358,23 @@ void hdf5_data_reader::load_sample(conduit::Node& node,
...
+
+        if (n_elts != expected_n_elts) {
+          LBANN_WARNING("Ingesting sample field ",
+                        pathname,
+                        " for sample ",
+                        sample_name,
+                        " where the dimensions in the metadata don't match the actual field: ",
+                        expected_n_elts,
+                        " != ",
+                        n_elts);
```
benson31, line 375: Might be good to highlight in the output which is the observed vs the expected value -- shouldn't have to look in source code to decipher warnings...

PR #2330, opened 2023-09-28
Fix Data Coordinator Map Lookup

fiedorowicz1:
Fix to handle the case where the data field is not in the map of active data fields.


benson31 (approved):
:shipit:

bvanessen (approved):

PR #2332, opened 2023-09-29
Reduce reliance on `data_*` throughout LBANN (1/?)

tbennun:


src/callbacks/summarize_images.cpp
```diff
@@ -230,69 +230,8 @@ build_categorical_accuracy_strategy_from_pbuf(
...
-      if (m_tracked_images.find(sample_index) != m_tracked_images.end()) {
-        img_indices.push_back(std::make_pair(ii, sample_index));
-      }
-      else if (m_tracked_images.size() < m_num_images) {
-        m_tracked_images.insert(sample_index);
-        img_indices.push_back(std::make_pair(ii, sample_index));
-      }
-    }
-  }
+  LBANN_ERROR("This feature is no longer supported");
```
bvanessen, line 234: @tbennun Is this just to clean it up or is there something that is broken in this one or by this PR.

tbennun: It is just cleanup, nothing broken here.

benson31: So I recognize that this is extremely hacky code, but do we just want to remove this capability?

benson31 (approved):

src/callbacks/summarize_images.cpp
tbennun: we can bring it back to the codebase and get rid of it in a callback cleanup PR `(4/?)`?

PR #2338, opened 2023-10-03
Bugfix user build

bvanessen:
Fixes the cache issues on Tioga by reverting back to `/tmp` and adds some helping commands to avoid building duplicate spack packages that are externally available.

### DIFF
--- a/ci_test/common_python/tools.py
+++ b/ci_test/common_python/tools.py
@@ -753,16 +753,16 @@ def create_tests(setup_func,
         else:
             environment = {}
 
-        if os.environ.get('USER') == 'lbannusr':
-            basepath = '/p/vast1/lbannusr'
-        else:
-            basepath = '/p/vast1/lbann'
+        # if os.environ.get('USER') == 'lbannusr':
+        #     basepath = '/p/vast1/lbannusr'
+        # else:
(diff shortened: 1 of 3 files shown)


ci_test/common_python/tools.py
```diff
@@ -753,16 +753,16 @@ def create_tests(setup_func,
         else:
             environment = {}
 
-        if os.environ.get('USER') == 'lbannusr':
-            basepath = '/p/vast1/lbannusr'
-        else:
-            basepath = '/p/vast1/lbann'
+        # if os.environ.get('USER') == 'lbannusr':
```
tbennun, line 756: We should add a comment as to why we commented this out.

tbennun (changes requested):
Minor change

ci_test/common_python/tools.py
benson31, line 1: Why is there a change in `ci_test` when the PR is to fix the user build?

ci_test/common_python/tools.py
bvanessen: I had the comment in there `        # Empirically the cache dir cannot be on a parallel file system`

benson31 (approved):
LGTM

PR #2342, opened 2023-10-04
Initial layer parallelism

benson31:
This is currently distinct from "subgraph parallelism", but the notion is a bit simpler.

The implementation crux of this functionality is that every `Layer` in the model manages its own data completely internally (weights, prev_activations, activations). At layer boundaries, when a layer and its parent are on different grids, a deep copy will be done explicitly to move the data from `parent_layer.grid` to `this->grid`.

See the blurb I put on Confluence for some other discussion and assumptions.

### DIFF
--- a/src/callbacks/check_dataset.cpp
+++ b/src/callbacks/check_dataset.cpp
@@ -65,15 +65,12 @@ void check_dataset::add_to_set(model* m,
     return;
   }
 
-  // FIXME (trb 10/03/2023): This is not the "right" fix (which is
-  // probably to reconsider this callback entirely, but this is the
-  // drop-in replacement for the line that was here
-  // (l->get_sample_indices_per_mb(), which just returns 'nullptr'
-  // (which is never checked, of course, so the loop below would
-  // segfault almost instantly))).
+  // FIXME (trb 10/03/2023): It is probably best to reconsider this
+  // callback entirely.
   auto const mode = m->get_execution_context().get_execution_mode();
   El::Matrix<El::Int> const* const indices =
     get_trainer().get_data_coordinator().get_sample_indices_per_mb(mode);
+  LBANN_ASSERT((bool) indices);
 
   for (El::Int j = 0; j < indices->Width(); j++) {
     for (El::Int i = 0; i < indices->Height(); i++) {
--- a/src/layers/data_type_layer.cpp
+++ b/src/layers/data_type_layer.cpp
@@ -167,9 +167,6 @@ modify_reference_counter(PointerRangeReferenceCounter& refcnt,
 template <typename InputTensorDataType, typename OutputTensorDataType>
 void data_type_layer<InputTensorDataType, OutputTensorDataType>::forward_prop()
 {
-  if (!this->is_participating())
-    return;
-
   // This bit is preprocessed out since the LBANN_CALIPER macro
   // won't help us out here.
 #ifdef LBANN_HAS_CALIPER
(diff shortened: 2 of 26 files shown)


src/layers/learning/fully_connected.cpp
```diff
@@ -823,12 +823,18 @@ void fully_connected_layer<T, L, D>::write_specific_proto(
 template <typename TensorDataType, data_layout T_layout, El::Device Dev>
 void fully_connected_layer<TensorDataType, T_layout, Dev>::fp_compute()
 {
+  if (!this->weights_values(0).Participating())
+    return;
```
tbennun, line 827: will we have to do this in every layer? shouldn't this just be in `class layer`/`class model`? If not, what makes FullyConnected a special case?

tbennun (commented):
Some comments and questions. In general, where is the code that skips the computation of a layer if it's not participating in the grid with the new grid tag?

src/layers/data_type_layer.cpp
```diff
@@ -830,29 +782,104 @@ void data_type_layer<InputTensorDataType, OutputTensorDataType>::setup_matrices(
...
   else {
+    this->do_setup_matrices_simple(grid);
+  }
+}
 
-    for (auto& input : m_inputs) {
-      input = input_mat_builder->MakeEmpty(grid, 0);
-    }
+template <typename InT, typename OutT>
+void data_type_layer<InT, OutT>::do_setup_matrices_simple(El::Grid const& grid)
```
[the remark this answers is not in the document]
benson31: It's not "subgraph" ;) And also, it's shorter.

src/layers/learning/fully_connected.cpp
benson31: I would prefer to make `Layer::forward_prop()` a "template method"-type driver function that can skip it entirely:
```
Layer::forward_prop() {
  if (not this->is_participating())
    return;
  // maybe some other common functionality (profiling annotations?)
  // finally, layer-specific bits via the "fp_compute" worker function they already implement:
  fp_compute();
  // maybe some common cleanup?
}
```
I'm evaluating whether this pattern actually works for the current set of layers in LBANN (in the worst case, I could make `forward_prop` a `virtual` function with a comment that it should only be overriden in exceptional cases where the developer is very aware of what they're doing, thereby shifting the onus onto them to replicate the common bits/participation bits as appropriate.

src/callbacks/check_dataset.cpp
```diff
@@ -65,28 +65,25 @@ void check_dataset::add_to_set(model* m,
...
-                              std::to_string(idx));
-      }
-      else {
-        set.insert(idx);
+  // FIXME (trb 10/03/2023): This is not the "right" fix (which is
+  // probably to reconsider this callback entirely, but this is the
+  // drop-in replacement for the line that was here
+  // (l->get_sample_indices_per_mb(), which just returns 'nullptr'
+  // (which is never checked, of course, so the loop below would
+  // segfault almost instantly))).
```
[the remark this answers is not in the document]
benson31: Fixed. Thanks.

src/layers/learning/fully_connected.cpp
benson31: I'm updating `data_type_layer::forward_prop` and `data_type_layer::back_prop_impl_` to try to figure this out.

tbennun (approved):

bvanessen (approved):

PR #2343, opened 2023-10-05
Refactor models to cleanup how mini-batch size is propagated

bvanessen:
Mini-batch size is now propagated between layers via size of the output matrices.


src/models/model.cpp
```diff
@@ -733,6 +735,10 @@ void model::setup(size_t max_mini_batch_size,
     setup_subcommunicators(grids_);
   }
 
+  // Because it is used as a default for layers with no inputs
+  // initailize the current mini-batch size before setting up the
```
tbennun, line 739:
```suggestion
  // initialize the current mini-batch size before setting up the
```

tbennun (approved):
Two minor comments but otherwise LGTM

PR #2351, opened 2023-10-11
Move check for layer construction failure before first use

ndryden:
Otherwise if construction fails, we segfault.

### DIFF
--- a/src/proto/factories/layer_graph_factory.cpp
+++ b/src/proto/factories/layer_graph_factory.cpp
@@ -172,6 +172,11 @@ construct_layer_graph(lbann_comm* comm,
 
 #undef TEMPLATE_INSTANTIATION
 
+    // Check that layer has been constructed
+    if (l == nullptr) {
+      LBANN_ERROR("Could not construct layer ", name);
+    }
+
     // Set up parallel strategy.
     ParallelStrategy& ps = l->get_parallel_strategy();
     ps.sample_groups = proto_layer.parallel_strategy().sample_groups();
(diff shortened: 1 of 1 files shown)


src/proto/factories/layer_graph_factory.cpp
```diff
@@ -172,6 +172,11 @@ construct_layer_graph(lbann_comm* comm,
 
 #undef TEMPLATE_INSTANTIATION
 
+    // Check that layer has been constructed
```
tbennun, line 175: Not sure about this part

src/proto/factories/layer_graph_factory.cpp
```diff
@@ -172,6 +172,11 @@ construct_layer_graph(lbann_comm* comm,
...
+    if (l == nullptr) {
+      LBANN_ERROR("Could not construct layer ", name);
```
tbennun, line 177: Why is that?

src/proto/factories/layer_graph_factory.cpp
```diff
@@ -188,10 +193,6 @@ construct_layer_graph(lbann_comm* comm,
     ps.depth_groups = proto_layer.parallel_strategy().depth_groups();
     ps.depth_splits = proto_layer.parallel_strategy().depth_splits();
 
-    // Check that layer has been constructed
-    if (l == nullptr)
-      LBANN_ERROR("Could not construct layer ", name);
-
```
tbennun, line 194: Might want to check that a second time here too, just in case.

tbennun (changes requested):

src/proto/factories/layer_graph_factory.cpp
ndryden: Why? The pointer can't change at this point.

src/proto/factories/layer_graph_factory.cpp
ndryden: If it's null, it means the instantiation attempts (generated by macros) failed, as the default value of `l` is `nullptr`.

tbennun (approved):

ndryden:
Thanks @tbennun, I know this was a complicated PR.

PR #2355, opened 2023-10-17
Double data type optional

bvanessen:
Changing the core of LBANN to have double as a core data type for
layers, weights, and optimizers to be a build time option.  This will 
allow developers to reduce compilation time (minimizing template 
instantiation).


tbennun (approved):
Does it mean that EvalType will stay double in any case?

bvanessen:
> Does it mean that EvalType will stay double in any case?

Yes for now.  We are trying not to perturb any tests, just reduce compilation time.

benson31 (approved):

PR #2358, opened 2023-10-19
Added cleanup call to remove old object files.

bvanessen:


.gitlab/common/common.yml
```diff
@@ -120,6 +120,7 @@
     - find ${CI_PROJECT_DIR}/builds -maxdepth 1 -mtime +2 -name "lbann_*" -type d -print -exec rm -r {} \;
     - find ${CI_PROJECT_DIR} -maxdepth 1 -mtime +2 -name "LBANN_*_setup_build_tools.sh" -type f -print -exec rm {} \;
     - find ${CI_PROJECT_DIR} -maxdepth 1 -mtime +2 -name "LBANN_*.cmake" -type f -print -exec rm {} \;
+    - find ${CI_PROJECT_DIR} -maxdepth 3 -mtime +1 -type d -path "*/builds/lbann_*/build" -print -exec cmake --build {} -t clean \;
```
benson31, line 123: Why not just delete the directory? CMake configuration/generation time is a negligible portion of our CI runtime...

bvanessen: I was thinking that we should keep those so that we can see what was happening in the catch2 tests within the last day if there is a failure.

benson31: What do you mean by "see what was happening"? If you want those binaries around to be able to rerun them, e.g., you shouldn't `clean` the build dir (line 121 above)

bvanessen: @benson31 Hmmm, good point.  I will change it to remove all build directories more than 2 days old.

benson31 (approved):

PR #2359, opened 2023-10-19
Automatic mixed precision

ndryden:
This is an initial pass at AMP. It appears correct in the cases I've tested.

It has some particularly pointy bits in batchnorm and in gradient unscaling.


src/utils/cudnn.cpp
```diff
@@ -1315,7 +1332,7 @@ AlgoType find_best_heuristic_algorithm(
     if (p.memory > max_ws_size) {
       continue;
     }
-    algos.push_back(p.algo);
+    algos.push_back(cudnn_alg_config<AlgoType>(p.algo, p.mathType));
```
benson31, line 1335: Does this work?
```suggestion
    algos.emplace_back({p.algo, p.mathType});
```
Might be better?

benson31 (commented):
I'm still working on the changes `model.cpp` and the `amp.{cpp,cu}` files. I suppose I should also read the `.py` file too. But just a few small things.

applications/vision/lenet.py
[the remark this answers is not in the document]
ndryden: It's also behind a command-line flag (AMP is only enabled with `--amp`), so I think this is fine.

include/lbann/layers/regularizers/batch_normalization.hpp
```diff
@@ -49,6 +49,25 @@ enum class batch_normalization_stats_aggregation
...
+template <typename TensorDataType>
+struct BNAccT {
+  using type = TensorDataType;
+};
+#ifdef LBANN_HAS_GPU_FP16
+template <>
+struct BNAccT<fp16> {
+  using type = float;
+};
+#endif
```
benson31, line 69: I recommend also adding a typedef to force the `::type`, which will save you some characters later on:
```
template <typename T>
using BNAcc = typename BNAccT<T>::type;
```
Now you can use `BNAcc<T>` anywhere you've used `typename BNAccT<T>::type` downstream (I notice a lot in the `.cu` file, e.g.).

src/utils/amp.cu
```diff
@@ -0,0 +1,148 @@
...
+__host__ __device__ __forceinline__ bool amp_is_finite(T val) {
+  return std::isfinite(val);
+}
+
+#ifdef LBANN_HAS_GPU_FP16
+template <>
+__host__ __device__ __forceinline__ bool amp_is_finite<fp16>(fp16 val) {
+  return std::isfinite(static_cast<float>(val));
+}
+#endif
```
benson31, line 50:
```suggestion
template <typename T>
__host__ __device__ __forceinline__ bool amp_is_finite(T val) {
  return lbann::gpu_lib::isfinite(val);
}
```
Or just use this function throughout. Don't forget to `#include <lbann/utils/gpu/gpu_lib.hpp>`.

benson31 (approved):

src/utils/amp.cu
ndryden: Done. Also _do_ forget to `#include "lbann/utils/gpu/gpu_lib.hpp"`, apparently that header is not meant to be included directly.

benson31 (approved):
Still looking good after the rebase/recent commits. I touched the memory_profile callback and added an example of how to use the `SwitchDispatcher`. I found it a bit clunkier than I remembered to use, but I think I have some ideas to improve it in future H2 versions (in particular, I want to make it possible to just use a lambda as the functor).

Also, please `git clang-format` the whole PR.

benson31 (approved):

PR #2376, opened 2023-11-02
Optimize gradient norm clipping

tbennun:
- [x] Reduces communication when calling global gradient norm clipping
- [x] Runs norm computation on the GPU (requires latest Hydrogen develop branch)


fiedorowicz1:
Needs to handle case where some weights are sharded and some are not. Otherwise, unsharded weights will be overcounted. Could be fixed by multiplying by an appropriate scaling factor or keeping separate sharded/unsharded sums.

benson31 (approved):
LGTM

fiedorowicz1 (approved):
Looks good! I recommend noting how "global" is defined in this case.

ndryden (approved):

PR #2377, opened 2023-11-03
Refactor async io to use uint64_t

bvanessen:
Cleanup the data type for the mini-batch, indices, and position in
data set fields.  Switched from a mish-mash of int, size_t, and
El::Int to uint64_t for consistency.


benson31 (approved):
I assume this is passing CI (or whatever counts as "passing")?

My primary concern is the number of `int`->`uint64_t` changes -- I did not verify that every arithmetic computation in all affected files are still valid with unsigned integer types. If it "passes" CI, though, that's a fair indicator things are still ok.

PR #2396, opened 2023-12-04
Fixes for half and CUDA 12.2

ndryden:
Various fixes throughout the codebase to build properly with the changes to half in CUDA 12.2.

Mainly this is now better handling CPU code that is instantiated with the GPU half type (even if we never execute it).

(I will wait to merge this until CI can run.)

### DIFF
--- a/include/lbann/layers/regularizers/local_response_normalization.hpp
+++ b/include/lbann/layers/regularizers/local_response_normalization.hpp
@@ -299,7 +299,7 @@ private:
 
     // Check if LRN is using default beta parameter
     // std::fabs is not defined for GPU half.
-#ifdef LBANN_HAS_HALF
+#ifdef LBANN_HAS_GPU_FP16
     const bool default_beta = [&] {
       if constexpr (!std::is_same_v<TensorDataType, __half>) {
         return (std::fabs((m_beta - El::To<TensorDataType>(0.75)) /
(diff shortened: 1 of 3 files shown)


tbennun (approved):

include/lbann/layers/regularizers/local_response_normalization.hpp
```diff
@@ -298,10 +298,25 @@ class local_response_normalization_layer
     const int num_per_channel = this->get_output_size() / num_channels;
 
     // Check if LRN is using default beta parameter
+    // std::fabs is not defined for GPU half.
+#ifdef LBANN_HAS_HALF
```
benson31, line 302:
```suggestion
#ifdef LBANN_HAS_GPU_FP16
```

include/lbann/layers/regularizers/local_response_normalization.hpp
```diff
@@ -298,10 +298,25 @@ class local_response_normalization_layer
...
+        LBANN_ERROR("Cannot use GPU half on CPU for LRN");
+        return false;
+      }
+    }();
+#else
     const bool default_beta =
       (std::fabs((m_beta - El::To<TensorDataType>(0.75)) /
                  El::To<TensorDataType>(0.75)) <
        2 * std::numeric_limits<DataType>::epsilon());
+#endif // LBANN_HAS_HALF
```
benson31, line 319:
```suggestion
#endif // LBANN_HAS_GPU_FP16
```

benson31 (changes requested):

src/callbacks/gradient_clipping.cpp
```diff
@@ -158,7 +158,7 @@ struct NormComputer
         }
 #else
         local_norm = El::Nrm2(gradmatrix);
-#endif
+#endif // LBANN_HAS_HALF
```
benson31, line 161:
```suggestion
#endif // LBANN_HAS_GPU_FP16
```
Also, the `#ifdef` line above should be changed accordingly.

benson31 (approved):

PR #2397, opened 2023-12-05
Python updates/cleanup for benchmarking

ndryden:
This is a cleaned up version of some updates that were requested for benchmarking.

* Better support `--setup-only` for GPT.
* Support manually specifying a parallel command launcher (in case our auto-selection gets it wrong) via `--launcher`.
* Support specifying additional arguments to pass to a parallel command launcher via `--launcher-args`.


tbennun (approved):

bvanessen (approved):
Looks good.  Once it runs through CI, I think that it is good to go.

PR #2400, opened 2023-12-07
Enabled the CosmoFlow integration test to run in the distconv CI

bvanessen:
branches.  Updated the CosmoFlow target metrics.  Removed CI code to cleanup build directorys that would have spurious failures when the downstream CI tasks ran with different agents.  Additionally, it became unnecessary when CI was allowed to cleanup the entire directory.


fiedorowicz1 (approved):
LGTM

PR #2403, opened 2023-12-12
Fix pooling workspace behavior on MIOpen

tbennun:


src/layers/transform/pooling.cpp
```diff
@@ -114,15 +114,16 @@ void pooling_layer<TensorDataType, Layout, Device>::fp_compute_dnn()
...
   auto& local_output = this->get_local_activations();
   if (local_input.Height() > 0 && local_input.Width() > 0) {
     // Initialize GPU workspace
-    El::Matrix<TensorDataType, El::Device::GPU> workspace;
     size_t workspace_size =
       dnn_lib::get_pooling_ws_size(m_pooling_dnn_desc,
-                                  m_tensors_dnn_desc.get_activations());
-    workspace.Resize(workspace_size / sizeof(TensorDataType), 1);
+                                   m_tensors_dnn_desc.get_activations());
+    m_workspace.reset(new MatrixType());
```
tbennun, line 125: I don't like this behavior (I'd rather keep a workspace matrix on storage mode 1 and Empty/Resize it), but this is what used to exist before.

ndryden (approved):
We should get a ticket or something open to improve the relevant docs.

bvanessen (approved):
LGTM

src/layers/transform/pooling.cpp
```diff
@@ -174,7 +169,7 @@ void pooling_layer<TensorDataType, Layout, Device>::bp_compute_dnn()
                               zero,
                               m_tensors_dnn_desc.get_error_signals(),
                               local_gradient_wrt_input,
-                              workspace);
+                              dynamic_cast<MatrixType&>(*m_workspace));
```
benson31, line 172:
```suggestion
                              *m_workspace);
```

include/lbann/layers/transform/pooling.hpp
```diff
@@ -102,13 +102,12 @@ class pooling_distconv_adapter
...
 
-template <typename TensorDataType,
-          data_layout T_layout = data_layout::DATA_PARALLEL,
-          El::Device Dev = El::Device::CPU>
+template <typename TensorDataType, data_layout T_layout, El::Device Dev>
 class pooling_layer : public data_type_layer<TensorDataType>
 {
   static_assert(T_layout == data_layout::DATA_PARALLEL,
                 "pooling only supports DATA_PARALLEL");
+  using AbsMatrixType = El::AbstractMatrix<TensorDataType>;
```
benson31, line 110:
```suggestion
```

benson31 (changes requested):

include/lbann/layers/transform/pooling.hpp
```diff
@@ -136,6 +135,8 @@ class pooling_layer : public data_type_layer<TensorDataType>
   /** Tensor DNN library descriptors. */
   dnn_lib::data_parallel_layer_tensor_manager<TensorDataType>
     m_tensors_dnn_desc;
+  /** Workspace (shared between forward and backprop by MIOpen). */
+  std::unique_ptr<AbsMatrixType> m_workspace;
```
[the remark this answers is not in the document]
tbennun: > But you have `TensorDataType` and `Dev`.
> 
> If we don't care about a few extra bytes, the `unique_ptr` also isn't necessary.

We do (and that's how I implemented it in the first round), but it doesn't compile because of the .reset() calls (`Copy()` returns an abstractmatrix)

src/layers/transform/pooling.cpp
```diff
@@ -133,7 +134,7 @@ void pooling_layer<TensorDataType, Layout, Device>::fp_compute_dnn()
                              zero,
                              m_tensors_dnn_desc.get_activations(),
                              local_output,
-                             workspace);
+                             dynamic_cast<MatrixType&>(*m_workspace));
```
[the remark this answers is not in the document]
tbennun: same

include/lbann/layers/transform/pooling.hpp
tbennun: same

include/lbann/layers/transform/pooling.hpp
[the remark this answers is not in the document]
benson31: Well the `.reset` calls should be `= std::make_unique` calls. And the `->Copy` call isn't needed when you can just use the copy constructor `other.m_workspace ? std::make_unique<MatrixType>(*other.m_workspace) : nullptr`. So let's consider just moving the whole thing to an `El::Matrix<T, Device::GPU>` and dispense with this ptr nonsense. I'll give you a patch offline.

benson31 (approved):

ndryden (approved):

tbennun:
@benson31 should we use Resize instead of reassigning a new matrix?

PR #2404, opened 2023-12-12
Bugfix cleanup rng omp

bvanessen:


src/utils/random_number_generators.cpp
```diff
@@ -32,32 +32,50 @@
 #include <thread>
 
 namespace {
+#if not defined(LBANN_DETERMINISTIC) && defined(_OPENMP)
 #ifdef __ICC
 lbann::rng_gen generator;
 #pragma omp threadprivate(generator)
```
tbennun, line 38: shouldn't this line be removed as well?

src/utils/random_number_generators.cpp
```diff
@@ -32,32 +32,50 @@
...
 
 lbann::fast_rng_gen fast_generator;
 #pragma omp threadprivate(fast_generator)
```
tbennun, line 41: same

tbennun (commented):
Looks clean, but there is one comment about the ICC version of the globals

benson31 (approved):

src/utils/random_number_generators.cpp
bvanessen: @tbennun They actually need to be renamed to `OMP_generator`

src/utils/random_number_generators.cpp
bvanessen: @tbennun Good catch.

PR #2410, opened 2023-12-24
Fix TSE LTFB test when top-k generates multiple sends

tbennun:
This PR fixes the LTFB test in corner cases of different numbers of trainers, and adds more validity tests than there were before. 
It also improves TSE's behavior (reproducibility and correctness) if multiple trainers achieve the same tournament score by using a stable sort with indices instead of sorting on the tournament values.


bvanessen (approved):
LGTM

PR #2419, opened 2024-01-10
Fix model-parallel layer normalization

tbennun:
Missed global normalization size in the model-parallel case.


ndryden (approved):
Looks reasonable to me if it passes CI.

tbennun:
Passing the tests.

benson31 (approved):
LGTM

PR #2420, opened 2024-01-29
Support layer parallelism in transformer application

tbennun:
This PR adds the capability to support layer parallelism in transformers, variable-length version of The Pile pretokenized dataset, updates to the LBANN graph visualizer script, and some minor tweaks to weights layer.


bvanessen (approved):
LGTM

PR #2421, opened 2024-01-30
Restore the superbuild

benson31:
It is with great sadness that this is done. The superbuild has been kept for internal use and for those situations in which real package managers prove inadequate, which were supposed to be few. Yet I have had to hand it out to nearly every user and/or developer of LBANN, because I guess our simple little package is too much for certain package managers to handle. Thus, here it is.

Do note that I do not want to spend any more of my life maintaining build systems or hand-holding developers (or users) than is programmatically necessary. Therefore this is presented as-is, with a couple of example scripts and a README. Good luck.

Finally, do continue to relentlessly hound the developers and/or maintainers of whichever package manager has failed to provide you with a working LBANN build so that this superbuild can fade away for a final time. It's not a hard package to build, and the dependencies are both quite few and quite common. So it's a real shame that there is so much trouble.

### DIFF
--- a/scripts/superbuild/CMakeLists.txt
+++ b/scripts/superbuild/CMakeLists.txt
@@ -1,0 +1,180 @@
+################################################################################
+## Copyright (c) 2014-2022, Lawrence Livermore National Security, LLC.
+## Produced at the Lawrence Livermore National Laboratory.
+## Written by the LBANN Research Team (B. Van Essen, et al.) listed in
+## the CONTRIBUTORS file. <lbann-dev@llnl.gov>
+##
+## LLNL-CODE-697807.
+## All rights reserved.
+##
+## This file is part of LBANN: Livermore Big Artificial Neural Network
+## Toolkit. For details, see http://software.llnl.gov/LBANN or
+## https://github.com/LLNL/LBANN.
+##
+## Licensed under the Apache License, Version 2.0 (the "Licensee"); you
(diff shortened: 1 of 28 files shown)


ndryden (approved):
🥲

scripts/superbuild/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
+################################################################################
+## Copyright (c) 2014-2022, Lawrence Livermore National Security, LLC.
```
tbennun, line 2: Copyright year

scripts/superbuild/adiak/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+################################################################################
+## Copyright (c) 2014-2022, Lawrence Livermore National Security, LLC.
```
tbennun, line 2: Year

scripts/superbuild/aluminum/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
+################################################################################
+## Copyright (c) 2014-2022, Lawrence Livermore National Security, LLC.
```
tbennun, line 2:
```suggestion
## Copyright (c) 2014-2024, Lawrence Livermore National Security, LLC.
```

tbennun (approved):

PR #2425, opened 2024-02-14
Fix CosmoFlow Double-Reading

fiedorowicz1:
Fixes double-reading of samples for cosmoflow when the datastore is disabled.
Before:
image

After:
image


ndryden (approved):
Nice performance improvement!

PR #2435, opened 2024-04-15
Fix to PeriodicPadding to match pytorch circular padding

jvwilliams23:
Hi,

I found that the output of `lm.PeriodicPadding2D` did not match the output of `torch.nn.functional.pad(*, mode="circular")`. See below for figure comparing the original implementation - clearly the middle image is doing the equivalent of `mode = "replicate"` padding (not periodic/circular). I modified the existing CI test to compare against the torch function (it previously compared against padding done using np.concatenate).

lbann_periodicpadding_error

I also made some modifications to the docstring.

Cheers,
Josh

### DIFF
--- a/ci_test/unit_tests/test_unit_module_periodic_padding.py
+++ b/ci_test/unit_tests/test_unit_module_periodic_padding.py
@@ -6,6 +6,13 @@ import pytest
 from torch import Tensor
 import torch.nn.functional as F
 
+try:
+    from torch import Tensor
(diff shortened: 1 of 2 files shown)


fiedorowicz1 (commented):
Great catch, thanks for the PR! There are a couple of other things I caught that I will add to this PR.

fiedorowicz1 (approved):

bvanessen (approved):

PR #2440, opened 2024-04-26
Add aws-ofi-rccl plugin to the superbuild

benson31:
This should not be used during a regular user builds or the usual development cycle. Rather, much like the RCCL build, this should be built only when the default plugin is insufficient (debugging/testing). Of course your center provides a default version of this plugin, right? OLCF and NERSC do...


tbennun (approved):

bvanessen (approved):
Works, especially when you read the output error messages.

PR #2441, opened 2024-04-29
Add NCCL to the superbuild

benson31:
This is only for basic NCCL builds that don't require any modifications directly to the makefile. Only a few options are supported, and there's no support for customizing C/C++/CUDA flags outside of specifying a gencode argument to limit arch support.

### DIFF
--- a/scripts/superbuild/nccl/CMakeLists.txt
+++ b/scripts/superbuild/nccl/CMakeLists.txt
@@ -1,0 +1,174 @@
+################################################################################
+## Copyright (c) 2014-2022, Lawrence Livermore National Security, LLC.
+## Produced at the Lawrence Livermore National Laboratory.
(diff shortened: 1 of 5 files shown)


scripts/superbuild/nccl/CMakeLists.txt
```diff
@@ -0,0 +1,174 @@
+################################################################################
+## Copyright (c) 2014-2022, Lawrence Livermore National Security, LLC.
```
ndryden, line 2: Nit: 2024.

ndryden (approved):

bvanessen (approved):

scripts/superbuild/nccl/CMakeLists.txt
benson31: Fixed.

PR #2442, opened 2024-04-30
Fixes In-Place Activation Reference Counting

fiedorowicz1:
Fixes an issue reported by @alexandraday-nu with in-place layers. Previously, in-place activations were not having their reference counts incremented, leading to the activations prematurely being freed. This PR ensures that the reference count is properly incremented for in-place activations.


tbennun (approved):
Thank you, looks good!

PR #2460, opened 2024-06-28
Use CUTENSOR for Tessellate backpropagation in supported cases

tbennun:
In cases where all dimensions are only extended (i.e., broadcast) from 1 to some value, backpropagation of Tessellate can be represented as a multi-dimensional reduction. This can yield potential performance improvements.

### DIFF
--- a/src/layers/transform/tessellate.cpp
+++ b/src/layers/transform/tessellate.cpp
@@ -134,6 +134,7 @@ void tessellate_layer<TensorDataType, T_layout, Dev>::bp_compute_3d(
   }
 }
 
+#ifdef LBANN_HAS_CUTENSOR
 template <typename TensorDataType, data_layout T_layout, El::Device Dev>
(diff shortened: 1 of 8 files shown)


benson31 (approved):
Looks great

src/layers/transform/tessellate.cpp
```diff
@@ -134,6 +134,18 @@ void tessellate_layer<TensorDataType, T_layout, Dev>::bp_compute_3d(
...
+  const std::vector<int>& input_dims,
+  const std::vector<int>& output_dims,
+  const El::AbstractDistMatrix<TensorDataType>& output_grad,
+  El::AbstractMatrix<TensorDataType>& input_grad)
+{
+  // Multi-dimensional reduction for CPU Tessellate backprop is the same as the
+  // regular backprop
+  bp_compute_3d(input_dims, output_dims, output_grad, input_grad);
+}
+
```
benson31, line 148:
```suggestion
```

bvanessen (approved):

PR #2467, opened 2024-07-30
Fix Adam

fiedorowicz1:
Removes an extra eps added to the gradients to improve training stability.

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -25,6 +25,9 @@ I/O & data ingestion:
 Build system:
 
 Bug fixes:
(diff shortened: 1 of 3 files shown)


tbennun (approved):
LGTM!

ndryden (approved):

bvanessen (approved):
@fiedorowicz1 Please add a note in the ReleaseNotes.txt that indicates that this eps was removed, and that it had originally been inserted to avoid a performance issue with CPUs and denormalized numbers.  Once that note is added, it is good to go.

PR #2469, opened 2024-09-03
Remove some logic from the superbuild

benson31:
The Superbuild cannot write to `CMAKE_INSTALL_PREFIX` during configure time. There's no guarantee that this variable is even set, or that it's a valid and writeable path if it is set (indeed, the default path is likely not writeable by the average superbuild user). Moreover, even if it is set, each package maintains its own install prefix, which may be wholly unrelated to whatever is in the `CMAKE_INSTALL_PREFIX`. Finally, the superbuild doesn't define an `install` target at all, so we cannot simply copy the file at "superbuild install time" since such a thing doesn't exist (and this certainly is not a compelling reason to change that).

In spite of this seemingly bleak status quo, clients of the superbuild are much more empowered than the superbuild can inherently be, and they have access to all of the relevant information. In the case of LBANN's CI infrastructure, we _do_ choose to install all of the dependencies under a common prefix, so **we** can manually copy this file over to that prefix any time after configuration has completed.

As the last bit of cleanup, I removed a message about python something or another that's LBANN's problem and should go in **LBANN**'s CMake, not the superbuild's (the message had nothing to do with interactions with or among any of the packages known to the superbuild, and therefore it is irrelevant to the goals of the superbuild system). Also, as far as I know, CNPY has no dependence on Python or the `numpy` python module, so I also culled that message, also believing it to be irrelevant and/or erroneous (LBANN's dependence (or not) on `numpy` is a concern for the LBANN project and not the superbuild as [it is not a package manager](https://github.com/LLNL/lbann/blob/develop/scripts/superbuild/README.md?plain=1#L42-L43)).

### DIFF
--- a/scripts/superbuild/CMakeLists.txt
+++ b/scripts/superbuild/CMakeLists.txt
@@ -154,25 +154,8 @@ if (LBANN_SB_BUILD_AWS_OFI_RCCL)
   message("export LD_LIBRARY_PATH=${LBANN_SB_AWS_OFI_RCCL_PREFIX}/lib:\$\{LD_LIBARY_PATH\}\n")
   message("-----------------------------------------------------------------\n")
   file(APPEND "${CMAKE_BINARY_DIR}/lbann_sb_suggested_cmake_prefix_path.sh"
-    "export AWS_OFI_RCCL_LIBDIR=${LBANN_SB_AWS_OFI_RCCL_PREFIX}/lib\n")
-  file(APPEND "${CMAKE_BINARY_DIR}/lbann_sb_suggested_cmake_prefix_path.sh"
-    "export LD_LIBRARY_PATH=${LBANN_SB_AWS_OFI_RCCL_PREFIX}/lib:\$\{LD_LIBRARY_PATH\}\n")
-  file(APPEND "${CMAKE_INSTALL_PREFIX}/logs/lbann_sb_suggested_cmake_prefix_path.sh"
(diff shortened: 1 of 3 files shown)


scripts/superbuild/CMakeLists.txt
```diff
@@ -154,25 +154,8 @@ if (LBANN_SB_BUILD_AWS_OFI_RCCL)
...
-  message("  python3 -m pip install --target \$\{LBANN_PYTHON_SITE_PACKAGES\} pytest")
-  message("  python3 -m pip install --target \$\{LBANN_PYTHON_SITE_PACKAGES\} protobuf")
-  if (LBANN_SB_FWD_LBANN_LBANN_WITH_CNPY)
-    message("\nLBANN was built with support for the NumPy (If you need to install it via pip you can in the LBANN site-packages with):")
-    message("  python3 -m pip install --target \$\{LBANN_PYTHON_SITE_PACKAGES\} numpy")
-  endif ()
-  message("\n-----------------------------------------------------------------\n")
+    "export AWS_OFI_RCCL_LIBDIR=${LBANN_SB_AWS_OFI_RCCL_PREFIX}/lib\n
+    export LD_LIBRARY_PATH=${LBANN_SB_AWS_OFI_RCCL_PREFIX}/lib:\$\{LD_LIBRARY_PATH\}\n")
 endif ()
```
bvanessen, line 159: @benson31 Why did you remove these messages?

bvanessen (approved):

PR #2473, opened 2024-09-24
added template for data reader to pass conduit node from driver

bvanessen:
added conduit to cmakelist

fixed error with global_trainer_

added simple conduit datareader to hold conduit node

prototyping use of data_store to hold conduit nodes

fixing bug with input buffers not being sized correctly

fixed problem with unpacking conduit node

moving {trainer, dc, dr, ds setup} and {loading inference samples} to separate functions

extended core API for many different input types

removed old code from first lbann-core impl

added simple run script

Fix things that have drifted in LBANN

Get core-drive compiling again

clang-format batch_functional_inference_algorithm

Steps toward debugging the segfault in the inference algo test

The test no longer segfaults. Now it just fails.

Don't shuffle when setting up for inference

Fix a spacing issue

Updated CMake to install the core driver

Build the core-driver


benson31 (commented):
Initial notes.

PR #2481, opened 2026-03-25
Patch things for Torch's 2.11 release

benson31:
The main change is that they've moved hip support to just exist under "cuda"/"CUDA" names (namespaces or symbols). This likely isn't the best fix, but hopefully it's an understandable and maintainable one.


fiedorowicz1 (approved):
LGTM

